PR #55, opened 2021-01-07
[node] load project tar

### LINKED ISSUE #8: [@subql/node] support load subquery project from its tarball pack
It is promised in the grant, and we have already supported packing subquery project into tarball.

For `@subql/node`, it will need to unpack the tarball file into a temp directory, then it will be no different with current flow.
Also the checksum of the tarball file need to be recorded in db.

jiqiang90:
Closes #8 support unpack the tarball file into a temp directory, and return the temp path for subquery project

### DIFF
--- a/packages/node/src/configure/configure.module.ts
+++ b/packages/node/src/configure/configure.module.ts
@@ -63,6 +63,11 @@ export class ConfigureModule {
       config.subquery,
     );
 
+    const project = async () =>
+      SubqueryProject.create(projectPath).catch((err) => {
+        console.error('Create Subquery project from given path failed !', err);
+        process.exit(1);
+      });
     return {
       module: ConfigureModule,
       providers: [
@@ -72,7 +77,7 @@ export class ConfigureModule {
         },
         {
           provide: SubqueryProject,
-          useFactory: async () => SubqueryProject.create(projectPath),
+          useFactory: project,
         },
       ],
       exports: [NodeConfig, SubqueryProject],
--- a/packages/node/src/configure/configure.module.ts
+++ b/packages/node/src/configure/configure.module.ts
@@ -72,10 +72,7 @@ export class ConfigureModule {
         },
         {
           provide: SubqueryProject,
-          useValue: SubqueryProject.create(
-            loadProjectManifest(projectPath),
-            projectPath,
-          ),
+          useFactory: async () => SubqueryProject.create(projectPath),
         },
       ],
       exports: [NodeConfig, SubqueryProject],
--- a/packages/node/src/configure/configure.module.ts
+++ b/packages/node/src/configure/configure.module.ts
@@ -65,7 +65,10 @@ export class ConfigureModule {
 
     const project = async () =>
       SubqueryProject.create(projectPath).catch((err) => {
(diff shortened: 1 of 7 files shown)


packages/node/src/utils/project.ts
```diff
@@ -0,0 +1,18 @@
+import fs from 'fs';
+import os from 'os';
+import path from 'path';
+import tar from 'tar';
+
+export async function processPath(projectPath: string): Promise<string> {
```
ianhe8x, line 6: can use a better name, like prepareProjectDir

packages/node/src/utils/project.ts
```diff
@@ -0,0 +1,18 @@
...
+  const stats = fs.statSync(projectPath);
+  if (stats.isFile()) {
+    const sep = path.basename(projectPath);
```
ianhe8x, line 9: `sep` means seperator, we don't really need the original basename to name temp dir

packages/node/src/configure/configure.module.ts
```diff
@@ -72,10 +72,7 @@ export class ConfigureModule {
         },
         {
           provide: SubqueryProject,
-          useValue: SubqueryProject.create(
-            loadProjectManifest(projectPath),
-            projectPath,
-          ),
+          useFactory: async () => SubqueryProject.create(projectPath),
```
ianhe8x, line 75: if path is not a valid tar file, the application will crash. We need to catch the error to make the reason of crash more clear.

ianhe8x (changes requested):

packages/node/src/utils/project.ts
```diff
@@ -0,0 +1,18 @@
+import fs from 'fs';
```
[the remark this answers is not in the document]
jiqiang90: license header added

packages/node/src/utils/project.ts
jiqiang90: Name changed

packages/node/src/utils/project.ts
jiqiang90: Fixed

packages/node/src/configure/configure.module.ts
jiqiang90: catch error added

packages/node/src/configure/project.model.ts
```diff
@@ -1,13 +1,21 @@
...
+  ProjectManifest,
+  SubqlDataSource,
+} from '@subql/common';
+import { prepareProjectDir } from '../utils/project';
 
 export class SubqueryProject implements ProjectManifest {
-  static create(source: ProjectManifest, path: string): SubqueryProject {
+  static async create(path: string): Promise<SubqueryProject> {
+    const projectPath = await prepareProjectDir(path);
+    console.log(projectPath);
```
ianhe8x, line 14: Is this necessary output? remove if it is only for debugging

ianhe8x (changes requested):

packages/node/src/configure/project.model.ts
jiqiang90: removed

ianhe8x (approved):

PR #60, opened 2021-01-11
add priority to find subquery project entry point from package json file

### LINKED ISSUE #44: [node] use main field in package.json for subquery's entry file
Currently subql node will try to load `dist/index` from the given local subquery path, but rather we should first find package.json from that path and use the path defined in `main` field as the entry file path.

zhex:
close #44

### DIFF
--- a/packages/node/src/indexer/indexer.manager.ts
+++ b/packages/node/src/indexer/indexer.manager.ts
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import path from 'path';
+import fs from 'fs';
 import { Inject, Injectable, OnApplicationBootstrap } from '@nestjs/common';
 import { buildSchema, getAllEntities, SubqlKind } from '@subql/common';
 import { QueryTypes, Sequelize } from 'sequelize';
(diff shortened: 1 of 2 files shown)


packages/node/src/indexer/indexer.manager.ts
```diff
@@ -157,6 +158,15 @@ export class IndexerManager implements OnApplicationBootstrap {
       sandbox: {
         store: this.storeService.getStore(),
         api: this.apiService.getApi(),
+        getProjectEntry: () => {
+          const pkgPath = path.join(this.nodeConfig.subquery, 'package.json');
```
ianhe8x, line 162: I think we better do this in the host context, e.g when we process project manifest file. 
If we do it in sandbox it will be called every time a block is feed to the mapping function in sandbox.

ianhe8x (changes requested):

packages/node/src/indexer/indexer.manager.ts
zhex: changed to lookup entry point when init the vm.

ianhe8x (approved):

PR #61, opened 2021-01-11
[node] Support node retry to establish connection with postgres

### LINKED ISSUE #57: [node] Node initialization should await Postgres get ready
Current node image often failed during initialization, due to it start running before the Postgres ready to accept connection.

jiqiang90:
Close #57 , now node can make 5 attempts to connect to Postgres with time interval of 3 seconds.

### DIFF
--- a/packages/node/src/db/db.module.ts
+++ b/packages/node/src/db/db.module.ts
@@ -1,6 +1,5 @@
 // Copyright 2020-2021 OnFinality Limited authors & contributors
 // SPDX-License-Identifier: Apache-2.0
-
 import { DynamicModule, Global } from '@nestjs/common';
 import { Sequelize } from 'sequelize';
 import * as entities from '../entities';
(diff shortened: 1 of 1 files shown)


packages/node/src/db/db.module.ts
```diff
@@ -13,6 +12,35 @@ export interface DbOption {
...
+): Promise<void> {
+  if (numRetries > 0) {
+    try {
+      console.log(
+        'Try to connect to database: ',
+        numRetries,
+        'attempts remaining',
+      );
+      await sequelize.authenticate();
+      console.log('Connection has been established successfully.');
```
ianhe8x, line 31: Before this pr we don't info user when db connection is established, so I suggest we keep same behaviour

packages/node/src/db/db.module.ts
```diff
@@ -13,6 +12,35 @@ export interface DbOption {
...
+    } catch (error) {
+      console.error('Unable to connect to the database:', error.message);
```
ianhe8x, line 33: combine these two logs together, e.g: 
```suggestion
      console.error('Unable to connect to the database, will retry: ', error.message);
```

packages/node/src/db/db.module.ts
```diff
@@ -13,6 +12,35 @@ export interface DbOption {
...
 
+async function timeout(ms) {
+  return new Promise((resolve) => setTimeout(resolve, ms));
+}
+
+async function establishConnection(
+  sequelize: Sequelize,
+  numRetries,
+): Promise<void> {
+  if (numRetries > 0) {
```
ianhe8x, line 23: check left retries in the `catch` block

ianhe8x (changes requested):

packages/node/src/db/db.module.ts
jiqiang90: Fixed

packages/node/src/db/db.module.ts
jiqiang90: Fixed

packages/node/src/db/db.module.ts
jiqiang90: Fixed

ianhe8x (approved):

PR #72, opened 2021-01-13
k8s deploy scripts

### LINKED ISSUE #71: deploy k8s yaml
For people who wants to use it in k8s

ianhe8x:
Closes #71

### DIFF
--- a/deploy/k8s/README.md
+++ b/deploy/k8s/README.md
@@ -1,0 +1,200 @@
+# K8s Deploy
+This [deploy.yaml](deploy.yaml) demonstrate how to run subquery in k8s.
+
+You will need to run 
+### 1. a subql-node to index data
+```yaml
+apiVersion: apps/v1
+kind: StatefulSet
+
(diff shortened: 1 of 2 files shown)


deploy/k8s/README.md
```diff
@@ -0,0 +1,200 @@
+# K8s Deploy
+This [deploy.yaml](deploy.yaml) demonstrate how to run subquery in k8s.
+
+You will need to run 
+### 1. a subql-node to index data
+```yaml
+apiVersion: apps/v1
+kind: StatefulSet
```
zhex, line 8: in my opinion, we can use deployment for the indexer, coz we use in-memory volume here, it will be destroyed after the pod was gone.

zhex (changes requested):

deploy/k8s/README.md
ianhe8x: good point

zhex (approved):

PR #79, opened 2021-01-14
[node] Fix BigInt transformation

### LINKED ISSUE #73: [node] fix BigInt transformation when column value is null
In objectTypeToModelAttributes we defined a custom getter and setter to transform column value to / from BigInt.

But it didn't handle null well.

jiqiang90:
Close #73

### DIFF
--- a/packages/node/src/utils/graphql.ts
+++ b/packages/node/src/utils/graphql.ts
@@ -34,11 +34,8 @@ export function objectTypeToModelAttributes(
     };
     if (type.toString() === 'BigInt') {
       columnOption.get = function () {
-        if (this.getDataValue(k)) {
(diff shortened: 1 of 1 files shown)


packages/node/src/utils/graphql.ts
```diff
@@ -34,10 +34,14 @@ export function objectTypeToModelAttributes(
     };
     if (type.toString() === 'BigInt') {
       columnOption.get = function () {
-        return BigInt(this.getDataValue(k));
+        if (this.getDataValue(k)) {
```
ianhe8x, line 37: assign the result to a variable

ianhe8x (changes requested):

ianhe8x (approved):

PR #98, opened 2021-01-20
Feature/query service

zhex:
implemented a basic query service base.


zhex:
will PR to next/** branch

PR #100, opened 2021-01-20
[SKIP CI] add test for cli

jiqiang90:


packages/cli/src/controller/codegen-controller.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+    writeFile: jest.fn(),
+    access: jest.fn(),
+  };
+  fsMock.readFileSync = jest.fn();
+  return fsMock;
+});
+
+jest.mock('ejs', () => {
+  const ejsMock = jest.createMockFromModule('ejs') as any;
+  ejsMock.renderFile = jest.fn();
```
ianhe8x, line 19: This line seems not necessary, it has been done by `createMockFromModule`

packages/cli/src/controller/codegen-controller.ts
```diff
@@ -8,33 +8,37 @@ import ejs from 'ejs';
 import {isNonNullType, getNullableType} from 'graphql';
 import {GraphQLFieldMap, GraphQLOutputType} from 'graphql/type/definition';
 import {transformTypes} from './types-mapping';
+
 const templatePath: string = path.resolve(__dirname, '../template/model.ts.ejs');
-const typesPath = `${process.cwd()}/src/types/models`;
 
 // 4. Save the rendered schema
-function makeSchema(className: string, data: string): void {
+export async function makeSchema(className: string, data: string): Promise<void> {
```
ianhe8x, line 15: this abstraction doesn't make sense to me, it is simply wrap of fs.write and once be called once

ianhe8x (changes requested):

packages/cli/src/controller/codegen-controller.spec.ts
jiqiang90: removed

packages/cli/src/controller/codegen-controller.ts
jiqiang90: merged into renderTemplate

PR #149, opened 2021-02-03
[common] load manifest should throw error if manifest contain array

jiqiang90:


packages/common/src/project/load.ts
```diff
@@ -21,8 +21,12 @@ export function loadProjectManifest(file: string): ProjectManifest {
   const errors = validateSync(manifest, {whitelist: true, forbidNonWhitelisted: true});
   if (errors?.length) {
     // TODO: print error details
-    const errorMsgs = errors.map((e) => e.toString()).join('\n');
-    throw new Error(`failed to parse project.yaml.\n${errorMsgs}`);
+    try {
+      const errorMsgs = errors.map((e) => e.toString()).join('\n');
```
ianhe8x, line 25: I upgrade `class-validator`, can you test it against the new version and see if we still need to make this patch?

jiqiang90: Working now, close this pr

PR #168, opened 2021-02-15
[node] resume connection and indexing

### LINKED ISSUE #142: [node] auto resume indexing after ws reconnect
Currently, if ws connection was disconnected and then recovered, indexing process can not automatically resume.

jiqiang90:
Closes #142

### DIFF
--- a/packages/node/src/indexer/indexer.manager.ts
+++ b/packages/node/src/indexer/indexer.manager.ts
@@ -96,6 +96,7 @@ export class IndexerManager {
         block.block.header.number.toNumber() + 1;
       await this.subqueryState.save();
       this.fetchService.latestProcessed(block.block.header.number.toNumber());
+      await tx.commit();
     } catch (e) {
       logger.error(
         `failed to handler block ${block.block.header.number.toString()}, error: ${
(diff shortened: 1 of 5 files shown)


packages/node/src/indexer/indexer.manager.ts
```diff
@@ -125,7 +122,12 @@ export class IndexerManager {
         // FIXME: retry before exit
         process.exit(1);
       });
-    this.fetchService.register((block) => this.indexBlock(block));
+    try {
```
ianhe8x, line 125: This won't work, register() always success

packages/node/src/utils/promise.ts
```diff
@@ -6,3 +6,13 @@ export async function delay(sec: number): Promise<void> {
...
   });
 }
+
+export async function timeout(
+  promise: Promise<void>,
+  sec: number,
+): Promise<void> {
+  return new Promise((resolve, reject) => {
+    setTimeout(() => reject(new Error('promise timeout')), sec * 1000);
+    promise.then(resolve).catch(reject);
```
ianhe8x, line 16: This way even a promise is resolved, reject() will still be called.
You should use `Promise.race()` instead.

ianhe8x (changes requested):

ianhe8x (approved):

PR #183, opened 2021-02-19
[node] Add benchmark service

### LINKED ISSUE #121: Add bps benchmark (block processed per second)
Similar output to substrate node
```
2021-01-17 00:00:00 [Subquery Project Name] 10.0 bps, target=#xxx, current: #xxx, estimate time: x d, y h, z m 
```

jiqiang90:
Close #121


packages/node/package.json
```diff
@@ -26,6 +26,8 @@
     "@willsoto/nestjs-prometheus": "^3.0.0",
     "app-module-path": "^2.2.0",
     "lodash": "^4.17.20",
+    "moment": "^2.29.1",
```
ianhe8x, line 29: `moment` is a bit heavy, we usually prefer to `dayjs`

packages/node/src/indexer/benchmark.service.ts
```diff
@@ -0,0 +1,83 @@
...
+export class BenchmarkService {
+  private currentProcessingHeight: number;
+  private currentProcessingTimestamp: number;
+  private targetHeight: number;
+  private lastRegisteredHeight: number;
+  private lastRegisteredTimestamp: number;
+  private blockPerSecond: number;
+  private needReset = true;
+
+  async init(): Promise<void> {
```
ianhe8x, line 24: use @nest/scheduler

packages/node/src/indexer/benchmark.service.ts
```diff
@@ -0,0 +1,83 @@
...
+const logger = getLogger('benchmark');
+
+export class BenchmarkService {
+  private currentProcessingHeight: number;
+  private currentProcessingTimestamp: number;
+  private targetHeight: number;
+  private lastRegisteredHeight: number;
+  private lastRegisteredTimestamp: number;
+  private blockPerSecond: number;
+  private needReset = true;
```
ianhe8x, line 22: why need this flag? under which circumstance we don't want to reset last height and last timestamp?

packages/node/src/indexer/benchmark.service.ts
```diff
@@ -0,0 +1,83 @@
...
+
+  private benchmark() {
+    if (this.needReset) {
+      this.resetBlock();
+    } else {
+      logger.debug(
+        `current #${this.currentProcessingHeight}, last #${
+          this.lastRegisteredHeight
+        }, take ${
+          (this.currentProcessingTimestamp - this.lastRegisteredTimestamp) /
```
ianhe8x, line 39: this calculation is done more than once, so pls assign the result to a variable

ianhe8x: actually this debug log doesn't make sense to me, if we want to know whether processing height is changing, we should add log in indexManager not here

packages/node/src/indexer/benchmark.service.ts
```diff
@@ -0,0 +1,83 @@
...
+    this.needReset = false;
+  }
+
+  @OnEvent('block.processing')
+  handleProcessingBlock(blockPayload: BlockPayload) {
+    this.currentProcessingHeight = blockPayload.data.height;
+    this.currentProcessingTimestamp = blockPayload.data.timestamp;
+  }
+
+  @OnEvent('metric.write')
```
ianhe8x, line 77: since you already changed event name for 'block.processing', you should do same refactor for `subql_indexer_target_block_height` as well

packages/node/src/indexer/types.ts
```diff
@@ -12,3 +12,12 @@ export interface BlockContent {
...
   events: SubstrateEvent[];
 }
+
+export enum BlockEvents {
+  BlockProcessing = 'processing_block_event',
+}
+
+export interface BlockPayload {
+  data: { height: number; timestamp: number };
+  name: BlockEvents;
```
ianhe8x, line 22: doesn't make sense here, event name is `block.processing`, why you need another name filed in payload, when the value of it is only one choice.

ianhe8x (changes requested):

packages/node/src/indexer/benchmark.service.ts
```diff
@@ -0,0 +1,68 @@
...
+  private currentProcessingHeight: number;
+  private currentProcessingTimestamp: number;
+  private targetHeight: number;
+  private lastRegisteredHeight: number;
+  private lastRegisteredTimestamp: number;
+  private blockPerSecond: number;
+
+  @Interval(SAMPLING_TIME_VARIANCE * 1000)
+  async benchmark(): Promise<void> {
+    if (!this.currentProcessingHeight && !this.currentProcessingTimestamp) {
```
ianhe8x, line 30:
```suggestion
    if (!this.currentProcessingHeight || !this.currentProcessingTimestamp) {
```

packages/node/src/indexer/benchmark.service.ts
```diff
@@ -0,0 +1,68 @@
...
+      await delay(10);
+    } else {
+      if (this.currentProcessingHeight && this.lastRegisteredTimestamp) {
```
ianhe8x, line 33: no need to check this.currentProcessingHeight

ianhe8x:
```suggestion
      if (this.lastRegisteredHeight && this.lastRegisteredTimestamp) {
```

ianhe8x (changes requested):

packages/node/src/indexer/indexer.manager.ts
```diff
@@ -42,9 +41,9 @@ export class IndexerManager {
   ) {}
 
   async indexBlock({ block, events, extrinsics }: BlockContent): Promise<void> {
-    this.eventEmitter.emit('metric.write', {
-      name: Metrics.ProcessingHeight,
-      value: block.block.header.number.toNumber(),
+    this.eventEmitter.emit('block_processing_height', {
```
ianhe8x, line 44: use IndexerEvent

ianhe8x (approved):

PR #206, opened 2021-03-01
reduce injectMetadata call

zhex:
`injectMetadata` method is heavy, this PR try to improve the performance, we only call this method when the spec version is changed.
this could improve the BPS from 3 to 20.

before the PR is merged, we need to solve the memory leak issue， this is enlarged by the BPS increases

### DIFF
--- a/packages/node/src/indexer/indexer.manager.ts
+++ b/packages/node/src/indexer/indexer.manager.ts
@@ -12,7 +12,6 @@ import { SubqueryProject } from '../configure/project.model';
 import { SubqueryModel, SubqueryRepo } from '../entities';
 import { objectTypeToModelAttributes } from '../utils/graphql';
 import { getLogger } from '../utils/logger';
-import { timeout } from '../utils/promise';
 import * as SubstrateUtil from '../utils/substrate';
 import { ApiService } from './api.service';
 import { IndexerEvent } from './events';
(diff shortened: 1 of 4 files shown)


zhex:
after we patch API only when the metadata changed, we can see the memory leak issue is gone. memory drop to a reasonable value after GC.

packages/node/src/indexer/indexer.manager.ts
```diff
@@ -49,8 +50,13 @@ export class IndexerManager {
     });
     const tx = await this.sequelize.transaction();
     this.storeService.setTransaction(tx);
+
+    const inject = block.specVersion !== this.prevSpecVersion;
+    this.prevSpecVersion = block.specVersion;
```
ianhe8x, line 55: the change of this.prevSpecVersion should be move into try block, after setBlockhash, in case setBlockhash fails

packages/node/src/indexer/indexer.manager.ts
```diff
@@ -49,8 +50,13 @@ export class IndexerManager {
...
+
     try {
-      await timeout(this.apiService.setBlockhash(block.block.hash), 10); //TODO remove this when polkadot/api issue #3197 solved
+      const hash = block.block.hash;
+      await timeout(this.apiService.setBlockhash(hash, inject), 10); //TODO remove this when polkadot/api issue #3197 solved
```
ianhe8x, line 59: btw, the timeout here can be removed, because 3.11.1 has fixed the issue.

packages/node/src/indexer/indexer.manager.ts
```diff
@@ -49,8 +50,13 @@ export class IndexerManager {
     });
     const tx = await this.sequelize.transaction();
     this.storeService.setTransaction(tx);
+
+    const inject = block.specVersion !== this.prevSpecVersion;
+    this.prevSpecVersion = block.specVersion;
+
     try {
-      await timeout(this.apiService.setBlockhash(block.block.hash), 10); //TODO remove this when polkadot/api issue #3197 solved
+      const hash = block.block.hash;
```
ianhe8x, line 58: just a nit-pick, I prefer making it inline, since the new var is only used once, also align with the style of whole project.

ianhe8x (changes requested):

packages/node/src/indexer/indexer.manager.ts
zhex: updated

packages/node/src/indexer/indexer.manager.ts
zhex: reordered the code to prevent the exception happen

ianhe8x (approved):

PR #211, opened 2021-03-02
use parent's specVersion to decide if metadata need to be injected

### LINKED ISSUE #210: fix the timing we do metadata preload
After investigation of an issue, I find what we were doing is wrong, to be specific that we call `api.getBlockRegistry` to load and cache metadata when we detect a specVersion change.

The correct thing to do is compare parent block's specVersion, since current block is run on the code stored in parent block's storage.

Another place is affected by this issue is the specVersion we wrapped in SubstrateBlock object. I'm not sure if anyone is using it at the moment, but just think about the reason why we wrapped it there. there's two,
1. it is used by specVersion filter
2. it is exposed to mapping function so developer can do special treatments for a special specVersion

Both cases I think make more sense if this specVersion is the parent block's specVersion instead of current block's, because that is actually the one indicating which runtime code this block is running.

ianhe8x:
Fixes #210

### DIFF
--- a/packages/node/src/utils/substrate.ts
+++ b/packages/node/src/utils/substrate.ts
@@ -179,7 +179,10 @@ export async function fetchBlocks(
   const blocks = await fetchBlocksRange(api, startHeight, endHeight);
   const parentBlockHashs = blocks.map((b) => b.block.header.parentHash);
   const [blockEvents, runtimeVersions] = await Promise.all([
-    fetchEventsRange(api, parentBlockHashs),
+    fetchEventsRange(
+      api,
+      blocks.map((b) => b.block.header.hash),
+    ),
     overallSpecVer
       ? undefined
       : fetchRuntimeVersionRange(api, parentBlockHashs),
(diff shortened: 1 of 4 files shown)


packages/node/src/utils/substrate.ts
```diff
@@ -163,25 +163,34 @@ export async function prefetchMetadata(
...
   endHeight: number,
   overallSpecVer?: number,
 ): Promise<BlockContent[]> {
   const blocks = await fetchBlocksRange(api, startHeight, endHeight);
-  const blockHashs = blocks.map((b) => b.block.header.hash);
+  const parentBlockHashs = blocks.map((b) => b.block.header.parentHash);
   const [blockEvents, runtimeVersions] = await Promise.all([
-    fetchEventsRange(api, blockHashs),
-    overallSpecVer ? undefined : fetchRuntimeVersionRange(api, blockHashs),
+    fetchEventsRange(api, parentBlockHashs),
```
zhex, line 182: if we use `parentBlockHashs` to fetch the events, then the events and blocks are miss matched

ianhe8x: good catch

zhex (approved):

PR #212, opened 2021-03-03
[common] add get models and relations for 1-1, and 1-N 

jiqiang90:
Will add more more test cases

### DIFF
--- a/packages/common/src/graphql/entities.ts
+++ b/packages/common/src/graphql/entities.ts
@@ -1,16 +1,18 @@
 // Copyright 2020-2021 OnFinality Limited authors & contributors
 // SPDX-License-Identifier: Apache-2.0
 
+import assert from 'assert';
 import {
   assertListType,
   getDirectiveValues,
   getNullableType,
   GraphQLField,
-  GraphQLObjectType, GraphQLOutputType,
+  GraphQLObjectType,
+  GraphQLOutputType,
   GraphQLSchema,
   isListType,
   isNonNullType,
-  isObjectType
+  isObjectType,
 } from 'graphql';
 import {DirectiveName} from './constant';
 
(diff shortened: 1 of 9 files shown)


packages/common/src/graphql/entities.ts
```diff
@@ -1,14 +1,126 @@
...
+    entityNameSet.push(entity.name);
+  }
+
+  for (const entity of entities){
+    const newModel: GraphQLModelsType=  {
+      name: entity.name,
+      fields : []
+    }
+    const fields = entity.getFields();
+    for (const k in fields){
```
zhex, line 51: if you change from 'for..in' to 'for..of', the code will be much simpler here, no hashOwnProperty check needed.

packages/common/src/graphql/entities.ts
```diff
@@ -1,14 +1,126 @@
...
 }
 
-// TODO: GraphQLNamedType -> EntitySchema
+export function getAllEntitiesRelations(schema: GraphQLSchema): GraphQLModelsRelations{
+  const modelRelations = {models:[],relations:[]} as GraphQLModelsRelations;
+  const entityNameSet = [];
+  const reverseLookupRelations = [];
+
+  const derivedFrom = schema.getDirective('derivedFrom');
+  const entities = Object.entries(schema.getTypeMap())
```
ianhe8x, line 36: if you only need value, you can use `Object.values()`

packages/common/src/graphql/entities.ts
```diff
@@ -1,14 +1,126 @@
...
     .filter(isObjectType);
 }
 
-// TODO: GraphQLNamedType -> EntitySchema
+export function getAllEntitiesRelations(schema: GraphQLSchema): GraphQLModelsRelations{
+  const modelRelations = {models:[],relations:[]} as GraphQLModelsRelations;
+  const entityNameSet = [];
+  const reverseLookupRelations = [];
+
+  const derivedFrom = schema.getDirective('derivedFrom');
```
ianhe8x, line 35: the declaration is so far away from the place it is used.

packages/common/src/graphql/entities.ts
```diff
@@ -1,14 +1,126 @@
...
+  const entities = Object.entries(schema.getTypeMap())
+      .filter(([, node]) => node.astNode?.directives?.find(({name: {value}}) => value === DirectiveName.Entity))
+      .map(([, node]) => node)
+      .filter(isObjectType);
+
+  for (const entity of entities){
```
ianhe8x, line 41: you can use entities.map()

packages/common/src/graphql/entities.ts
```diff
@@ -1,14 +1,126 @@
...
+    nullable: !isNonNullType(field.type),
+  }
+  if(fieldType === "foreignKey"){
+    entityField.name = `${entityField.name}Id`; //Update name
+    entityField.type = 'String';
+  }
+  return entityField
+}
+//Get the type, ready to be convert to string
+function extractType(type:any){
```
ianhe8x, line 104: need complete types

packages/common/src/graphql/entities.ts
```diff
@@ -1,14 +1,126 @@
...
+      }
+    }
+    modelRelations.models.push(newModel);
+  }
+  handleDerived(reverseLookupRelations, modelRelations.relations);
+  return modelRelations;
+}
+
+
+function packEntityField(typeString:any, field:GraphQLField<any,any>, fieldType?:string) : GraphQLEntityField{
```
ianhe8x, line 90: use enum for fieldType rather than string

ianhe8x: also fieldType='base' is never checked, that means your abstraction for this method can be improved

packages/common/src/graphql/entities.ts
```diff
@@ -1,14 +1,126 @@
...
+    const fields = entity.getFields();
+    for (const k in fields){
+      if (Object.prototype.hasOwnProperty.call(fields, k)) {
+        const typeString = extractType(fields[k].type).toString();
+        //If is a basic scalar type
+        if (Object.values(FieldScalar).includes(typeString)) {
+          newModel.fields.push(packEntityField(typeString, fields[k],'basic'));
+        }
+        // If is a foreign key
+        else if (entityNameSet.includes(typeString) && fields[k].astNode.directives.length ===0){
```
ianhe8x, line 59: `fields[k].astNode.directives.length ===0` is not the only scenario that `@deriveFrom` isn't present, you better improve the condition branches otherwise it is likely you will have a lot of duplicated codes

packages/common/src/graphql/entities.ts
```diff
@@ -1,14 +1,126 @@
...
+            reverseLookupRelations.push({
+              fromEntity:entity.name,  //Account
+              toEntity: typeString,     //Identity
+              type: isListType(isNonNullType(fields[k].type) ? getNullableType(fields[k].type) : fields[k].type)? 'hasMany':'hasOne',
+              foreignKey: `${directValues.field}Id`, //accountId
+              fieldName: fields[k].name, //identity
+            })
+          }
+        }
+      }
```
ianhe8x, line 81: what if typeString is not basic types and isn't found in entityNameSet?

packages/common/src/graphql/entities.ts
```diff
@@ -1,14 +1,126 @@
...
+  return entityField
+}
+//Get the type, ready to be convert to string
+function extractType(type:any){
+  const offNullType = isNonNullType(type) ?  getNullableType(type) : type ;
+  const offListType = isListType(offNullType)?assertListType(offNullType).ofType: type;
+  return isNonNullType(offListType)? getNullableType(offListType) : offListType;
+}
+
+function handleDerived(reverseLookupRelations,relations:GraphQLRelationsType[]):void{
```
ianhe8x, line 110: It's confusing what you want to achieve here. Is it necessary to differentiate  reverseLookupRelations from other relations?

packages/common/src/graphql/entities.ts
```diff
@@ -1,14 +1,126 @@
...
+  for (const r of reverseLookupRelations){
+    for (const relation of relations){
```
ianhe8x, line 112: you should use `relations.find()` which stops at first match

packages/common/src/graphql/graphql.spec.ts
```diff
@@ -41,5 +40,60 @@ describe('utils that handle schema.graphql', () => {
...
+  });
+
+  it('can extract list field with entity from the schema', () => {
+    const graphqlSchema = gql`
+      type KittyBirthInfo @entity {
+        id: ID!,
+        owner: String,
+        paper: paper @derivedFrom(field: "account")
+      }
+      type paper @entity {
```
ianhe8x, line 58: could use a better name than paper.
and pls use capital case for type definition

ianhe8x (changes requested):

packages/common/src/graphql/entities.ts
```diff
@@ -1,14 +1,114 @@
...
+  const offListType = isListType(offNullType)?assertListType(offNullType).ofType: type;
+  return isNonNullType(offListType)? getNullableType(offListType) : offListType;
+}
+
+function validateRelations(modelRelations:GraphQLModelsRelations):void{
+  for (const r of modelRelations.relations.filter((model)=> (model.type === 'hasMany'|| model.type === 'hasOne'))){
+    if (
+        modelRelations.models.find((model)=> model.name === r.to &&
+            model.fields.find((field)=> field.name === r.foreignKey) )){
+      continue
```
ianhe8x, line 108: continue is not necessary here. and you can use `assert` to achieve the same thing.

packages/common/src/graphql/entities.ts
```diff
@@ -1,14 +1,108 @@
...
+          foreignKey: `${directValues.field}Id`,
+          fieldName: field.name,
+        } as GraphQLRelationsType);
+      } else {
+        throw new Error(`${typeString} is not an valid type`);
+      }
+    }
+    modelRelations.models.push(newModel);
+  }
+  validateRelations(modelRelations);
```
zhex, line 76: `validateRelations` looks like a test function, we need to make sure every model/relations is valid before insert;
 I don't think we need this validate here.

jiqiang90: To validate relations will require prepare full set of models in advance. Actually, only relations with 'hasMany' need to be validated here, and we need to check it has valid foreignKey in the corresponding entity. 

Also, if validation failed, it will throw errors before it gets inserted into the database.

ianhe8x (approved):

ianhe8x:
It's good enough for now, I will merge the PR.

PR #235, opened 2021-03-11
[node ]Fix benchmark output format issues

jiqiang90:


packages/node/src/indexer/benchmark.service.ts
```diff
@@ -39,13 +39,26 @@ export class BenchmarkService {
...
           .duration(
             (this.targetHeight - this.currentProcessingHeight) /
               this.blockPerSecond,
             'seconds',
           )
-          .format('D [days] HH [hours] MM [mins]');
+          .format('HH [hours] mm [mins]');
+
+        const days = Math.floor(
+          dayjs
```
ianhe8x, line 51: duplicated duration calculation

ianhe8x (approved):

PR #236, opened 2021-03-15
[SKIP CI] Docs update nested relations

jiqiang90:


docs/define_a_subquery.md
```diff
@@ -85,16 +85,113 @@ filter:
...
+
+type Contact @entity {
+  id: ID!
+  phone: String!
+}
+```
+
+#### Many-to-Many relationships
+
+Many-to-Many relationships can be implement by pointing to the same entity name in multiple fields of another entity.
```
ianhe8x, line 160: this is a special case for many to many, when both side is the same entity. but there are other cases as well.

I think the key point is a middle entity to connect both side

ianhe8x (approved):

PR #247, opened 2021-03-25
[node] add network filter for dataSources

### LINKED ISSUE #188: Support network filter for datasource
## Background
It is very common that user will create a subquery and expect to (re)use it for their testnet and mainnet. (e.g polkadot and kusama).

Network endpoint will be different for different networks, and since we already supported `--network-endpoint` flag to override it, this won't be an issue.

The other problem is startBlock, i suppose in some cases user will want it different.

After some discussion, our solution to this is let user define datasource for each of network they need to support, with a network filter. For the `mapping` part, we can make use of yaml's anchor to avoid redundancy.
Something like
```
definitions:
  - mapping: &mymapping
       handlers:
         - handler: handleBlock
           kind: substrate/BlockHandler

dataSources:
  - name: runtime
    kind: substrate/Runtime
    filter: 
        specName: polkadot
    startBlock: 1000
    mapping: *mymapping
  - name: runtime
    kind: substrate/Runtime
    filter: 
        specName: kusama
    startBlock: 2000
    mapping: *mymapping
```

jiqiang90:
Closes #188

### DIFF
--- a/packages/node/src/indexer/indexer.manager.ts
+++ b/packages/node/src/indexer/indexer.manager.ts
@@ -53,11 +53,13 @@ export class IndexerManager {
       const inject = block.specVersion !== this.prevSpecVersion;
       await this.apiService.setBlockhash(block.block.hash, inject);
 
-      const dataSources = this.project.dataSources.filter(ds=>
-        ds.startBlock <= block.block.header.number.toNumber() &&
-        !ds.filter?.specName||ds.filter.specName === this.api.runtimeVersion.specName.toString()
-      )
-      if (dataSources.length === 0){
+      const dataSources = this.project.dataSources.filter(
+        (ds) =>
+          (ds.startBlock <= block.block.header.number.toNumber() &&
+            !ds.filter?.specName) ||
+          ds.filter.specName === this.api.runtimeVersion.specName.toString(),
+      );
+      if (dataSources.length === 0) {
         logger.error(
           `Did not find any dataSource match with network specName ${this.api.runtimeVersion.specName}`,
         );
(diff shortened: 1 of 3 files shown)


packages/node/src/indexer/indexer.manager.ts
```diff
@@ -53,10 +53,20 @@ export class IndexerManager {
       const inject = block.specVersion !== this.prevSpecVersion;
       await this.apiService.setBlockhash(block.block.hash, inject);
 
-      for (const ds of this.project.dataSources) {
-        if (ds.startBlock > block.block.header.number.toNumber()) {
-          continue;
-        }
+      const dataSources = this.project.dataSources.filter(
+        (ds) =>
+          (ds.startBlock <= block.block.header.number.toNumber() &&
```
ianhe8x, line 58:
```suggestion
          (ds.startBlock <= block.block.header.number.toNumber()) &&
          (!ds.filter?.specName || ds.filter.specName === this.api.runtimeVersion.specName.toString())
```

packages/node/src/indexer/indexer.manager.ts
```diff
@@ -157,7 +167,14 @@ export class IndexerManager {
         dbSchema: projectSchema,
         hash: '0x',
         nextBlockHeight: Math.min(
-          ...this.project.dataSources.map((item) => item.startBlock ?? 1),
+          ...this.project.dataSources
```
ianhe8x, line 170: what if it's empty after applying filter. Math.min() will resolve to Infinlity.

ianhe8x (changes requested):

packages/node/src/indexer/indexer.manager.ts
```diff
@@ -134,6 +144,22 @@ export class IndexerManager {
...
+  private getStartBlockFromDataSources() {
+    const startBlocksList = this.project.dataSources
+      .filter(
+        (ds) =>
+          !ds.filter?.specName ||
+          ds.filter.specName === this.api.runtimeVersion.specName.toString(),
+      )
+      .map((item) => item.startBlock ?? 1);
+    if (startBlocksList.length === 0) {
+      logger.error(`Failed to get start block from valid data sources`);
```
ianhe8x, line 156: error message not very clear, try `Failed to find a valid datasource, Please check your endpoint if specName filter is used.`

ianhe8x (approved):

PR #271, opened 2021-04-14
support @index and query by indexed field

### LINKED ISSUE #255: Allow user to annotate Indexes on fields
Support a new annotation `@Index`, that can be added on any field of entity except primary key or foreign key.

So that `@subql/node` will recognize it and create table with proper indexes.

Later on, we can enable query on those indexed fields via `global.store`.

ianhe8x:
Closes #255

### DIFF
--- a/packages/common/src/graphql/schema/directives.ts
+++ b/packages/common/src/graphql/schema/directives.ts
@@ -6,4 +6,5 @@ import gql from 'graphql-tag';
 export const directives = gql`
   directive @derivedFrom(field: String!) on FIELD_DEFINITION
   directive @entity on OBJECT
+  directive @index(unique: Boolean) on FIELD_DEFINITION
 `;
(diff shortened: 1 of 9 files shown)


packages/common/src/graphql/schema/directives.ts
```diff
@@ -6,4 +6,5 @@ import gql from 'graphql-tag';
 export const directives = gql`
   directive @derivedFrom(field: String!) on FIELD_DEFINITION
   directive @entity on OBJECT
+  directive @index(unique: Boolean) on FIELD_DEFINITION
```
zhex, line 9: I saw you have index  type enum, so the directive should have another optional property to set index type

ianhe8x: Yes, I defined the enum but not plan to implement it in this PR

ianhe8x: Probably don't have to let user to decide index type, so we can enforce the best practice

zhex (approved):

PR #275, opened 2021-04-15
Support json type

### LINKED ISSUE #270: Support JSON types
- [x] Add type conversion in common scalar types.
- [x] Support generate correct types in codegen
- [x] Support in query 

Preferred pattern : 

```graphql

type starterEntity @entity {
  field6: JsonField
  field7: [JsonField]
  field8: [String]
}


type JsonField @jsonField {
  field1: Int!
  field2: BigInt
  field3: []
}



```

jiqiang90:
Closes #270

### DIFF
--- a/packages/cli/src/controller/codegen-controller.ts
+++ b/packages/cli/src/controller/codegen-controller.ts
@@ -88,7 +88,7 @@ export function processFields(
       required: !field.nullable,
       isArray: field.isArray,
     } as ProcessedField;
-    if (type === 'entity' && indexFields) {
+    if (type === 'entity') {
       const [indexed, unique] = indexFields.reduce<[boolean, boolean]>(
         (acc, indexField) => {
           if (indexField.fields.includes(field.name)) {
--- a/packages/cli/src/controller/codegen-controller.ts
+++ b/packages/cli/src/controller/codegen-controller.ts
@@ -4,20 +4,33 @@
 import fs from 'fs';
 import path from 'path';
 import {promisify} from 'util';
-import {getAllEntitiesRelations, loadProjectManifest} from '@subql/common';
-import {GraphQLEntityField, GraphQLEntityIndex} from '@subql/common/graphql/types';
+import {
+  getAllEntitiesRelations,
+  buildSchema,
+  loadProjectManifest,
+  getAllJsonObjects,
+  setJsonObjectType,
+} from '@subql/common';
+import {GraphQLEntityField, GraphQLJsonFieldType, GraphQLEntityIndex} from '@subql/common/graphql/types';
 import ejs from 'ejs';
 import {upperFirst} from 'lodash';
 import rimraf from 'rimraf';
 import {transformTypes} from './types-mapping';
 
-const TEMPLATE_PATH = path.resolve(__dirname, '../template/model.ts.ejs');
-
+const MODEL_TEMPLATE_PATH = path.resolve(__dirname, '../template/model.ts.ejs');
+const MODELS_INDEX_TEMPLATE_PATH = path.resolve(__dirname, '../template/models-index.ts.ejs');
+const TYPES_INDEX_TEMPLATE_PATH = path.resolve(__dirname, '../template/types-index.ts.ejs');
+const INTERFACE_TEMPLATE_PATH = path.resolve(__dirname, '../template/interface.ts.ejs');
+const TYPE_ROOT_DIR = 'src/types';
 const MODEL_ROOT_DIR = 'src/types/models';
+const exportTypes = {
+  models: false,
+  interfaces: false,
+};
 
 // 4. Render entity data in ejs template and write it
-export async function renderTemplate(outputPath: string, templateData: ejs.Data): Promise<void> {
-  const data = await ejs.renderFile(TEMPLATE_PATH, templateData);
+export async function renderTemplate(template_path: string, outputPath: string, templateData: ejs.Data): Promise<void> {
+  const data = await ejs.renderFile(template_path, templateData);
   await fs.promises.writeFile(outputPath, data);
 }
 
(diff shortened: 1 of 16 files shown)


packages/cli/src/controller/codegen-controller.ts
```diff
@@ -4,20 +4,33 @@
...
 const MODEL_ROOT_DIR = 'src/types/models';
+const exportTypes = {
+  models: false,
+  interfaces: false,
+};
 
 // 4. Render entity data in ejs template and write it
-export async function renderTemplate(outputPath: string, templateData: ejs.Data): Promise<void> {
-  const data = await ejs.renderFile(TEMPLATE_PATH, templateData);
+export async function renderTemplate(template_path: string, outputPath: string, templateData: ejs.Data): Promise<void> {
```
ianhe8x, line 32: camel case for variable name

packages/cli/src/controller/codegen-controller.ts
```diff
@@ -26,36 +39,102 @@ export interface processedField {
...
+      await renderTemplate(INTERFACE_TEMPLATE_PATH, path.join(typesDir, `interfaces.ts`), interfaceTemplate);
+      exportTypes.interfaces = true;
+    } catch (e) {
+      throw new Error(`When render json interfaces having problems.`);
+    }
+  }
 }
 
-export function processFields(className: string, fields: GraphQLEntityField[]): processedField[] {
+export function processFields(type: string, className: string, fields: GraphQLEntityField[] | GraphQLJsonFieldType[]) {
```
ianhe8x, line 73: can't find where `type` is used

ianhe8x, line 73: `GraphQLEntityField[] | GraphQLJsonFieldType[]` should be more like 
`(GraphQLEntityField | GraphQLJsonFieldType)[]`

jiqiang90: Two types 'entity' and 'jsonField' were pass though type for throw Error after transform field type , so the error message is more accurate. Also currently only 'entity' type need to extract indexed fields.

jiqiang90: Fixed

packages/cli/src/controller/codegen-controller.ts
jiqiang90: Fixed

packages/cli/src/controller/codegen-controller.ts
```diff
@@ -28,55 +41,128 @@ export interface ProcessedField {
...
       name: field.name,
-      type: newType,
       required: !field.nullable,
-      indexed,
-      unique,
-    });
+      isArray: field.isArray,
+    } as ProcessedField;
+    if (type === 'entity' && indexFields) {
+      const [indexed, unique] = indexFields.reduce<[boolean, boolean]>(
```
ianhe8x, line 92: these several lines need to run for type: entity no matter indexFields exists or not.
though you can give indexFields a default value `[]`

packages/cli/src/controller/codegen-controller.ts
```diff
@@ -28,55 +41,128 @@ export interface ProcessedField {
...
+    try {
+      await renderTemplate(TYPES_INDEX_TEMPLATE_PATH, path.join(projectPath, TYPE_ROOT_DIR, `index.ts`), {
+        props: {
+          exportTypes,
+        },
+      });
+    } catch (e) {
+      throw new Error(`When render index in types having problems.`);
+    }
+    console.log(`* Types index generated !`);
```
ianhe8x, line 164: no console.log

packages/cli/src/template/interface.ts.ejs
```diff
@@ -0,0 +1,12 @@
+// Copyright 2020-2021 OnFinality Limited authors & contributors
```
ianhe8x, line 1: we can remove license header for code template. We may own the copyright of template, but not the code generated from the template.

packages/cli/src/template/model.ts.ejs
```diff
@@ -1,6 +1,11 @@
 // Auto-generated , DO NOT EDIT
 import {Entity} from "@subql/types";
 import assert from 'assert';
+<%if (props.importJsonInterfaces.length !== 0) { %>
+import {<% props.importJsonInterfaces.forEach(function(interface){ %>
+    <%= interface %>
```
ianhe8x, line 6: need comma to split interfaces

packages/cli/src/controller/codegen-controller.ts
```diff
@@ -28,55 +41,128 @@ export interface ProcessedField {
...
+      await renderTemplate(INTERFACE_TEMPLATE_PATH, path.join(typesDir, `interfaces.ts`), interfaceTemplate);
+      exportTypes.interfaces = true;
+    } catch (e) {
+      throw new Error(`When render json interfaces having problems.`);
+    }
+  }
 }
 
 export function processFields(
+  type: string,
```
ianhe8x, line 79: make it an enum if the only possible values are 'entity' and 'json'

packages/cli/src/controller/codegen-controller.ts
```diff
@@ -28,55 +41,128 @@ export interface ProcessedField {
...
+    const injectField = {
       name: field.name,
-      type: newType,
       required: !field.nullable,
-      indexed,
-      unique,
-    });
+      isArray: field.isArray,
+    } as ProcessedField;
+    if (type === 'entity' && indexFields) {
```
ianhe8x, line 91:
```suggestion
    if (type === 'entity') {
```

ianhe8x (changes requested):

ianhe8x (approved):

PR #289, opened 2021-04-22
get Index fields from database

### LINKED ISSUE #278: Get index field from database
Since now our indexed filed is extracted from models, but we want to get it from the entity in database. 

In store.service, need to update

`const indexed =
          modelDef.indexes.findIndex((idx) => idx.fields.includes(field)) > -1;`

jiqiang90:
Closes #278


packages/node/src/utils/schema.ts
```diff
@@ -0,0 +1,59 @@
+// Copyright 2020-2021 OnFinality Limited authors & contributors
+// SPDX-License-Identifier: Apache-2.0
+
+import { camelCase } from 'lodash';
+
+export interface indexField {
```
ianhe8x, line 6:
```suggestion
export interface IndexField {
```

packages/node/src/utils/schema.ts
```diff
@@ -0,0 +1,59 @@
...
+  schema: string,
+  entity: string,
+): Promise<indexField[]> {
+  const model = this.sequelize.model(entity);
+
+  const rows = await this.sequelize.query(
+    `select
+    n.nspname as schema_name ,
+    tab.relname as table_name,
+    '${entity}' as entity_name,
```
ianhe8x, line 25: Need a test, I think entity name is converted into underscore case when creating table

packages/node/src/utils/schema.ts
```diff
@@ -0,0 +1,59 @@
...
+from
+    pg_index idx
+    JOIN pg_class cls ON cls.oid=idx.indexrelid
+    JOIN pg_class tab ON tab.oid=idx.indrelid
+    JOIN pg_am am ON am.oid=cls.relam,
+    pg_namespace n,
+    pg_attribute a
+where
+  n.nspname = '${schema}'
+  and tab.relname = '${model.tableName}'
```
ianhe8x, line 38: Need a test to see if tableName here is underscore case

packages/node/src/utils/schema.ts
ianhe8x, line 25: why we need to pass entity into this sql

packages/node/src/utils/schema.ts
```diff
@@ -0,0 +1,59 @@
...
+    tab.relname,
+    idx.indisunique,
+    am.amname`,
+  );
+  const results = rows[0];
+  return results.map((r) => {
+    return Object.keys(r).reduce(
+      (result, key) => ({
+        ...result,
+        [camelCase(key)]: r[key],
```
ianhe8x, line 54: abstract this into a utility function

packages/node/src/utils/schema.ts
```diff
@@ -0,0 +1,59 @@
+// Copyright 2020-2021 OnFinality Limited authors & contributors
+// SPDX-License-Identifier: Apache-2.0
+
+import { camelCase } from 'lodash';
+
+export interface indexField {
+  schemaName: string;
```
ianhe8x, line 7: can't find where `schemaName` and `tablename` is used

packages/node/src/indexer/store.service.ts
```diff
@@ -120,6 +137,15 @@ export class StoreService {
     tx.afterCommit(() => (this.tx = undefined));
   }
 
+  private async getAllIndexFields(schema: string) {
+    const fields = [] as indexField[][];
```
ianhe8x, line 141:
```suggestion
    const fields: IndexField[][] = [];
```

packages/node/src/utils/object.ts
```diff
@@ -12,3 +12,15 @@ export function assign<TObject, TSource1, TSource2>(
     isUndefined(srcValue) ? objValue : srcValue,
   );
 }
+
+export function camelCaseObjectKey(object) {
```
ianhe8x, line 16: typing
```suggestion
export function camelCaseObjectKey(object: object[]) {
```

ianhe8x: or rather this util function only handle single object, leave `.map` to where it is called

ianhe8x (approved):

Issue #230, opened 2021-03-11
sandbox whitelist


ntduan:
I want to query some off-chain data while processing event data and write them together to the database. But I found that I can't call third party dependencies. I would like to know if there is a way to configure my own whitelist, or if there is another way to solve this problem

ianhe8x:
We want to make sure the indexing result is deterministic, therefore can't allow off-chain data participate the process.

I suggest you see the graphql endpoint of subquery as one of datasource of your dapp (not the only one), and combine these on-chain and off-chain data in your own application.

You can also try Apollo federation which supports some advanced usage to achieve that.

ntduan:
Ok, I will try to solve this problem externally. 

But I think there is no way to avoid off-chain data. For example, you need to get the abi of the contract to resolve the events of the contract.

ianhe8x:
Of course we will provide a solution for contracts, and I won't category abi same as other external data like token trading price. Abi is metadata of your on-chain contract, just like the additional types we ask you to provide in project.yaml. It could be even stored on-chain, I roughly remember there was a discussion around this topic.

ilionic:
@ianhe8x I am getting an error 
```
ERROR failed to index block at height 1 handleCall() VMError: The module '@polkadot/util' is not whitelisted in VM.
```
Please suggest the way calling polkadot utils.

ianhe8x:
`@polkadot/*` can be white labeled, I will add it soon.

ianhe8x:
@ilionic It turns out more difficult than I thought, whitelisting it will lead to an error 
```
/..../node_modules/@polkadot/util/index.js:3
import "./detectPackage.js";
^^^^^^

SyntaxError: Cannot use import statement outside a module
```
due to a recent change since @polkadot/api v4.x, that build esm as default format.

An alternative way we can do it is inject `@polkadot/util` from host context to global so you can have access to its functions in sandbox.

Can you provide details of your use case why any function in `@polkadot/util` is required?

ilionic:
Thanks for looking into it. I actually ended up with custom GraphQL solution due to additional indexer requirements and dependencies. Plus I need Subscribtion API. 

 > Can you provide details of your use case why any function in @polkadot/util is required?

My use case is to parse / validate certain metadata from blocks, so I need some polkadot utils like `u8aToHex, hexToString, stringToHex`

yooml:
when I want to do some data processing:
[...]

PR #292, opened 2021-04-28
Sandbox whitelist 

jiqiang90:
#230


ianhe8x (approved):
Tested, network and io related operation are still banned. Still let's be careful.

PR #294, opened 2021-04-30
Update allowing builtins

jiqiang90:


packages/node/src/indexer/sandbox.ts
```diff
@@ -23,7 +23,7 @@ const DEFAULT_OPTION: NodeVMOptions = {
   wasm: false,
   sandbox: {},
   require: {
-    builtin: ['assert'],
+    builtin: ['assert', 'buffer', 'crypto', 'util', 'events', 'path'],
```
ianhe8x, line 26: events is not supported by vm2, https://github.com/patriksimek/vm2/issues/216#issuecomment-528866482

ianhe8x: let's drop it for now.

ianhe8x (approved):

PR #304, opened 2021-05-12
Support historic api rpc

### LINKED ISSUE #302: Support custom rpc
- Support historic rpc

jiqiang90:
Closes #302

### DIFF
--- a/packages/node/src/indexer/api.service.test.ts
+++ b/packages/node/src/indexer/api.service.test.ts
@@ -260,9 +260,9 @@ describe('ApiService', () => {
     const blockhash = await api.rpc.chain.getBlockHash(4401242);
     await apiService.setBlockhash(blockhash, true);
 
-    const patchedBlock = await patchedApi.rpc.chain.getBlock();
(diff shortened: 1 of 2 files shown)


packages/node/src/indexer/api.service.test.ts
```diff
@@ -252,4 +251,40 @@ describe('ApiService', () => {
...
+
+  it('support historic api rpc', async () => {
+    const apiService = await prepareApiService();
+    const api = apiService.getApi();
+    const patchedApi = await apiService.getPatchedApi();
+
+    const blockhash = await api.rpc.chain.getBlockHash(4401242);
+    await apiService.setBlockhash(blockhash, true);
+
+    const patchedBlock = await patchedApi.rpc.chain.getBlock();
```
ianhe8x, line 263: can you improve the naming, `patchedBlock` feels like the block is patched, but actually it means the block is get from patchedApi. Even `b1`, `b2`,... is better

jiqiang90: Fixed

ianhe8x (approved):

PR #307, opened 2021-05-13
Get finalized block from api call, instead of using subscription

### LINKED ISSUE #305: Failed to get finalized block number
Currently we using api to subscribe finalized head, however it is likely api stay connected but subscription failed. We will unable to get the latest block number.

We can try to call the api to get the latest block number every 5 secs.

jiqiang90:
Closes #305

### DIFF
--- a/packages/node/src/indexer/fetch.service.ts
+++ b/packages/node/src/indexer/fetch.service.ts
@@ -74,16 +74,14 @@ export class FetchService implements OnApplicationShutdown {
   }
 
   async init(): Promise<void> {
-    this.api.on('connected', this.getFinalizedHead);
-    await this.getFinalizedHead();
+    this.api.on('connected', () => this.getFinalizedBlockHead());
+    await this.getFinalizedBlockHead();
   }
 
   @Interval(FINALIZED_BLOCK_TIME_VARIANCE * 1000)
-  async getFinalizedHead() {
+  async getFinalizedBlockHead() {
     const finalizedHead = await this.api.rpc.chain.getFinalizedHead();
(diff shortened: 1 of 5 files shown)


packages/node/src/indexer/fetch.service.ts
```diff
@@ -72,15 +74,18 @@ export class FetchService implements OnApplicationShutdown {
...
-    const subscribeHeads = () =>
-      this.api.rpc.chain.subscribeFinalizedHeads((head) => {
-        this.latestFinalizedHeight = head.number.toNumber();
-        this.eventEmitter.emit(IndexerEvent.BlockTarget, {
-          height: this.latestFinalizedHeight,
-        });
-      });
-    this.api.on('connected', subscribeHeads);
-    await subscribeHeads();
+    this.api.on('connected', () => this.getFinalizedBlockHead());
```
ianhe8x, line 77: We don't even need to listen to `connected` event, do we?

packages/node/src/indexer/fetch.service.ts
```diff
@@ -72,15 +74,18 @@ export class FetchService implements OnApplicationShutdown {
...
+    await this.getFinalizedBlockHead();
+  }
+
+  @Interval(FINALIZED_BLOCK_TIME_VARIANCE * 1000)
+  async getFinalizedBlockHead() {
+    const finalizedHead = await this.api.rpc.chain.getFinalizedHead();
```
ianhe8x, line 83: do we need to have some error handling code here?
Not sure what nest will do if error thrown here

ianhe8x (approved):

PR #319, opened 2021-05-24
Set finalized block after api ready

jiqiang90:


packages/node/src/indexer/fetch.service.ts
```diff
@@ -78,17 +78,20 @@ export class FetchService implements OnApplicationShutdown {
...
-      const finalizedHead = await this.api.rpc.chain.getFinalizedHead();
-      const finalizedBlock = await this.api.rpc.chain.getBlock(finalizedHead);
-      this.latestFinalizedHeight = finalizedBlock.block.header.number.toNumber();
-    } catch (e) {
-      logger.error(e, `Having a problem when get finalized block`);
+    if (this.api) {
+      try {
+        const finalizedHead = await this.api.rpc.chain.getFinalizedHead();
+        const finalizedBlock = await this.api.rpc.chain.getBlock(finalizedHead);
+        this.latestFinalizedHeight = finalizedBlock.block.header.number.toNumber();
```
ianhe8x, line 85: emit event only if the new finalized height is different from last

ianhe8x (approved):

Issue #186, opened 2021-02-19
Accelerate indexing via blocks/events cache


ianhe8x:
Need to look into https://github.com/paritytech/substrate-archive first

jiqiang90:
Phases:

- [ ] Use event/call map to accelerate the indexing speed
- [ ] Use specVersion map to improve fetch runtime version range
- [ ] Re-design project configurations and manifest to support multiple dictionary(maps) for different spec networks

jiqiang90:
if map query return batch size is 0 (no block height is returned), we need to find a way to update lastProcessedHeight

jiqiang90:
We have take the dictionary approach, we can close this now

PR #342, opened 2021-06-15
Indexer dictionary

jiqiang90:
- test need fix


jiqiang90:
#186

packages/common/src/dictionary/types.ts
```diff
@@ -0,0 +1,14 @@
+// Copyright 2020-2021 OnFinality Limited authors & contributors
+// SPDX-License-Identifier: Apache-2.0
+
+export type MetaData = {
```
ianhe8x, line 4: I'm wondering if this is the right place for MetaData Definition. It is used by dictionary but doesn't just describe dictionary. it is generic for a query.

packages/node/package.json
```diff
@@ -28,6 +28,7 @@
     "@subql/types": "workspace:*",
     "@willsoto/nestjs-prometheus": "^3.0.0",
     "app-module-path": "^2.2.0",
+    "axios": "^0.21.1",
```
ianhe8x, line 31: have you considered @apollo/client?

ianhe8x:
```
import {
  ApolloClient,
  InMemoryCache,
  ApolloProvider,
  useQuery,
  gql
} from "@apollo/client";
const client = new ApolloClient({
  uri: 'https://48p1r2roz4.sse.codesandbox.io',
  cache: new InMemoryCache()
});
client
  .query({
    query: gql`
      query GetRates {
        rates(currency: "USD") {
          currency
        }
      }
    `
  })
  .then(result => console.log(result));
```

packages/node/src/indexer/BlockedQueue.ts
```diff
@@ -22,10 +22,31 @@ export class BlockedQueue<T> {
     this._queue.push(item);
   }
 
+  putAll(items: T[]): void {
+    if (this._queue.length >= this._maxSize) {
+      throw new Error('BlockedQueue exceed max size');
+    }
+    this._queue = this._queue.concat(items);
```
ianhe8x, line 29:
```suggestion
       this._queue.push(...item);
```
It mutates the original object, not copy a new one

packages/node/src/indexer/BlockedQueue.ts
```diff
@@ -22,10 +22,31 @@ export class BlockedQueue<T> {
...
+  async takeAll(max?: number): Promise<T[]> {
+    while (!this.size) {
+      await delay(0.1);
+    }
+    let result;
+    if (max) {
+      result = this._queue.slice(0, max);
+      this._queue = this._queue.slice(max);
+    } else {
+      result = [...this._queue];
```
ianhe8x, line 47: return `this._queue directly` seems fine to me, since it is replaced with an empty array right after.

packages/node/src/indexer/fetch.service.ts
```diff
@@ -45,6 +58,107 @@ export class FetchService implements OnApplicationShutdown {
...
+  async getDictionary(startBlock: number) {
+    const indexFilters = this.projectIndexFilters;
+    if (
+      indexFilters.existBlockHandler ||
+      (indexFilters.existEventHandler &&
+        indexFilters.eventFilters.length === 0) ||
+      (indexFilters.existExtrinsicHandler &&
+        indexFilters.extrinsicFilters.length === 0)
+    ) {
+      this.useDictionary = false;
```
ianhe8x, line 141: this logic is duplicated here.  `this.useDictionary` is already set in constructor time

packages/node/src/indexer/fetch.service.ts
```diff
@@ -45,6 +58,107 @@ export class FetchService implements OnApplicationShutdown {
...
+    }
+    const query = this.dictionaryService.dictionaryQuery(
```
ianhe8x, line 143: `dictionaryService.dictionaryQuery` doesn't have to be exported. 
or let's say `dictionaryService.dictionaryQuery`, `this.dictionaryService.getDictionary`, and `this.dictionaryService.getBlockBatch` can be merged into one method `this.dictionaryService.getBlockBatch(startBlock, batchSize,...handlers)` we don't really need these middle results

packages/node/src/indexer/fetch.service.ts
```diff
@@ -100,41 +214,99 @@ export class FetchService implements OnApplicationShutdown {
...
   }
-
   async startLoop(initBlockHeight: number): Promise<void> {
+    await Promise.all([
+      this.setNextBlockBuffer(initBlockHeight),
+      this.setBlockBuffer(),
+    ]);
+  }
+
+  async setNextBlockBuffer(initBlockHeight: number): Promise<void> {
```
ianhe8x, line 224:
```suggestion
  async fillNextBlockBuffer(initBlockHeight: number): Promise<void> {
```

packages/node/src/indexer/fetch.service.ts
```diff
@@ -100,41 +214,99 @@ export class FetchService implements OnApplicationShutdown {
...
     if (isUndefined(this.latestProcessedHeight)) {
       this.latestProcessedHeight = initBlockHeight - 1;
     }
     await this.fetchMeta(initBlockHeight);
     // eslint-disable-next-line no-constant-condition
+    let startBlockHeight;
+    startBlockHeight = initBlockHeight;
```
ianhe8x, line 231: merge these two in one line

packages/node/src/indexer/fetch.service.ts
```diff
@@ -45,6 +58,107 @@ export class FetchService implements OnApplicationShutdown {
...
+      (this.projectIndexFilters.existExtrinsicHandler &&
+        this.projectIndexFilters.extrinsicFilters.length === 0)
+    ) {
+      return false;
+    }
+
+    return true;
+  }
+
+  async getDictionary(startBlock: number) {
```
ianhe8x, line 132: return types

packages/node/src/indexer/fetch.service.ts
```diff
@@ -100,41 +214,99 @@ export class FetchService implements OnApplicationShutdown {
...
+              `Dictionary is not matching with this network endpoint`,
+            );
+            this.useDictionary = false;
+            continue;
+          }
+          if (metaData.lastProcessedHeight < startBlockHeight) {
+            logger.warn(
+              `Dictionary indexed block is behind current indexing block height`,
+            );
+            this.useDictionary = false;
```
ianhe8x, line 256: if this is the case we would want dictionary be temporary disabled. the way you do this makes it permanant

packages/node/src/indexer/fetch.service.ts
```diff
@@ -100,41 +214,99 @@ export class FetchService implements OnApplicationShutdown {
...
+            continue;
+          }
+          if (isUndefined(batchBlocks)) {
+            await delay(1);
+            continue;
+          }
+          if (batchBlocks.length !== 0) {
```
ianhe8x, line 263: where `nextBlockBuffer.size` is compared with max?

packages/node/src/indexer/fetch.service.ts
```diff
@@ -100,41 +214,99 @@ export class FetchService implements OnApplicationShutdown {
...
+            this.nextBlockBuffer.putAll(batchBlocks);
+            this.latestPreparedHeight = batchBlocks[batchBlocks.length - 1];
```
ianhe8x, line 265: `latestPreparedHeight` doesn't change until the block is fetched into blockBuffer

packages/node/src/indexer/fetch.service.ts
```diff
@@ -24,17 +27,27 @@ export class FetchService implements OnApplicationShutdown {
   private latestProcessedHeight: number;
   private latestPreparedHeight: number;
```
ianhe8x, line 28: you might need another cursor to indicate where nextBlockBuffer is

packages/node/src/indexer/fetch.service.ts
```diff
@@ -100,41 +214,99 @@ export class FetchService implements OnApplicationShutdown {
...
+        await delay(1);
+        continue;
+      }
+      this.nextBlockBuffer.putAll(range(startHeight, endHeight));
+      this.latestPreparedHeight = endHeight;
+      startBlockHeight = Number(this.latestPreparedHeight) + 1;
+    }
+  }
+
+  async setBlockBuffer(): Promise<void> {
```
ianhe8x, line 285:
```suggestion
  async fillBlockBuffer(): Promise<void> {
```

ianhe8x (changes requested):

packages/node/src/indexer/BlockedQueue.ts
```diff
@@ -22,10 +22,31 @@ export class BlockedQueue<T> {
...
+    this._queue.push(...items);
+  }
+
   async take(): Promise<T> {
     while (!this.size) {
       await delay(0.1);
     }
     return this._queue.shift();
   }
+  async takeAll(max?: number): Promise<T[]> {
```
ianhe8x, line 38: need some tests for this

packages/node/src/indexer/dictionary.service.ts
```diff
@@ -0,0 +1,156 @@
...
+@Injectable()
+export class DictionaryService implements OnApplicationShutdown {
+  private isShutdown = false;
+
+  onApplicationShutdown(): void {
+    this.isShutdown = true;
+  }
+
+  //generate dictionary query
+  private dictionaryQuery(
```
ianhe8x, line 30: sort methods, public methods first and then protected and private

packages/node/src/indexer/dictionary.service.ts
```diff
@@ -0,0 +1,156 @@
...
+} from '@apollo/client/core';
+import { Injectable, OnApplicationShutdown } from '@nestjs/common';
+import { SubqlEventFilter, SubqlCallFilter, MetaData } from '@subql/common';
+import fetch from 'node-fetch';
+import { ProjectIndexFilters } from './types';
+
+export type Dictionary = {
+  _metadata: MetaData;
+  batchBlocks: number[];
+  specVersions: number[];
```
ianhe8x, line 18: specVersions can be organized as
```
{
'2002': [0, 1020],
// [specVersion]: [ <startHeight>, <endHeight or empty>]
}
```

packages/node/src/indexer/dictionary.service.ts
```diff
@@ -0,0 +1,156 @@
...
+    const query = this.dictionaryQuery(
+      startBlock,
+      batchSize,
+      indexFilters.existEventHandler,
+      indexFilters.existExtrinsicHandler,
+      indexFilters.eventFilters,
+      indexFilters.extrinsicFilters,
+    );
+    const client = new ApolloClient({
+      cache: new InMemoryCache(),
```
ianhe8x, line 122: do we need to enable the cache?

packages/node/src/indexer/dictionary.service.ts
```diff
@@ -0,0 +1,156 @@
...
+    return `query{${baseQuery}}`;
+  }
+
+  async getDictionary(
+    startBlock: number,
+    batchSize: number,
+    api: string,
+    indexFilters: ProjectIndexFilters,
+  ): Promise<Dictionary> {
+    let resp;
```
ianhe8x, line 112: seems you don't have to declare it using `let`

packages/node/src/indexer/dictionary.service.ts
```diff
@@ -0,0 +1,156 @@
...
+  }`;
+      baseQuery = baseQuery.concat(extrinsicQueryQuery);
+    }
+    return `query{${baseQuery}}`;
+  }
+
+  async getDictionary(
+    startBlock: number,
+    batchSize: number,
+    api: string,
```
ianhe8x, line 109: I suggest api be set when init this service, as a private property.

packages/node/src/indexer/dictionary.service.ts
```diff
@@ -0,0 +1,156 @@
...
+    indexFilters: ProjectIndexFilters,
```
ianhe8x, line 110: what if the given `indexFilters` includes blockHandler, means should'nt go with dictionary path.

packages/node/src/indexer/fetch.service.test.ts
```diff
@@ -77,6 +109,6 @@ describe('FetchService', () => {
       }
     });
     await loopPromise;
-    expect(getMetaSpy).toBeCalledTimes(2);
+    expect(getMetaSpy).toBeCalledTimes(19);
```
ianhe8x, line 112: why metadata will be called that much times?

packages/node/src/indexer/fetch.service.test.ts
```diff
@@ -50,14 +79,17 @@ describe('FetchService', () => {
     expect(getMetaSpy).toBeCalledTimes(1);
   });
 
-  it('fetch meta data twice when spec version changed in range', async () => {
-    const batchSize = 10;
+  it('fetch meta data batchsize -1 times when spec version changed in range', async () => {
```
ianhe8x, line 82: the correct behaviour is calling metadata twice

packages/node/src/indexer/fetch.service.ts
```diff
@@ -45,6 +60,77 @@ export class FetchService implements OnApplicationShutdown {
     return this.apiService.getApi();
   }
 
+  getIndexFilters(): ProjectIndexFilters {
+    const dataSources = this.project.dataSources.filter(
+      (ds) =>
+        !ds.filter?.specName ||
+        ds.filter.specName === this.api.runtimeVersion.specName.toString(),
+    );
+    const response = {
```
ianhe8x, line 69:
```suggestion
    const response: ProjectIndexFilters = {
```

packages/node/src/indexer/fetch.service.ts
```diff
@@ -45,6 +60,77 @@ export class FetchService implements OnApplicationShutdown {
...
+      existBlockHandler: false,
+      existEventHandler: false,
+      existExtrinsicHandler: false,
+      eventFilters: [],
+      extrinsicFilters: [],
+    } as ProjectIndexFilters;
```
ianhe8x, line 75:
```suggestion
    };
```

packages/node/src/indexer/fetch.service.ts
```diff
@@ -100,41 +186,106 @@ export class FetchService implements OnApplicationShutdown {
...
+          const batchBlocks = dictionary.batchBlocks;
+          //TODO
+          // const specVersionMap = dictionary.specVersions;
+          if (
+            metaData.chain !== this.api.runtimeChain.toString() ||
+            metaData.specName !== this.api.runtimeVersion.specName.toString() ||
+            metaData.genesisHash !== this.api.genesisHash.toString()
+          ) {
+            logger.warn(
+              `Dictionary is not matching with this network endpoint`,
```
ianhe8x, line 227: add `dictionary is disabled since now` or something similar

packages/node/src/indexer/fetch.service.ts
```diff
@@ -100,41 +186,106 @@ export class FetchService implements OnApplicationShutdown {
...
+            );
+            this.useDictionary = false;
+            continue;
+          }
+          if (metaData.lastProcessedHeight < startBlockHeight) {
+            logger.warn(
+              `Dictionary indexed block is behind current indexing block height`,
+            );
+            continue;
```
ianhe8x, line 236: not continue here, but switch to original fetch block way

packages/node/src/indexer/fetch.service.ts
```diff
@@ -100,41 +186,106 @@ export class FetchService implements OnApplicationShutdown {
...
     }
     await this.fetchMeta(initBlockHeight);
     // eslint-disable-next-line no-constant-condition
+    let startBlockHeight = initBlockHeight;
+
     while (!this.isShutdown) {
-      const [startBlockHeight, endBlockHeight] =
-        this.nextBlockRange(initBlockHeight) ?? [];
-      if (isUndefined(startBlockHeight)) {
+      if (this.nextBlockBuffer.size > this.bufferAllowSize) {
```
ianhe8x, line 205: There are other condition branches you need to take into consideration.
like latestProcessedHeight == latest finalised block height

packages/node/src/indexer/fetch.service.ts
```diff
@@ -100,41 +186,106 @@ export class FetchService implements OnApplicationShutdown {
...
+            continue;
+          }
+          if (metaData.lastProcessedHeight < startBlockHeight) {
+            logger.warn(
+              `Dictionary indexed block is behind current indexing block height`,
+            );
+            continue;
+          }
+          if (isUndefined(batchBlocks)) {
+            await delay(1);
```
ianhe8x, line 239: it is not going to be undefined, only empty array.
And you need to update lastBufferedHeight to metaData.lastProcessedHeight in this case

packages/node/src/indexer/fetch.service.ts
```diff
@@ -100,41 +186,106 @@ export class FetchService implements OnApplicationShutdown {
...
+            continue;
+          }
+          if (batchBlocks.length !== 0) {
+            this.nextBlockBuffer.putAll(batchBlocks);
+            this.latestBufferedHeight = batchBlocks[batchBlocks.length - 1];
+            startBlockHeight = Number(this.latestBufferedHeight) + 1;
```
ianhe8x, line 245: why convert it into Number here, rather than the place you assign new value to this.latestBufferedHeight

packages/node/src/indexer/fetch.service.ts
```diff
@@ -158,15 +309,14 @@ export class FetchService implements OnApplicationShutdown {
   private nextBlockRange(
```
ianhe8x, line 309: some checking logic here need to be pull out and share to path of dictionary

packages/node/src/indexer/indexer.manager.ts
```diff
@@ -34,6 +35,7 @@ export class IndexerManager {
     protected apiService: ApiService,
     protected storeService: StoreService,
     protected fetchService: FetchService,
+    protected dictionaryService: DictionaryService,
```
ianhe8x, line 38: where is it used?

packages/node/src/indexer/fetch.service.ts
jiqiang90: I think we can set `this.useDictionary = false;` first,  and when after process with original fetch block way, add `    this.useDictionary = this.isUseDictionary();` to reset

packages/node/src/indexer/dictionary.service.ts
jiqiang90: cache is not an optional parameter, but we can disable cache though 
`
defaultOptions: {
        watchQuery:{
          fetchPolicy: 'no-cache',
        },
        query:{
          fetchPolicy: 'no-cache'
        }`

packages/node/src/indexer/dictionary.service.spec.ts
```diff
@@ -0,0 +1,33 @@
+// Copyright 2020-2021 OnFinality Limited authors & contributors
+// SPDX-License-Identifier: Apache-2.0
+
+import { DictionaryService } from './dictionary.service';
+
+describe('DictionaryService', () => {
+  it('return dictionary query result', async () => {
```
ianhe8x, line 7: this is not a unit test, it requires external resources

ianhe8x: You can test useDictionary method here, and more combination are welcome

packages/node/src/indexer/fetch.service.spec.ts
```diff
@@ -71,52 +88,36 @@ describe('FetchService', () => {
...
-    // eslint-disable-next-line @typescript-eslint/require-await
-    fetchService.register(async (content) => {
-      if (content.block.block.header.number.toNumber() === 10) {
-        fetchService.onApplicationShutdown();
-      }
-    });
-    await loopPromise;
-  });
-
   it('load batchSize of blocks', () => {
```
ianhe8x, line 103: can you make it clear if this test uses dictionary or not?

packages/node/src/indexer/fetch.service.spec.ts
```diff
@@ -126,31 +127,39 @@ describe('FetchService', () => {
...
       new NodeConfig({ subquery: '', subqueryName: '', batchSize }),
+      project,
+      dictionaryService,
       new EventEmitter2(),
     );
     (fetchService as any).latestFinalizedHeight = 1000;
-    (fetchService as any).latestPreparedHeight = undefined;
+    (fetchService as any).latestBufferedHeight = undefined;
     (fetchService as any).latestProcessedHeight = undefined;
     expect((fetchService as any).nextBlockRange(1001)).toBeUndefined();
```
ianhe8x, line 160: I think now `nextBlockRange` won't change latestBufferedHeight and latestProcessedHeight

packages/node/src/indexer/fetch.service.test.ts
```diff
@@ -16,20 +17,48 @@ function testSubqueryProject(): SubqueryProject {
...
     },
   };
+  project.dataSources = [];
   return project;
 }
 
 jest.setTimeout(200000);
 
 describe('FetchService', () => {
+  it('loop until shutdown', async () => {
```
ianhe8x, line 27: why it is moved to integration test?

packages/node/src/indexer/dictionary.service.ts
```diff
@@ -0,0 +1,156 @@
...
+          specVersionBlockHeightSet.add(node.blockHeight);
+        }
+      }
+      const _metadata = resp.data._metadata;
+      const batchBlocks = Array.from(blockHeightSet);
+      const specVersions = Array.from(specVersionBlockHeightSet);
+
+      return { _metadata, batchBlocks, specVersions };
+    } catch (err) {
+      throw new Error(err);
```
ianhe8x, line 153: the try-catch here is for nothing

packages/node/src/indexer/BlockedQueue.spec.ts
```diff
@@ -38,4 +38,16 @@ describe('BlockedQueue', () => {
     await takePromise;
     expect(msecondTooks).toBeGreaterThanOrEqual(delay);
   });
+
+  it('block takeAll() with batchsize', async () => {
```
ianhe8x, line 42: how about takeAll without max

packages/node/src/indexer/indexer.manager.ts
```diff
@@ -13,6 +13,7 @@ import { SubqueryModel, SubqueryRepo } from '../entities';
 import { getLogger } from '../utils/logger';
 import * as SubstrateUtil from '../utils/substrate';
 import { ApiService } from './api.service';
+import { DictionaryService } from './dictionary.service';
```
ianhe8x, line 16: it's not used

packages/node/src/meta/event.listener.ts
```diff
@@ -19,6 +19,8 @@ export class MetricEventListener {
     private injectedApiConnectedMetric: Gauge<string>,
     @InjectMetric('subql_indexer_block_queue_size')
     private blockQueueSizeMetric: Gauge<string>,
+    @InjectMetric('subql_indexer_next_block_queue_size')
+    private nextBlockQueueSizeMetric: Gauge<string>,
```
ianhe8x, line 23:
```suggestion
    private blocknumberQueueSizeMetric: Gauge<string>,
```

ianhe8x (changes requested):

ianhe8x (approved):

PR #369, opened 2021-07-02
Add profiler & dictionary metrics

jiqiang90:
- add profiler to monitoring indexer performance
- add metrics to listen status of using dictionary and number of times it been skipped.

### DIFF
--- a/packages/node/src/meta/event.listener.ts
+++ b/packages/node/src/meta/event.listener.ts
@@ -11,9 +11,9 @@ import {
   TargetBlockPayload,
 } from '../indexer/events';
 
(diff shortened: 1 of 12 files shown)


packages/node/src/meta/event.listener.ts
```diff
@@ -11,6 +11,8 @@ import {
   TargetBlockPayload,
 } from '../indexer/events';
 
+let skipDictionaryCount = 0;
```
ianhe8x, line 14: I suggest moving it into MetricEventListener

ianhe8x (approved):

PR #371, opened 2021-07-05
Fix default add index on foreign field

jiqiang90:


packages/common/src/graphql/entities.ts
```diff
@@ -101,19 +108,15 @@ export function getAllEntitiesRelations(_schema: GraphQLSchema | string): GraphQ
...
             fields: [field.name],
           });
         } else if (typeString !== 'ID' && entityNameSet.includes(typeString)) {
-          newModel.indexes.push({
-            unique: indexDirectiveVal.unique,
-            fields: [`${field.name}Id`],
-            using: IndexType.HASH,
-          });
+          //is foreign key
+          continue;
```
ianhe8x, line 119: `continue` seems too dangerous be used here. You only want to skip this if block but `continue` skip the while block

ianhe8x: you can do nothing here, or just print a warning saying `@index` here is ignored

packages/common/src/graphql/entities.ts
```diff
@@ -74,6 +74,13 @@ export function getAllEntitiesRelations(_schema: GraphQLSchema | string): GraphQ
           to: typeString,
           foreignKey: `${field.name}Id`,
         } as GraphQLRelationsType);
+        if (typeString !== 'ID') {
```
ianhe8x, line 77: how is it going to be `ID` when it is included in entityNameSet

packages/common/src/graphql/entities.ts
```diff
@@ -101,18 +105,17 @@ export function getAllEntitiesRelations(_schema: GraphQLSchema | string): GraphQ
...
             unique: indexDirectiveVal.unique,
             fields: [field.name],
           });
-        } else if (typeString !== 'ID' && entityNameSet.includes(typeString)) {
-          newModel.indexes.push({
-            unique: indexDirectiveVal.unique,
-            fields: [`${field.name}Id`],
-            using: IndexType.HASH,
+        } else if (typeString !== 'ID' && entityNameSet.includes(typeString) && indexDirectiveVal.unique) {
+          newModel.indexes.map((index) => {
```
ianhe8x, line 115: we don't want to throw error when user add `@index` on a foreign key right？

ianhe8x: why use `map`, do you expect there will be multiple matches?

ianhe8x (approved):

PR #384, opened 2021-07-09
support project from js files

### LINKED ISSUE #383: support to configure with `project.js`
Allows creating a SubQuery project with default exports from `project.js`.

Benefits:
- Projects may programmatically import Substrates type definitions (e.g import from other packages)

amphineko:
Resolves #383

### DIFF
--- a/packages/node/src/configure/NodeConfig.spec.ts
+++ b/packages/node/src/configure/NodeConfig.spec.ts
@@ -29,6 +29,18 @@ describe('NodeConfig', () => {
     expect(config.configDir).toBeTruthy();
   });
 
+  it('support read from js', () => {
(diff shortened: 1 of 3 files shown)


ianhe8x:
Code of user's project are considered high risky and untrusted, therefore we restrict their execution in sandbox only, your pr expose vulnerability in the system. So not going to be accepted sadly.

ianhe8x:
After a second thought, since my concerns are mostly security concern for `require(filePath)`. If you can wrap it in the sandbox, most of them will be solved.

ianhe8x:
close it. And a reminder that we are working on https://github.com/subquery/subql/pull/495, though will keep it backward compatible for a while

PR #385, opened 2021-07-12
Update polkadot/api to 5.0.1

jiqiang90:


packages/node/src/indexer/api.service.ts
```diff
@@ -106,36 +106,38 @@ export class ApiService implements OnApplicationShutdown {
...
+    await this.patchApi();
     return this.patchedApi;
   }
 
-  private patchApi(registry?: Registry): void {
+  private async patchApi(
+    registry?: Registry,
+    blockHash?: BlockHash,
+  ): Promise<void> {
     if (registry) {
```
ianhe8x, line 117: pls add a comment, that this is for api.findCall & api.findError

ianhe8x:
Do we still need to patch `api.query.xxx.yyy.multi()` ?

ianhe8x (approved):

PR #386, opened 2021-07-12
Documentation updates

seandotau:
Updated create page + added new Tutorials section


jamesbayly:
I've made some updates @seandotau:
- I'm not a fan of the orphan examples page so moved it to the main tutorial page for now
- Added more information about each example under the topics. Make it easier for a reader to find a topic that shows an example of a learning concept that they're looking at
- I think it should be our goal to have a full tutorial for each example SubQuery project

jamesbayly (approved):

PR #392, opened 2021-07-15
Expose best block height in meta

### LINKED ISSUE #349: Shows lag of finalized block height in _Metadata query
subql only indexes finalized blocks but sometimes the lag between finalized blocks and best blocks are way larger than normal. It's good to include the best block height in the _Metadata to indicate if there's issue with block finalization in the network.

jiqiang90:
Closes #349

### DIFF
--- a/packages/node/src/indexer/fetch.service.ts
+++ b/packages/node/src/indexer/fetch.service.ts
@@ -21,7 +21,7 @@ import { IndexerEvent } from './events';
 import { BlockContent, ProjectIndexFilters } from './types';
 
 const logger = getLogger('fetch');
-const FINALIZED_BLOCK_TIME_VARIANCE = 5;
+const BLOCK_TIME_VARIANCE = 5;
 const DICTIONARY_MAX_QUERY_SIZE = 10000;
 const { argv } = getYargsOption();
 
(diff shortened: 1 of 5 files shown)


packages/node/src/indexer/fetch.service.ts
```diff
@@ -187,6 +188,18 @@ export class FetchService implements OnApplicationShutdown {
     } catch (e) {
       logger.error(e, `Having a problem when get finalized block`);
     }
+    try {
+      const bestHeader = await this.api.rpc.chain.getHeader();
```
ianhe8x, line 192: put in a separated method like `getBestBlockHead`

packages/node/src/indexer/fetch.service.ts
```diff
@@ -188,6 +189,13 @@ export class FetchService implements OnApplicationShutdown {
     } catch (e) {
       logger.error(e, `Having a problem when get finalized block`);
     }
+  }
+
+  async getBestBlockHead() {
```
ianhe8x, line 194: `@Interval(FINALIZED_BLOCK_TIME_VARIANCE * 1000)`

ianhe8x (approved):

PR #417, opened 2021-08-05
Improve data sources filtering handling and error messages

jiqiang90:


packages/node/src/indexer/indexer.manager.ts
```diff
@@ -44,6 +50,30 @@ export class IndexerManager {
     private eventEmitter: EventEmitter2,
   ) {}
 
+  filterDataSources(): SubqlRuntimeDatasource[] {
```
ianhe8x, line 53: suggest making it private

packages/node/src/indexer/indexer.manager.ts
```diff
@@ -44,6 +50,30 @@ export class IndexerManager {
...
+        !ds.filter?.specName ||
+        ds.filter.specName === this.api.runtimeVersion.specName.toString(),
+    );
+    if (dataSourcesFilteredSpecName.length === 0) {
+      logger.error(
+        `Did not find any dataSource match with network specName ${this.api.runtimeVersion.specName}`,
+      );
+      process.exit(1);
+    }
+    const dataSourcesFilteredStartBlock = dataSourcesFilteredSpecName.filter(
```
ianhe8x, line 65: we still need both checks be put together, for cases that if you check seperatedly they are all pass but together false.

packages/node/src/indexer/indexer.manager.ts
```diff
@@ -136,7 +151,7 @@ export class IndexerManager {
         // FIXME: retry before exit
         process.exit(1);
       });
-
+    this.filteredDataSources = this.filterDataSources();
```
ianhe8x, line 154: this only need to be done once i think, within the init() likely

ianhe8x (changes requested):

ianhe8x (approved):

PR #443, opened 2021-08-23
Proof of index (part 1)

### LINKED ISSUE #410: Proof of index
In the process of decentralized indexing, indexer strives to query data to obtain rewards, but it also faces the following challenges:

- Prove that the indexing of the block height has been completed
- It is also very difficult to trace the database operation of update/delete, we cannot know start from which block the indexer input incorrect data. 
- The current implementation cannot ensure the accuracy and reliability of the data.

Therefore,  we proposed a "proof of index" approach.  
(https://docs.google.com/drawings/d/1hYMt-IJDnvQFWR5iaLEMrgA9e4ksaKujZDbU7hnwZ7w/edit?usp=sharing). This method uses Merkle tree to manage the database operations in the current block of indexing, and generates a root hash, which is stored in block data. We regard the block as the leaf of the Merkle mountain range (https://github.com/mimblewimble/grin/blob/master/doc/mmr.md) , and continuously generate the MMR root and store with the block together.

- By comparing the indexer's MMR root value, we can quickly locate start from which block has a disagreement.
- According to the characteristics of the Merkel mountain range, we only need a small number of peak MMR root values to calculate the validity of a block data.
- By comparing the hashes of operations on the same block height, we can determine whether the same operation is performed by different indexers.

jiqiang90:
#410

### DIFF
--- a/packages/node/package.json
+++ b/packages/node/package.json
@@ -32,6 +32,8 @@
     "app-module-path": "^2.2.0",
     "dayjs": "^1.10.5",
     "lodash": "^4.17.21",
+    "merkle-mountain-range": "^2.0.0",
+    "merkle-tools": "^1.4.1",
     "parse-json": "^5.2.0",
     "pg": "^8.6.0",
     "pino": "^6.11.3",
--- a/packages/node/src/indexer/PoiBlock.ts
+++ b/packages/node/src/indexer/PoiBlock.ts
@@ -1,0 +1,84 @@
+// Copyright 2020-2021 OnFinality Limited authors & contributors
+// SPDX-License-Identifier: Apache-2.0
+// eslint-disable-next-line header/header
+import { u8aConcat, numberToU8a, hexToU8a, isHex } from '@polkadot/util';
+import { blake2AsU8a } from '@polkadot/util-crypto';
+import { ProofOfIndex } from './entities/Poi.entity';
+
+const poiBlockHash = (
+  id: number,
+  chainBlockHash: string | Uint8Array,
+  operationHashRoot: Uint8Array,
+  parentHash: Uint8Array,
+  projectId: string,
+): Uint8Array => {
+  if (!id || !chainBlockHash || !operationHashRoot || !projectId) {
+    throw new Error('Poof of index: can not generate block hash');
+  }
+  return blake2AsU8a(
+    u8aConcat(
+      numberToU8a(id),
+      chainBlockHash,
+      operationHashRoot,
+      Buffer.from(projectId),
+      parentHash,
+    ),
+  );
+};
+
+export class PoiBlock implements ProofOfIndex {
+  readonly id: number;
+  readonly chainBlockHash: Uint8Array;
+  readonly hash: Uint8Array;
+  readonly parentHash: Uint8Array;
+  readonly operationHashRoot: Uint8Array;
+  mmrRoot: Uint8Array;
+  readonly projectId: string;
+
+  constructor(
+    id: number,
+    chainBlockHash: Uint8Array,
+    hash: Uint8Array,
+    parentHash: Uint8Array,
+    operationHashRoot: Uint8Array,
+    projectId: string,
+  ) {
+    this.id = id;
+    this.chainBlockHash = chainBlockHash;
+    this.hash = hash;
+    this.parentHash = parentHash;
+    this.operationHashRoot = operationHashRoot;
+    this.projectId = projectId;
+  }
+
+  static create(
+    id: number,
+    chainBlockHash: string | Uint8Array,
+    operationHashRoot: Uint8Array,
+    parentHash: Uint8Array,
+    projectId: string,
+  ): PoiBlock {
+    const _poiBlockHash = poiBlockHash(
+      id,
+      chainBlockHash,
+      operationHashRoot,
+      parentHash,
+      projectId,
+    );
+    let _chainBlockHash: Uint8Array;
+    if (isHex(chainBlockHash)) {
+      _chainBlockHash = hexToU8a(chainBlockHash);
+    } else {
+      _chainBlockHash = chainBlockHash;
+    }
+    const poiBlock = new PoiBlock(
+      id,
+      _chainBlockHash,
+      _poiBlockHash,
+      parentHash,
+      operationHashRoot,
+      projectId,
+    );
+    return poiBlock;
+  }
+}
--- a/packages/node/src/indexer/StoreOperations.spec.ts
+++ b/packages/node/src/indexer/StoreOperations.spec.ts
@@ -65,49 +65,64 @@ const testOperations = [
   },
 ];
 
+const falseOperation = {
+  operation: OperationType.Remove,
+  entityType: 'StarterEntity',
+  data: {
+    id: '0x2494bd5d089cf370c366351e851755ee42e8477df1c17ea1d9c2ae94e4f77ea8',
+    field3: null,
+    field1: 41914,
+    field2: null,
+    field4: new Date('2020-05-29T13:28:36.000Z'),
+    field5: true,
+    field6: { meat: 0, fruit: { apple: 'Apple' } },
+    createdAt: new Date('2021-08-18T02:36:06.549Z'),
+    updatedAt: new Date('2021-08-18T02:36:06.549Z'),
+  },
+};
+const apple: GraphQLJsonFieldType = {
+  name: 'apple',
+  type: 'String',
+  nullable: true,
+  isArray: false,
+};
+const fruitJson: GraphQLJsonFieldType = {
+  name: 'fruit',
+  type: 'Json',
+  jsonInterface: { name: 'Fruit', fields: [apple] },
+  isArray: false,
+  nullable: true,
+};
+const meatJson: GraphQLJsonFieldType = {
+  name: 'meat',
+  type: 'Int',
+  nullable: true,
+  isArray: false,
+};
+const models = [
+  {
+    name: 'StarterEntity',
+    fields: [
+      { name: 'id', type: 'ID', isArray: false, nullable: false },
+      { name: 'field1', type: 'Int', isArray: false, nullable: false },
+      { name: 'field2', type: 'String', isArray: false, nullable: true },
+      { name: 'field3', type: 'BigInt', isArray: false, nullable: true },
+      { name: 'field4', type: 'Date', isArray: false, nullable: true },
+      { name: 'field5', type: 'Boolean', isArray: false, nullable: true },
+      {
+        name: 'field6',
+        type: 'Json',
+        jsonInterface: { name: 'Food', fields: [meatJson, fruitJson] },
+        isArray: false,
+        nullable: true,
+      },
+    ],
+    indexes: [],
+  },
+];
+
 describe('StoreOperations', () => {
   it('put operation into the merkel leaf, and generate Merkle Root, also able to reset', () => {
-    const apple: GraphQLJsonFieldType = {
-      name: 'apple',
-      type: 'String',
-      nullable: true,
-      isArray: false,
-    };
-    const fruitJson: GraphQLJsonFieldType = {
-      name: 'fruit',
-      type: 'Json',
-      jsonInterface: { name: 'Fruit', fields: [apple] },
-      isArray: false,
-      nullable: true,
-    };
-    const meatJson: GraphQLJsonFieldType = {
-      name: 'meat',
-      type: 'Int',
-      nullable: true,
-      isArray: false,
-    };
-    const models = [
-      {
-        name: 'StarterEntity',
-        fields: [
-          { name: 'id', type: 'ID', isArray: false, nullable: false },
-          { name: 'field1', type: 'Int', isArray: false, nullable: false },
-          { name: 'field2', type: 'String', isArray: false, nullable: true },
-          { name: 'field3', type: 'BigInt', isArray: false, nullable: true },
-          { name: 'field4', type: 'Date', isArray: false, nullable: true },
-          { name: 'field5', type: 'Boolean', isArray: false, nullable: true },
-          {
-            name: 'field6',
-            type: 'Json',
-            jsonInterface: { name: 'Food', fields: [meatJson, fruitJson] },
-            isArray: false,
-            nullable: true,
-          },
-        ],
-        indexes: [],
-      },
-    ];
-
     const operationStack = new StoreOperations(models);
     for (const o of testOperations) {
       operationStack.put(o.operation, o.entityType, o.data);
@@ -120,5 +135,17 @@ describe('StoreOperations', () => {
     );
     operationStack.reset();
     expect(operationStack.getOperationLeafCount()).toBe(0);
+  });
+
+  it('throw error when remove data is not string type', () => {
+    const operationStack = new StoreOperations(models);
+
+    expect(() =>
+      operationStack.put(
+        falseOperation.operation,
+        falseOperation.entityType,
+        falseOperation.data,
+      ),
+    ).toThrow(`Remove operation only accept data in string type`);
   });
 });
--- a/packages/node/src/indexer/StoreOperations.spec.ts
+++ b/packages/node/src/indexer/StoreOperations.spec.ts
@@ -1,0 +1,124 @@
+// Copyright 2020-2021 OnFinality Limited authors & contributors
+// SPDX-License-Identifier: Apache-2.0
+
+import { u8aToHex } from '@polkadot/util';
+import { GraphQLJsonFieldType } from '@subql/common/graphql/types';
+import { StoreOperations } from './StoreOperations';
+import { OperationType } from './types';
+
+const testOperations = [
+  {
+    operation: OperationType.Set,
+    entityType: 'StarterEntity',
+    data: {
+      id: '0x2494bd5d089cf370c366351e851755ee42e8477df1c17ea1d9c2ae94e4f77ea8',
+      field1: 41914,
+    },
+  },
+  {
+    operation: OperationType.Set,
+    entityType: 'StarterEntity',
+    data: {
+      id: '0x2494bd5d089cf370c366351e851755ee42e8477df1c17ea1d9c2ae94e4f77ea8',
+      field3: null,
+      field1: 41914,
+      field2: null,
+      field4: new Date('2020-05-29T13:28:36.000Z'),
+      field5: true,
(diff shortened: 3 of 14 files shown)


packages/node/src/indexer/StoreOperations.ts
```diff
@@ -0,0 +1,141 @@
...
+                dataBuffer,
+                Buffer.from(Buffer.from(fieldValue)),
+              );
+              break;
+            default:
+              //JSON
+              dataBuffer = u8aConcat(dataBuffer, Buffer.from(field.name));
+              dataBuffer = u8aConcat(
+                dataBuffer,
+                Buffer.from(JSON.stringify(fieldValue)),
```
ianhe8x, line 97: probably we should check how sequalize handle js object, and we align our code with sequlize's

packages/node/src/indexer/entities/Poi.entity.ts
```diff
@@ -0,0 +1,60 @@
...
+export function PoiFactory(sequelize: Sequelize, schema: string): PoiRepo {
+  return <PoiRepo>sequelize.define(
+    `_poi`,
+    {
+      id: {
+        type: DataTypes.INTEGER,
+        primaryKey: true,
+      },
+      chainBlockHash: {
+        type: DataTypes.STRING,
```
ianhe8x, line 29: DataTypes.BLOB

packages/node/src/indexer/StoreOperations.ts
```diff
@@ -0,0 +1,141 @@
...
+import { u8aConcat, numberToU8a, u8aToBuffer } from '@polkadot/util';
+import { blake2AsHex } from '@polkadot/util-crypto';
+
+import { GraphQLModelsType } from '@subql/common/graphql/types';
+import { Entity } from '@subql/types';
+import MerkleTools from 'merkle-tools';
+import { OperationEntity, OperationType } from './types';
+
+export class StoreOperations {
+  private _stack: OperationEntity[] = [];
```
ianhe8x, line 13: where is this used?

packages/node/src/indexer/poi.service.ts
```diff
@@ -0,0 +1,71 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+import { OnApplicationShutdown } from '@nestjs/common';
+import { u8aConcat, numberToU8a, u8aToBuffer } from '@polkadot/util';
+import { blake2AsHex } from '@polkadot/util-crypto';
+import { SubqueryProject } from '../configure/project.model';
+
+const DEFAULT_PARENT_HASH = '0xabcd';
+
+export class PoiService implements OnApplicationShutdown {
```
ianhe8x, line 11: is it stateless? it looks more like a utility

packages/node/src/indexer/store.service.ts
```diff
@@ -17,8 +18,13 @@ import {
...
   smartTags,
 } from '../utils/sync-helper';
-
+import { MetadataFactory, MetadataRepo } from './entities/Metadata.entity';
+import { PoiFactory, PoiRepo, ProofOfIndex } from './entities/Poi.entity';
+import { PoiService } from './poi.service';
+import { StoreOperations } from './StoreOperations';
+import { OperationType } from './types';
 const logger = getLogger('store');
+const NULL_MERKEL_ROOT = '0x100';
```
ianhe8x, line 27: this is not valid, it need to be align to bytes. like `0x1000` or `0x0100`

packages/node/src/indexer/indexer.manager.ts
```diff
@@ -53,6 +56,8 @@ export class IndexerManager {
     const tx = await this.sequelize.transaction();
     this.storeService.setTransaction(tx);
 
+    let PoiBlockHash: string;
```
ianhe8x, line 59: lower case

packages/node/src/indexer/poi.service.ts
ianhe8x: you probably can define a PoiBlock class and define toHash() on it

packages/node/src/indexer/entities/Poi.entity.ts
```diff
@@ -0,0 +1,60 @@
...
+  mmrRoot: string | Buffer;
+  projectId: string;
+}
+
+export type PoiRepo = typeof Model & {
+  new (values?: unknown, options?: BuildOptions): ProofOfIndex;
+};
+
+export function PoiFactory(sequelize: Sequelize, schema: string): PoiRepo {
+  return <PoiRepo>sequelize.define(
```
ianhe8x, line 21: don't forget to define indexes.
I think on `id` and `hash`

ianhe8x (changes requested):

packages/node/package.json
```diff
@@ -31,6 +31,9 @@
     "app-module-path": "^2.2.0",
     "dayjs": "^1.10.5",
     "lodash": "^4.17.21",
+    "merkle-mountain-range": "^2.0.0",
+    "merkle-tools": "^1.4.1",
+    "object-hash": "^2.2.0",
```
ianhe8x, line 36: is this the lib you think can help with generate hash for object? I can't find it been used anywhere.

packages/node/src/indexer/StoreOperations.spec.ts
```diff
@@ -74,6 +75,22 @@ describe('StoreOperations', () => {
           { name: 'field3', type: 'BigInt', isArray: false, nullable: true },
           { name: 'field4', type: 'Date', isArray: false, nullable: true },
           { name: 'field5', type: 'Boolean', isArray: false, nullable: true },
+          { name: 'field6', type: 'Food', isArray: false, nullable: true },
+        ],
+        indexes: [],
+      },
+      {
+        name: 'Food',
```
ianhe8x, line 83: is food a jsonField or entity?

packages/node/src/indexer/StoreOperations.ts
```diff
@@ -24,87 +23,56 @@ export class StoreOperations {
     let operationBuffer: Uint8Array = new Uint8Array();
     //expect output: blake2AsHex(["id","1", "field1","12300000",...]) = "0xasdasd..."
     let dataBuffer: Uint8Array = new Uint8Array();
```
ianhe8x, line 25: this is no needed anymore

packages/node/src/indexer/StoreOperations.ts
```diff
@@ -24,87 +23,56 @@ export class StoreOperations {
...
           }
         }
       }
     }
+    dataBuffer = u8aConcat(dataBuffer, ...dataBufferArray);
     operationBuffer = u8aConcat(
       operationBuffer,
+      Buffer.from(operation.operation),
+      Buffer.from(operation.entityType),
       Buffer.from(blake2AsHex(dataBuffer)),
```
ianhe8x, line 73: Can we skip hashing here? or there are other concerns?

packages/node/src/indexer/StoreOperations.ts
```diff
@@ -24,87 +23,56 @@ export class StoreOperations {
...
-              );
+              dataBufferArray.push(Buffer.from(JSON.stringify(fieldValue)));
               break;
           }
         }
       }
     }
+    dataBuffer = u8aConcat(dataBuffer, ...dataBufferArray);
     operationBuffer = u8aConcat(
       operationBuffer,
```
ianhe8x, line 70: operationBuffer was empty, so you can just change it to 
```
const operationBuffer = u8aConcat(
 Buffer.from(operation.operation),
       Buffer.from(operation.entityType),
       Buffer.from(dataBuffer),
)

```

packages/node/src/indexer/entities/Poi.entity.ts
```diff
@@ -13,8 +13,10 @@ export interface ProofOfIndex {
   projectId: string;
 }
 
+export interface PoiModel extends Model<ProofOfIndex>, ProofOfIndex {}
+
 export type PoiRepo = typeof Model & {
-  new (values?: unknown, options?: BuildOptions): ProofOfIndex;
+  new (values?: unknown, options?: BuildOptions): PoiModel;
```
ianhe8x, line 19: I don't think PoiRepo will create PoiModel.

packages/node/src/indexer/entities/Poi.entity.ts
```diff
@@ -37,7 +39,7 @@ export function PoiFactory(sequelize: Sequelize, schema: string): PoiRepo {
       },
       parentHash: {
         type: DataTypes.BLOB,
-        allowNull: true,
+        allowNull: false,
```
ianhe8x, line 42: first block doesn't have parentHash, are you planning to use `0x00` for the first block?

packages/node/src/indexer/PoiBlock.ts
```diff
@@ -0,0 +1,55 @@
...
+    const poiBlock = new PoiBlock();
+    poiBlock.id = id;
+    poiBlock.chainBlockHash = chainBlockHash;
+    poiBlock.operationHashRoot = operationHashRoot;
+    poiBlock.parentHash = parentHash;
+    poiBlock.projectId = projectId;
+    return poiBlock;
+  }
+
+  get toHash(): string {
```
ianhe8x, line 35: toHash is a method, if you use getter, you can just name it after `get hash()`

packages/node/src/indexer/poi.service.ts
```diff
@@ -1,71 +1,62 @@
...
-      throw new Error('Poof of index: can not generate block hash');
+  async getLatestPoiBlockHash(schema?: string): Promise<string | Buffer> {
+    if (schema) {
+      this.poiRepo = PoiFactory(this.sequelize, schema);
+      const lastPoi = await this.poiRepo.findOne({
+        order: [['createdAt', 'DESC']],
+      });
+      if (lastPoi === null) {
+        this.latestPoiBlockHash = DEFAULT_PARENT_HASH;
+      } else if (lastPoi !== null && lastPoi.hash) {
```
ianhe8x: `poiRepo.findOne()` doesn't suppose to return `PoiBlock` object. so lastPoi.hash won't work.

packages/node/src/indexer/PoiBlock.ts
```diff
@@ -0,0 +1,55 @@
+// Copyright 2020-2021 OnFinality Limited authors & contributors
+// SPDX-License-Identifier: Apache-2.0
+// eslint-disable-next-line header/header
+import { u8aConcat, numberToU8a, u8aToBuffer } from '@polkadot/util';
+import { blake2AsHex } from '@polkadot/util-crypto';
+import { SubqueryProject } from '../configure/project.model';
+
+const DEFAULT_PARENT_HASH = '0xabcd';
```
ianhe8x, line 8: Why is this the default parent hash?

packages/node/src/indexer/poi.service.ts
```diff
@@ -1,71 +1,62 @@
...
+import { SubqueryRepo } from '../entities';
+import {
+  PoiFactory,
+  PoiRepo,
+  ProofOfIndex,
+  PoiModel,
+} from './entities/Poi.entity';
+import { PoiBlock } from './PoiBlock';
 
 const DEFAULT_PARENT_HASH = '0xabcd';
```
ianhe8x, line 10: also here

packages/node/src/indexer/poi.service.ts
```diff
@@ -1,71 +1,62 @@
...
-        this.latestParentHash ? this.latestParentHash : DEFAULT_PARENT_HASH,
-      ),
-    );
-    return blake2AsHex(buffer);
-  }
-
-  getLatestBlockHash(): string {
-    //TODO, try to get latest hash from database
-    return this.latestParentHash ? this.latestParentHash : DEFAULT_PARENT_HASH;
+    return this.latestPoiBlockHash
```
ianhe8x: I don't follow the logic here

ianhe8x (changes requested):

packages/node/src/indexer/poi.service.ts
```diff
@@ -36,24 +27,69 @@ export class PoiService implements OnApplicationShutdown {
...
+      return null;
+    } else if (lastPoi !== null && lastPoi.hash) {
+      return lastPoi.hash;
+    } else {
+      throw new Error(`Poi found but can not get latest hash`);
+    }
+  }
+
+  async fetchBlockOffsetFromDb(schema: string): Promise<number | null> {
+    this.metadataRepo = MetadataFactory(this.sequelize, schema);
```
ianhe8x, line 56: this.metadataRepo initialization should only happen once.
in the constructor() or init()

packages/node/src/indexer/poi.service.ts
```diff
@@ -36,24 +27,69 @@ export class PoiService implements OnApplicationShutdown {
...
+      : DEFAULT_PARENT_HASH;
+    this.blockOffset = await this.fetchBlockOffsetFromDb(schema);
+  }
+
+  async fetchLatestPoiBlockHashFromDb(
+    schema: string,
+  ): Promise<string | Buffer | null> {
+    this.poiRepo = PoiFactory(this.sequelize, schema);
+    const lastPoi = await this.poiRepo.findOne({
+      order: [['createdAt', 'DESC']],
```
ianhe8x, line 44: don't sort, it is too expensive and slow. You shall know the last processed height, so you can query via the height.

packages/node/src/indexer/poi.service.ts
```diff
@@ -36,24 +27,69 @@ export class PoiService implements OnApplicationShutdown {
...
+    const latestPoiBlockHash = await this.fetchLatestPoiBlockHashFromDb(schema);
+    this.latestPoiBlockHash = latestPoiBlockHash
+      ? latestPoiBlockHash
+      : DEFAULT_PARENT_HASH;
+    this.blockOffset = await this.fetchBlockOffsetFromDb(schema);
+  }
+
+  async fetchLatestPoiBlockHashFromDb(
+    schema: string,
+  ): Promise<string | Buffer | null> {
```
ianhe8x, line 41: pls tidy up return types.

packages/node/src/indexer/poi.service.ts
```diff
@@ -36,24 +27,69 @@ export class PoiService implements OnApplicationShutdown {
...
-        this.latestPoiBlockHash = DEFAULT_PARENT_HASH;
-      } else if (lastPoi !== null && lastPoi.hash) {
-        this.latestPoiBlockHash = lastPoi.hash;
+  async init(schema: string): Promise<void> {
+    this.schema = schema;
+    const latestPoiBlockHash = await this.fetchLatestPoiBlockHashFromDb(schema);
+    this.latestPoiBlockHash = latestPoiBlockHash
+      ? latestPoiBlockHash
+      : DEFAULT_PARENT_HASH;
+    this.blockOffset = await this.fetchBlockOffsetFromDb(schema);
```
ianhe8x, line 36: parallelize fetchLatestPoiBlockHashFromDb and fetchBlockOffsetFromDb

packages/node/src/indexer/poi.service.ts
```diff
@@ -36,24 +27,69 @@ export class PoiService implements OnApplicationShutdown {
...
+    } else {
+      throw new Error(`Poi found but can not get latest hash`);
+    }
+  }
+
+  async fetchBlockOffsetFromDb(schema: string): Promise<number | null> {
+    this.metadataRepo = MetadataFactory(this.sequelize, schema);
+    const blockOffset = await this.metadataRepo.findOne({
+      where: { key: 'blockOffset' },
+      order: [['createdAt', 'DESC']],
```
ianhe8x, line 59: no need to sort

packages/node/src/indexer/poi.service.ts
```diff
@@ -36,24 +27,69 @@ export class PoiService implements OnApplicationShutdown {
...
+    });
+    if (blockOffset === null) {
```
ianhe8x, line 61: when not found, are you sure it will be null not undefined?

packages/node/src/indexer/poi.service.ts
```diff
@@ -36,24 +27,69 @@ export class PoiService implements OnApplicationShutdown {
...
+          schema ? schema : this.schema,
+        );
+      }
+    }
+    return this.blockOffset;
+  }
+
+  //allow pass schema here , can be re-used for future validation
+  async getLatestPoiBlockHash(
+    schema?: string,
```
ianhe8x, line 80: In what cases, we might need to pass a schema different from this.schema?

packages/node/src/indexer/poi.service.ts
```diff
@@ -36,24 +27,69 @@ export class PoiService implements OnApplicationShutdown {
...
+  ): Promise<string | Buffer | null> {
+    if (!this.latestPoiBlockHash) {
+      if (this.schema || schema) {
+        this.latestPoiBlockHash = await this.fetchLatestPoiBlockHashFromDb(
+          schema ? schema : this.schema,
+        );
         return this.latestPoiBlockHash;
```
ianhe8x, line 87: check if the return from db is null/undefined then return DEFAULT_PARENT_HASH

packages/node/src/indexer/poi.service.ts
```diff
@@ -36,24 +27,69 @@ export class PoiService implements OnApplicationShutdown {
...
       } else {
-        throw new Error(`Poi found but can not get latest hash`);
+        throw new Error(`Failed to fetch latest Poi`);
       }
     }
-    return this.latestPoiBlockHash
-      ? this.latestPoiBlockHash
-      : DEFAULT_PARENT_HASH;
+    return this.latestPoiBlockHash;
```
ianhe8x, line 92: In what cases it will hit this line?

ianhe8x (changes requested):

ianhe8x (approved):

ianhe8x:
I'm ok with it, solve the conflicts and you are ready to go.

packages/node/src/indexer/StoreOperations.ts
```diff
@@ -18,7 +18,10 @@ export class StoreOperations {
 
   private operationEntityToUint8Array(operation: OperationEntity): Uint8Array {
     const dataBufferArray: Uint8Array[] = [];
-    if (operation.operation === OperationType.Remove) {
+    if (
+      operation.operation === OperationType.Remove &&
+      isString(operation.data)
```
ianhe8x, line 23: In which cases data is not string? If it isn't, shall we throw error?

ianhe8x (approved):

PR #454, opened 2021-08-28
Adds dictionary section to tutorials

seandotau:


docs/tutorials_examples/howto.md
```diff
@@ -152,3 +152,48 @@ If you navigate to Postgres, you should see two tables created. `public.subqueri
...
+  dictionary: https://api.subquery.network/sq/subquery/dictionary-polkadot
+```
+
+### What happens when a dictionary IS NOT used?
+
+When a dictionary is NOT used, an indexer will fetch every block data via the polkadot api according to the `batch-size` flag which is 100 by default, and place this in a buffer for processing. Later, the indexer takes all these blocks from the buffer and while processing the block data, checks whether the event and extrinsic in these blocks match the user-defined filter.
+
+### What happens when a dictionary IS used?
+
+When a dictionary IS used, the indexer will first take the event filter as parameters and merge this into a GraphQL query. It then uses the dictionary's API to obtain a list of relevant block heights only that contains the specific events. Often this is substantially less than 100 if the default is used.
```
ianhe8x, line 185: events and extrinsics

jamesbayly:
@seandotau  I've rearranged the docs slightly to better show the work you've done in the navigation menu. Can you please update /tutorials_examples/introduction.md to link off to your tutorials

ianhe8x (approved):

ianhe8x:
@jamesbayly feel free to approve the doc update PR and merge it.

Issue #464, opened 2021-09-02
subql pack and make it platform irrelevant


ianhe8x:
This task is preparing for decentralization subql services. Currently in our hosted service, we pull and compile from source code. but npm package system doesn't guarantee consistency, even when lockfile is present.

Originally we consider to use `npm pack`, but it doesn't include node_modules so still can not provide the consistency.

So the idea is 
1. we decide a platform(e.g: linux amd64) and compile the project in the docker container of this platform
2. package schema.graphql, project.yaml, package.json, dist/, node_modules into a tgz file
3. in the following ticket, we can implement a upload subcommand to upload the file to ipfs

ianhe8x:
Actually we can also consider tools that package a npm project and dependencies into single js file. Like rollup or webpack with proper configuration.

Issue #463, opened 2021-09-01
Refactor project manifest file


ianhe8x:
<del> From the view of subquery developer, even though they need an archive node when developing the project, the endpoint is not necessary bond to this subquery project. That means if he decides to use managed indexing service from 3rd party, he is not responsible to provide the chain node endpoint.</del>

Goals:
1. Separate endpoint, dictionary into runtimeConfig.yaml (and passed from cli only, with default search place)
2. Separate type definition into its own file
3. Explicitly specify target network in the manifest file
  1. use genesisHash to identify network
  2. chainName for known chains
  3. deprecate support of specName filter
4. Allow project.yaml use other file name, if a subquery project supports both polkadot and kusama, they can create project-polkadot.yaml and project-kusama.yaml.
5. Define entry of mapping function in project.yaml, decouple it from package.json

Suggest
```
# project.yaml
specVersion: "0.2.0"
description: ""
repository: "https://github.com/subquery/subql-starter"

schema:
  file: ./schema.graphql

network:
  genesisHash: '0x'
  chaintypes:
    file: ./types.json

dataSources:
  - name: main
    kind: substrate/Runtime
    startBlock: 1
    mapping:
      file: dist/index.js
      handlers:
        - handler: handleBlock
          kind: substrate/BlockHandler
        - handler: handleEvent
          kind: substrate/EventHandler
          filter: #Filter is optional
            module: balances
            method: Deposit
        - handler: handleCall
          kind: substrate/CallHandler
```

stwiname:
I think it would be good to have a `version` field to this to help decouple from `package.json`

PR #475, opened 2021-09-15
Add pack command to cli

stwiname:
Adds command `subql pack` which webpacks the project code into a single file

Fixes https://github.com/subquery/subql/issues/464


stwiname:
This somewhat conflicts with the build command and the template project https://github.com/subquery/subql-starter/blob/main/package.json#L8 scripts. Although `rm -rf dist && npm run build` will be run twice, once in the build command and when `npm pack` is called.

I think updating the build command to run webpack would also make sense here

packages/cli/src/commands/pack.ts
```diff
@@ -0,0 +1,97 @@
...
+        }
+
+        this.log('Finished building code');
+        resolve(true);
+      });
+    });
+
+    // Simplest way to bundle up the files we need into an archive
+    this.log('Packinging your SubQuery Project ...');
+    child.execSync('npm pack');
```
stwiname, line 94: I think this makes sense until https://github.com/subquery/subql/issues/463 is complete then we can reassess once we don't need package.json

ianhe8x: We can safely remove `npm pack`, which is not used anyway. the hosted service pull source code directly.

ianhe8x: 👍

ianhe8x:
> This somewhat conflicts with the build command and the template project https://github.com/subquery/subql-starter/blob/main/package.json#L8 scripts. Although `rm -rf dist && npm run build` will be run twice, once in the build command and when `npm pack` is called.
> 
> I think updating the build command to run webpack would also make sense here

They are conflict because we actually created another build approach and use that in pack command. And before we finish everything they will be both available.

ianhe8x:
image

ianhe8x:
Rename bundle to build (replace the current build)
Run `npm pack` or just `oclif-dev manifest && oclif-dev readme` in packages/cli and commit the generated README

ianhe8x (approved):

PR #485, opened 2021-09-19
Expand on cli `init` capabilities

stwiname:
Changes:

- Install dependencies by default when init a project, can be skipped through `--skip-install`
- ~~Run `codegen` when init a project, can be skipped through `--skip-codegen`~~
- Add the ability to specify a directory to init project in, same applies to `codegen` command

### DIFF
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -21,7 +21,7 @@ $ npm install -g @subql/cli
 $ subql COMMAND
 running command...
 $ subql (-v|--version|version)
-@subql/cli/0.12.1-0 linux-x64 node-v14.17.6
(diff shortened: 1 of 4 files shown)


stwiname:
I'm not sure if there's some place that would need docs or tutorials updated to match these changes

jamesbayly:
@stwiname you can add a section under /docs/references/references.md to add information about the `subql` command. 

All docs are synced from the docs folder

ianhe8x:
I think better make the flag default disabled. If dependency installation fails it won't look like `init` itself fails. 
And can you run `yarn pack` or `oclif-dev manifest && oclif-dev readme` manually under the cli package folder? So oclif will generate the updated README for cli

ianhe8x (approved):

PR #486, opened 2021-09-19
Implement `publish` command to upload project to IPFS

stwiname:
Uploads relevant project files to ipfs:

- `project.yaml`
- `package.json`
- `schema.graphql`
- file specified by main field in package.json

Depends on https://github.com/subquery/subql/pull/485


packages/cli/src/commands/publish.ts
```diff
@@ -0,0 +1,50 @@
...
+import {Command, flags} from '@oclif/command';
+import Build from './build';
+import {uploadToIpfs} from '../controller/publish-controller';
+
+export default class Publish extends Command {
+  static description = 'Upload this SubQuery project to IPFS';
+
+  static flags = {
+    location: flags.string({char: 'l', description: 'local folder'}),
+    ipfs: flags.string({description: 'IPFS gateway endpoint', default: 'http://localhost:5001/api/v0'}),
```
stwiname, line 15: This endpoint should default to a publicly accessible gateway. We probably need to host our own gateway.

packages/cli/src/controller/publish-controller.ts
```diff
@@ -0,0 +1,65 @@
...
+type AddResult = IterableType<ReturnType<IPFS.IPFSHTTPClient['addAll']>>;
+
+export async function uploadToIpfs(ipfsEndpoint: string, projectDir: string): Promise<AddResult[]> {
+  const ipfs = IPFS.create({url: ipfsEndpoint});
+
+  const files = getFiles(projectDir);
+
+  const pendingResults = ipfs.addAll(files, {
+    wrapWithDirectory: true,
+    pin: true, // Theres no guarantee the server will respect this
```
stwiname, line 31: IPFS gateways have no incentive to pin (persist) content. This could mean that the project is inaccessible on IPFS. 
There are pinning services such as [Pinata](https://www.pinata.cloud/) but they do pinning via their own api.

packages/cli/src/controller/publish-controller.test.ts
```diff
@@ -0,0 +1,56 @@
...
+    await Codegen.run(['-l', projectDir]);
+    await Build.run(['-l', projectDir]);
+  });
+
+  afterEach(() => {
+    childProcess.execSync(`rm -rf ${projectDir}`);
+  });
+
+  it('should upload appropriate files to IPFS', async () => {
+    const results = await uploadToIpfs('http://localhost:5001/api/v0', projectDir);
```
stwiname, line 43: This will fail for [this](https://github.com/subquery/subql/pull/486/files#r711821190) reason

ianhe8x:
I think this PR should depends on the rework of project manifest, and also can you have a look of the flow i drew 
image

packages/cli/src/controller/publish-controller.ts
```diff
@@ -0,0 +1,65 @@
...
+
+  for await (const res of pendingResults) {
+    results.push(res);
+  }
+
+  return results;
+}
+
+function getFiles(projectDir: string): FileObject[] {
+  const projectManifestPath = path.resolve(projectDir, 'project.yaml');
```
stwiname, line 44: Is `.yml` a supported extension?

ianhe8x: .Yml is supported by the lib, though I think we can stick to .yaml at the moment. To avoid extra work for the hosted service

ianhe8x:
Am i right that, we will need to have our pinning service running first before merging this PR. @stwiname

ianhe8x:
Also this PR depends on specVersion 0.2.0 I believe.

stwiname:
> Am i right that, we will need to have our pinning service running first before merging this PR. @stwiname

Its not strictly necessary, the user would just be required to provide their own IPFS node.

stwiname:
> Also this PR depends on specVersion 0.2.0 I believe.

Yes, I would like to rebase it with https://github.com/subquery/subql/pull/495 but its currently relying on https://github.com/subquery/subql/pull/485 in order to run this test https://github.com/subquery/subql/pull/486/files#diff-ad4a2851ac8e80a5b7f63429f9f3aa6f29377ff6b00aeab0b6dc95d953c54f1aR37

ianhe8x:
When we replace file reference to ipfs cid, will it do us good if we replace `file: /path` to `ipfs: /xxxx`

stwiname:
> When we replace file reference to ipfs cid, will it do us good if we replace `file: /path` to `ipfs: /xxxx`

I've got it so what it will do `ipfs://<cid>`. I think that's better because we could support multiple protocols if we wanted to use something other than IPFS or the filesystem. `file` could possibly be changed to something more descriptive though

PR #488, opened 2021-09-21
Proof of index part 2

### LINKED ISSUE #410: Proof of index

jiqiang90:
Closes #410


packages/node/src/indexer/mmr.service.ts
```diff
@@ -0,0 +1,159 @@
...
+    await this.updatePoiMmrRoot(poiBlock.id, mmrRoot);
+    this.mmrStartHeight = poiBlock.id + 1;
+  }
+
+  async updatePoiMmrRoot(id: number, mmrValue: Uint8Array) {
+    const poiBlock = await this.poiRepo.findByPk(id);
+    if (poiBlock.mmrRoot === null) {
+      poiBlock.mmrRoot = mmrValue;
+      await poiBlock.save();
+    } else if (u8aToHex(poiBlock.mmrRoot) !== u8aToHex(mmrValue)) {
```
jiqiang90, line 100: Need to check Uint8Array value here, without u8aTohex this report two value is not the same.

jiqiang90: use `u8aEq`

packages/node/src/indexer/mmr.service.ts
```diff
@@ -0,0 +1,149 @@
...
+import { SubqueryProject } from '../configure/project.model';
+import { getLogger } from '../utils/logger';
+import { delay } from '../utils/promise';
+import { FileBasedDb, keccak256FlyHash } from '../vendor/merkle-mountain-range';
+import { MetadataFactory, MetadataRepo } from './entities/Metadata.entity';
+import { PoiFactory, PoiRepo, ProofOfIndex } from './entities/Poi.entity';
+
+const logger = getLogger('mmr');
+const DEFAULT_WORD_SIZE = 32;
+const DEFAULT_LEAF = Buffer.from('0x000000000000000000000000000000');
```
ianhe8x, line 19:
```suggestion
const DEFAULT_LEAF = Buffer.alloc(32);
```

packages/node/src/indexer/mmr.service.ts
```diff
@@ -0,0 +1,149 @@
...
+
+  async appendMmrNode(poiBlock: ProofOfIndex): Promise<void> {
+    let mmrRoot: Uint8Array;
+    const newLeaf = poiBlock.hash;
+    if (newLeaf.length !== DEFAULT_WORD_SIZE) {
+      throw new Error(
+        `Append Mmr failed, input data length should be ${DEFAULT_WORD_SIZE}`,
+      );
+    }
+    const blockLeafHeight = poiBlock.id - this.blockOffset - 1; //leaf index alway -1
```
ianhe8x, line 71: I found the blockOffset you implemented is '0', shouldn't it be '1' ? if we made this change, do we still need `-1` here?

packages/node/src/indexer/mmr.service.ts
```diff
@@ -0,0 +1,149 @@
...
+  async syncFileBaseFromPoi(): Promise<void> {
+    while (!this.isShutdown) {
+      if (!this.poiRepo || !this.fileBasedMmr) {
+        await delay(10);
+      }
+      const poiBlocks = await this.getPoiBlocksByRange(this.mmrStartHeight);
+      if (poiBlocks.length !== 0) {
+        for (const block of poiBlocks) {
+          if (this.mmrStartHeight + 1 < block.id) {
+            const nextAppendHeight = this.mmrStartHeight + 1;
```
ianhe8x, line 108: `nextAppendHeight` is only used once, can make it inline

packages/node/src/indexer/mmr.service.ts
```diff
@@ -0,0 +1,149 @@
...
+const DEFAULT_FETCH_RANGE = 100;
+
+@Injectable()
+export class MmrService implements OnApplicationShutdown {
+  private isShutdown = false;
+  private blockOffset: number;
+  private metadataRepo: MetadataRepo;
+  private fileBasedMmr: MMR;
+  private poiRepo: PoiRepo;
+  private mmrStartHeight: number;
```
ianhe8x, line 29: can u use a better name, because `mmrStartHeight` doesn't sound like it will be changed.

packages/node/src/indexer/mmr.service.ts
```diff
@@ -0,0 +1,149 @@
...
+          poiBlock.mmrRoot,
+        )} not same as filebased mmr: ${u8aToHex(mmrValue)}`,
+      );
+    }
+  }
+
+  async syncFileBaseFromPoi(): Promise<void> {
+    while (!this.isShutdown) {
+      if (!this.poiRepo || !this.fileBasedMmr) {
+        await delay(10);
```
ianhe8x, line 102: is it actually possible?

packages/node/src/vendor/merkle-mountain-range/digests.ts
```diff
@@ -0,0 +1,42 @@
+// Copyright 2020-2021 OnFinality Limited authors & contributors
```
ianhe8x, line 1: we shouldn't claim copyright for this file I believe.

packages/node/src/vendor/merkle-mountain-range/fileBasedDb.ts
```diff
@@ -0,0 +1,162 @@
+// Copyright 2020-2021 OnFinality Limited authors & contributors
```
ianhe8x, line 1: we shouldn't claim copyright for this file I believe.
And why you adding the whole merkle-mountain-range project? we only need to make some changes in FileBasedDb I think.

ianhe8x (changes requested):

packages/node/src/indexer/mmr.service.ts
jiqiang90: I think offset to 0 is make more sense than 1. 

`poiBlock.id - this.blockOffset ` should give us the the total number of blocks in the mmr (aka, leafLength)

`blockOffset` also used in line 46, 

```
    this.mmrStartHeight =
      (await this.fileBasedMmr.getLeafLength()) + this.blockOffset;
```

When mmr service initialization, we want to repeat calculate previous mmr value for validation. Eg, last fileDb mmr block height is 100, we want it repeat generate mmr for block 100 again and compare to value in poi table.
Therefore in this case, it start from 100 + 0 (which is equal to initial start block -1) .

And If we changed the blockOffset to 1 (which is equal to initial start block), it start from 100 + 1 (offset) -1 (to previous block).

jiqiang90:
TODO:

- [ ] Folk MMR repo, fix relate bugs and test
- [ ] Test if we set mmr to nullable will not cause conflict in our current projects

packages/node/src/configure/NodeConfig.ts
```diff
@@ -136,6 +137,12 @@ export class NodeConfig implements IConfig {
     return this._config.proofOfIndex;
   }
 
+  get mmrPath(): string {
+    return this._config.mmrPath
+      ? this._config.mmrPath
+      : path.join(process.cwd(), `../../mmrs/${this.subqueryName}.mmr`);
```
ianhe8x, line 143:
```suggestion
      : path.join(process.cwd(), `.mmr/${this.subqueryName}.mmr`);
```

ianhe8x: And make sure the folder exists, otherwise mkdir first

packages/node/src/indexer/mmr.service.ts
```diff
@@ -0,0 +1,149 @@
...
+import { NodeConfig } from '../configure/NodeConfig';
+import { SubqueryProject } from '../configure/project.model';
+import { getLogger } from '../utils/logger';
+import { delay } from '../utils/promise';
+import { MetadataFactory, MetadataRepo } from './entities/Metadata.entity';
+import { PoiFactory, PoiRepo, ProofOfIndex } from './entities/Poi.entity';
+
+const logger = getLogger('mmr');
+const DEFAULT_WORD_SIZE = 32;
+const DEFAULT_LEAF = Buffer.alloc(32);
```
ianhe8x, line 22:
```suggestion
const DEFAULT_LEAF = Buffer.alloc(DEFAULT_WORD_SIZE);
```

packages/node/src/indexer/mmr.service.ts
```diff
@@ -0,0 +1,149 @@
...
+      return poiBlocks;
+    } else {
+      return [];
+    }
+  }
+
+  ensureFileBasedMmr(projectMmrPath: string) {
+    let fileBasedDb: FileBasedDb;
+    if (fs.existsSync(projectMmrPath)) {
+      fileBasedDb = FileBasedDb.open(projectMmrPath);
```
ianhe8x, line 139: the flag `as+` already cover the scenario that the file exists. I think the additional scenario need to be tend to is when the folder (not the file) doesn't exists.

PR #492, opened 2021-09-23
Substrate api throw with block height if failed fetch block

### LINKED ISSUE #469: api get block failed should print its block height

jiqiang90:
Closes #469

### DIFF
--- a/packages/node/src/utils/substrate.ts
+++ b/packages/node/src/utils/substrate.ts
@@ -23,6 +23,9 @@ import {
 } from '@subql/types';
 import { last, merge, range } from 'lodash';
 import { BlockContent } from '../indexer/types';
+import { getLogger } from './logger';
+
+const logger = getLogger('substrate');
 
 export function wrapBlock(
   signedBlock: SignedBlock,
@@ -247,8 +250,12 @@ export async function fetchBlocksRange(
 ): Promise<SignedBlock[]> {
   return Promise.all(
     range(startHeight, endHeight + 1).map(async (height) => {
-      const blockHash = await api.rpc.chain.getBlockHash(height);
-      return api.rpc.chain.getBlock(blockHash);
+      try {
+        const blockHash = await api.rpc.chain.getBlockHash(height);
+        return api.rpc.chain.getBlock(blockHash);
+      } catch (err) {
+        logger.error(err, `failed to fetch block at height ${height}`);
+      }
     }),
   );
 }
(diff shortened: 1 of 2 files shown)


packages/node/src/utils/substrate.ts
```diff
@@ -247,8 +250,12 @@ export async function fetchBlocksRange(
 ): Promise<SignedBlock[]> {
   return Promise.all(
     range(startHeight, endHeight + 1).map(async (height) => {
-      const blockHash = await api.rpc.chain.getBlockHash(height);
-      return api.rpc.chain.getBlock(blockHash);
+      try {
+        const blockHash = await api.rpc.chain.getBlockHash(height);
+        return api.rpc.chain.getBlock(blockHash);
+      } catch (err) {
+        logger.error(err, `failed to fetch block at height ${height}`);
```
stwiname, line 257: I don't know the logger behaviour, but should the error be rethrown or does the logger rethrow?

jiqiang90: Thanks, you are right, should be rethrown rather than just log it.

packages/node/src/utils/substrate.ts
```diff
@@ -247,8 +247,12 @@ export async function fetchBlocksRange(
 ): Promise<SignedBlock[]> {
   return Promise.all(
     range(startHeight, endHeight + 1).map(async (height) => {
-      const blockHash = await api.rpc.chain.getBlockHash(height);
-      return api.rpc.chain.getBlock(blockHash);
+      try {
+        const blockHash = await api.rpc.chain.getBlockHash(height);
+        return await api.rpc.chain.getBlock(blockHash);
+      } catch (err) {
+        throw new Error(`failed to fetch block at height ${height}`);
```
ianhe8x, line 254: the original error object should be re-thrown, not the one you created.

ianhe8x (changes requested):

packages/node/src/utils/substrate.ts
```diff
@@ -23,7 +23,9 @@ import {
 } from '@subql/types';
 import { last, merge, range } from 'lodash';
 import { BlockContent } from '../indexer/types';
+import { getLogger } from './logger';
 
+const logger = getLogger('substrate');
```
ianhe8x, line 28: We never used `substrate` this category before, Don't think its good idea adding another.
Can you use the same log category as fetch service?

ianhe8x (changes requested):

ianhe8x (approved):

PR #495, opened 2021-09-24
Introduce project manifest version 0.0.2

stwiname:
Addresses https://github.com/subquery/subql/issues/463

## Overview of changes:
- Define new project manifest schema
- Update validator for new schema, validate chain types
- Update CLI for new schema
  - ~~Update `init` to create a v0.0.2 project~~
  - Update validate to validate all supported versions
- Update Node for new schema
  - Rely on genesis hash for network identification
  - Add `--network-registry` flag to provide mapping from genesis hash to endpoint
  - Support new chain types in external file
  - Support new entry point for indexer sandbox

## Todo:

- [x] ~~Merge https://github.com/subquery/subql-starter/pull/30 and update `init` command once merged~~ Not merging so CLI releases before these changes are not broken
- [x] Setup and write tests for `IndexerManager`
- [ ] ~~Expand upon default values for `--network-registry` flag~~
- [x] Update docs (might be out of scope)


packages/cli/src/commands/init.ts
```diff
@@ -36,8 +36,8 @@ export default class Init extends Command {
       throw new Error(`Directory ${project.name} exists, try another project name`);
     }
     project.repository = await cli.prompt('Git repository', {required: false});
-    project.endpoint = await cli.prompt('RPC endpoint', {
-      default: 'wss://polkadot.api.onfinality.io/public-ws',
+    project.network = await cli.prompt('Netork genesis hash', {
+      default: '0x91b171bb158e2d3848fa23a9f1c25182fb8e20313b2c1eb49219da7a70ce90c3',
```
stwiname, line 40: It might be worth adding more human readable options here. Like a list of network names and the ability to add a custom genesis

packages/cli/src/controller/init-controller.ts
```diff
@@ -15,7 +15,7 @@ const STARTER_PATH = 'https://github.com/subquery/subql-starter';
 export async function createProject(localPath: string, project: ProjectSpec): Promise<void> {
   const projectPath = path.join(localPath, project.name);
   try {
-    await simpleGit().clone(STARTER_PATH, projectPath);
+    await simpleGit().clone(STARTER_PATH, projectPath, ['-b', 'v0.0.2', '--single-branch']);
```
stwiname, line 18: TODO: Remove once https://github.com/subquery/subql-starter/pull/30

stwiname: Actually its probably a good idea to never merge this branch, CLI should always point to a specific version. Merging this would be break `init` for any CLI version before this.

packages/cli/src/commands/init.ts
ianhe8x: We can ask user for an endpoint and retrieve genesishash from that endpoint and ask confirm.

packages/common/src/project/load.ts
```diff
@@ -3,16 +3,34 @@
...
+import {ChainTypes} from './models';
 import {ProjectManifestVersioned, VersionedProjectManifest} from './versioned';
 
+export function loadFromJsonOrYaml(file: string): unknown {
+  const fileInfo = path.parse(file);
+
+  const rawContent = fs.readFileSync(file, 'utf-8');
+
+  if (fileInfo.ext === '.json') {
+    return parseJson(rawContent, file);
```
ianhe8x, line 19: `yaml.load` supports json so it was actually fine.

ianhe8x:
Probably need to reconsider the version, for this scale of change, i think we can use 0.2.0

packages/node/src/indexer/api.service.ts
```diff
@@ -93,13 +92,7 @@ export class ApiService implements OnApplicationShutdown {
...
       return this.patchedApi;
     }
-    // TODO: remove once https://github.com/polkadot-js/api/pull/3949 is merged
-    const {
-      network: { endpoint },
-    } = this.project;
-    const patchedApi = endpoint.startsWith('ws')
-      ? this.getApi().clone()
-      : new ApiPromise(this.apiOption);
+    const patchedApi = this.getApi().clone();
```
ianhe8x, line 95: Need to keep it as it is for now. The release contains this PR is `@polkadot/api@6.0.1`, we are still using v5.9.1

packages/node/src/indexer/indexer.manager.ts
```diff
@@ -158,28 +177,33 @@ export class IndexerManager {
       });
     this.filteredDataSources = this.filterDataSources();
     this.fetchService.register((block) => this.indexBlock(block));
+
+    // Start VM for each data source
+    this.vms = await Promise.all(
```
ianhe8x, line 182: Can you shared IndexerSandbox across ds if their entries are same?

ianhe8x (changes requested):

ianhe8x:
@seandotau @zhex This is a quite large change for the project manifest. Your inputs are so important and welcomed.

packages/common/src/project/types.ts
```diff
@@ -3,17 +3,17 @@
 
 import {SubqlKind} from './constants';
 
-export interface IProjectManifest {
+export interface IProjectManifest<M extends SubqlMapping = SubqlMapping> {
   specVersion: string;
   description: string;
   repository: string;
-  dataSources: SubqlDataSource[];
-  schema: string;
```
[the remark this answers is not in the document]
stwiname: I have removed it here because the type has changed from `string` to `{ file: string; }`. Each of the versioned interfaces include the specific type.

I'm happy to add it back in as a union type or ensure that its mapped to `{ file: string; }`

packages/node/src/indexer/indexer.manager.ts
stwiname: Yeah, I wanted to do this too, but found it a little hard to identify a ds as going off the index doesn't work because we filter. 

I want to write some tests for this class then I will reassess sharing the VM

packages/cli/src/controller/init-controller.ts
ianhe8x: I'm thinking we leave `init` untouched. So even with this PR merged, end user still by default init a v0.0.1 project.

stwiname: Done. We also might want to offer a CLI command to update projects from 0.0.1 to 0.2.0

packages/node/src/configure/NodeConfig.ts
```diff
@@ -42,6 +46,14 @@ const DEFAULT_CONFIG = {
   indexCountLimit: 10,
   timestampField: true,
   proofOfIndex: false,
+  networkRegistry: {
+    /* TODO provide a default list */
```
stwiname, line 50: This is probably the largest change on the hosted service side. If the project is using the new manifest spec then it only provides a genesis hash, so there needs to be a mapping to the endpoints.

packages/common/src/project/types.ts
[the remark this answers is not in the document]
ianhe8x: We can even make the type of schema `GraphQLDocument` or something alike.

ianhe8x:
If there are still uncertainty in networkRegistry, we can leave it in the future PR. Currently we allow user provides overrides for endpoint, which should be enough.

stwiname:
> If there are still uncertainty in networkRegistry, we can leave it in the future PR. Currently we allow user provides overrides for endpoint, which should be enough.

I've removed the networkRegistry flag. `networkEndpoint` is required for v0.2.0 projects now and will throw an error if its not provided

docs/references/references.md
```diff
@@ -35,6 +35,10 @@ Options:
       --timestamp-field     Enable/disable created_at and updated_at in schema
                                                        [boolean] [default: true]
   -d, --network-dictionary  Specify the dictionary api for this network [string]
+      --proof-of-index      Enable/disable proof of index
+                                                      [boolean] [default: false]
+      --network-registry    JSON object mapping genesisHash of each network to
```
ianhe8x, line 40: update here since it has been removed

packages/cli/src/commands/init.ts
```diff
@@ -41,10 +46,19 @@ export default class Init extends Command {
...
     }
     project.repository = await cli.prompt('Git repository', {required: false});
-    project.endpoint = await cli.prompt('RPC endpoint', {
-      default: 'wss://polkadot.api.onfinality.io/public-ws',
-      required: true,
-    });
+
+    if (flags.specVersion === '0.2.0') {
+      (project as ProjectSpecV0_2_0).genesisHash = await cli.prompt('Network genesis hash', {
+        default: '0x91b171bb158e2d3848fa23a9f1c25182fb8e20313b2c1eb49219da7a70ce90c3', // Polkadot
```
ianhe8x, line 52: for v0.2.0 we can still ask user for the endpoint first. with the endpoint we can
1. retrieve genesis hash and ask user to confirm it or they input another one
2. store endpoint in runtime config (is it included in v0.2.0?)

ianhe8x:
> > If there are still uncertainty in networkRegistry, we can leave it in the future PR. Currently we allow user provides overrides for endpoint, which should be enough.
> 
> I've removed the networkRegistry flag. `networkEndpoint` is required for v0.2.0 projects now and will throw an error if its not provided

Sorry I change the idea and forget to update the issue. it says `runtimeConfig.yaml only provided via cli flags`. But after second though I think we can have it in the project.yaml (or have its reference), and remove it when we convert project.yaml to deployment..

We can have this change in a separate PR.

stwiname:
I can't seem to figure out why these tests cause other tests to fail https://github.com/subquery/subql/pull/495/files#diff-07519fb2dbc5155256ed036d08aece637848c043d153094ca7dab73099e6f34dR168-R180

My only thought could be because of something due to the sandbox vm or the process being overworked trying to index even though the fetch service gets stopped.

@ianhe8x any ideas on this? At this stage I'm happy just to disable these tests

ianhe8x (approved):

PR #505, opened 2021-09-29
Update polkadot api

jiqiang90:
TODO: 

- [ ] Find a solution to disable apiAt , current used in polkadot/api decorate (https://github.com/polkadot-js/api/blob/09527202372e5d4686da55d71195d36e118262c9/packages/api/src/base/Decorate.ts#L442)
- [ ] Wait issue(https://github.com/polkadot-js/api/issues/3999) to be resolved


packages/node/src/indexer/PoiBlock.ts
```diff
@@ -68,7 +68,7 @@ export class PoiBlock implements ProofOfIndex {
     let _chainBlockHash: Uint8Array;
     if (isHex(chainBlockHash)) {
       _chainBlockHash = hexToU8a(chainBlockHash);
-    } else {
+    } else if (isU8a(chainBlockHash)) {
```
jiqiang90, line 71: fly-by fix type

stwiname:
Can this be addressed now? https://github.com/subquery/subql/blob/9d902c0867a5237479154082c76ee7624157472a/packages/node/src/indexer/api.service.ts#L96-L102

packages/node/package.json
```diff
@@ -60,7 +60,7 @@
     "@types/yargs": "^16.0.4",
     "dotenv": "^10.0.0",
     "nodemon": "^2.0.7",
-    "supertest": "^6.1.3"
+    "supertest": "^6.1.3-0"
```
ianhe8x, line 63: don't use dev version

packages/node/src/indexer/api.service.ts
```diff
@@ -134,53 +131,40 @@ export class ApiService implements OnApplicationShutdown {
       });
     }
     this.patchApiAt(this.patchedApi);
-    this.patchApiQuery(this.patchedApi);
+    if (blockHash) {
+      await this.patchApiQuery(this.patchedApi, blockHash);
+    }
     this.patchApiTx(this.patchedApi);
     this.patchApiQueryMulti(this.patchedApi);
```
ianhe8x, line 138: some of the patch items don't need to be re-patched. Like this one.

packages/node/src/indexer/api.service.ts
```diff
@@ -121,11 +115,14 @@ export class ApiService implements OnApplicationShutdown {
...
     return this.patchedApi;
   }
 
-  private patchApi(registry?: Registry): void {
+  private async patchApi(
+    registry?: Registry,
+    blockHash?: BlockHash,
+  ): Promise<void> {
     if (registry) {
       Object.defineProperty(this.patchedApi, 'registry', {
```
ianhe8x, line 127: is this still needed? I remember with `api.at` it's not.

ianhe8x: Can you add a comment once confirmed? What's the original purpose we did this here? Is it for api.findCall / api.findError/ api.findEvent ?

packages/node/src/indexer/api.service.ts
```diff
@@ -134,53 +131,40 @@ export class ApiService implements OnApplicationShutdown {
...
   ): QueryableStorageEntry<'promise' | 'rxjs', AnyTuple> {
-    const newEntryFunc = this.replaceToAtVersion(
-      original,
-      'at',
-    ) as QueryableStorageEntry<'promise' | 'rxjs', AnyTuple>;
+    const newEntryFunc = original;
     newEntryFunc.at = NOT_SUPPORT('at');
     newEntryFunc.creator = original.creator;
-    newEntryFunc.entries = this.replaceToAtVersion(original, 'entriesAt');
+    newEntryFunc.entries = original.entries;
```
ianhe8x, line 160: the var name `original` is confusing, it used to be `original` but now more like `decorated`

packages/node/src/indexer/api.service.ts
```diff
@@ -360,12 +345,30 @@ export class ApiService implements OnApplicationShutdown {
   ): StorageKey[] {
     return calls.map((callMultiArg) => {
       if (callMultiArg instanceof Array) {
-        const [storageFunc, ...args] = callMultiArg;
-        const key = new StorageKey(api.registry, storageFunc.key(...args));
+        let key: StorageKey;
+        const storageFunc = callMultiArg[0];
```
ianhe8x, line 349: what's difference between this and `const [storageFunc, ...args] = callMultiArg;`, seems the same to me.

packages/node/src/indexer/api.service.ts
```diff
@@ -360,12 +345,30 @@ export class ApiService implements OnApplicationShutdown {
...
+        if (storageFunc.creator.meta.type.isPlain) {
+          key = new StorageKey(api.registry, storageFunc.creator);
+        } else {
+          if (storageFunc.creator.meta.type.asMap.hashers.length === 1) {
+            // single arg
+            key = new StorageKey(api.registry, storageFunc.key(args[0]));
+          } else {
+            // can be array or expanded
+            // eg. args = [call,[2038, `HAGcVQikZmEEgBBaChwjTVdwdA53Qopg2AYUtqw738C5kUq`]]
+            if (args.length === 1 && args[0] instanceof Array) {
```
ianhe8x, line 360: Seems quit4e dangerous to assume you can extend args by checking the args type. What if it is meant to not be expanded but the args happens to be an array. There're plenty types in polkadot-js extend Array, like `Tuple`

ianhe8x (changes requested):
check above

packages/node/src/indexer/api.service.ts
jiqiang90: Maybe these lines can be removed now, as `const newEntryFunc = original;`

packages/node/src/indexer/api.service.test.ts
```diff
@@ -272,6 +272,16 @@ describe('ApiService', () => {
     );
   });
 
+  //TODO, test this result when https://github.com/polkadot-js/api/issues/4020 resolved.
+  it.skip('support .entries', async () => {
```
ianhe8x: can we re-enable this testcase, since the github issue has been closed

packages/node/src/indexer/api.service.ts
```diff
@@ -125,96 +118,74 @@ export class ApiService implements OnApplicationShutdown {
...
+        this.patchRxStorageEntryMulti(
+          newEntryFunc as QueryableStorageEntry<'rxjs', AnyTuple>,
         );
       }
-    }) as any;
-    newEntryFunc.range = NOT_SUPPORT('range');
-    newEntryFunc.size = this.replaceToAtVersion(original, 'sizeAt');
-    newEntryFunc.sizeAt = NOT_SUPPORT('sizeAt');
+    } else {
+      newEntryFunc.multi = ((args: unknown[]) => {
```
ianhe8x, line 171: Make use of queryMulti() in apiAt instance.
```
      newEntryFunc.multi = ((args: unknown[]) => {
return apiAt.queryMulti([newEntryFunc.creator, args])
}
```

packages/types/package.json
```diff
@@ -19,7 +19,7 @@
     "@polkadot/api": "^5"
   },
   "devDependencies": {
-    "@polkadot/api": "^5.9.1",
+    "@polkadot/api": "6.3.1",
```
ianhe8x, line 22:
```suggestion
    "@polkadot/api": "^6",
```

PR #511, opened 2021-10-04
Support running the indexer from locations other than the filesystem

stwiname:
Primarily adds support for running projects from IPFS

Depends on https://github.com/subquery/subql/pull/486

Changes:

- Move readers from `validator` package to `common` package
- Utilise readers in `node` to handle loading all necessary files (project schema, graphql, js, etc), it also unifies file loading into a single place

TODO:
- [ ] Update custom DS to load via Readers


packages/cli/src/controller/codegen-controller.ts
```diff
@@ -186,7 +187,10 @@ export async function codegen(projectPath: string): Promise<void> {
   const interfacesPath = path.join(projectPath, TYPE_ROOT_DIR, `interfaces.ts`);
   await prepareDirPath(modelDir, true);
   await prepareDirPath(interfacesPath, false);
-  const manifest = loadProjectManifest(projectPath);
+
+  const reader = await ReaderFactory.create(projectPath);
+  const manifest = parseProjectManifest(await reader.getProjectSchema());
```
ianhe8x, line 192: `reader.getProjectSchema()` seems always be used together with `parseProjectManifest()`, I think you can combine them together.

packages/node/src/utils/project.ts
```diff
@@ -62,3 +85,225 @@ export function isCustomHandler<K extends string, F>(
...
+  if (reader instanceof LocalReader) {
+    return loadChainTypes(file, root);
+  } else {
+    const res = await reader.getFile(file);
+    let raw: unknown;
+    try {
+      raw = yaml.load(res);
+      return parseChainTypes(raw);
+    } catch (e) {
+      const chainTypesPath = await syncRemoteFileToLocal(
```
ianhe8x, line 263: Can you add some comments to explain what's happening?

packages/cli/src/controller/migrate-controller.ts
```diff
@@ -85,6 +86,7 @@ export async function migrate(
   const originManifestPath = path.join(projectPath, MANIFEST_PATH);
   const manifestV0_0_1 = path.join(projectPath, MANIFEST_V_0_0_1);
   const manifestV0_2_0 = path.join(projectPath, MANIFEST_V_0_2_0);
+  const reader = await ReaderFactory.create(projectPath);
```
ianhe8x, line 89: Shouldn't the reader initialized after the v0.2.0 manifest created?

packages/common/src/project/models.ts
```diff
@@ -2,6 +2,14 @@
...
 
 import {RegisteredTypes, RegistryTypes, OverrideModuleType, OverrideBundleType} from '@polkadot/types/types';
+import {
+  CustomDatasourceV0_2_0,
+  CustomDataSourceV0_2_0Impl,
+  FileType,
+  ProjectNetworkDeploymentV0_2_0,
+  RuntimeDataSourceV0_2_0,
+  RuntimeDataSourceV0_2_0Impl,
+} from '@subql/common/project/versioned';
```
ianhe8x, line 12: not used

packages/common/src/project/models.ts
```diff
@@ -30,6 +38,7 @@ import {
   IsObject,
   ValidateNested,
   IsEthereumAddress,
+  Equals,
```
ianhe8x, line 41: not used

packages/common/src/project/readers/github-reader.ts
```diff
@@ -17,25 +17,23 @@ export class GithubReader implements Reader {
     });
   }
 
+  get root(): undefined {
+    return undefined;
+  }
+
   async getPkg(): Promise<IPackageJson | undefined> {
-    return this.getFile('package.json');
+    return this.getFile('package.json') as Promise<unknown> as Promise<IPackageJson>;
```
ianhe8x, line 25: What if `package.json` doesn't exist?
```suggestion
    return this.getFile('package.json') as Promise<unknown> as Promise<IPackageJson | undefined>;
```

packages/common/src/project/readers/github-reader.ts
```diff
@@ -17,25 +17,23 @@ export class GithubReader implements Reader {
...
   }
 
   async getProjectSchema(): Promise<unknown | undefined> {
-    return this.getFile('project.yaml');
+    return yaml.load(await this.getFile('project.yaml'));
```
ianhe8x, line 29: same here, if project.yaml doesn't exist, will it return undefined or throw?

packages/common/src/project/readers/ipfs-reader.ts
```diff
@@ -16,22 +17,27 @@ export class IPFSReader implements Reader {
...
+    return undefined;
+  }
+
   async getPkg(): Promise<IPackageJson | undefined> {
     return Promise.resolve(undefined);
   }
 
   async getProjectSchema(): Promise<unknown | undefined> {
-    return this.getFile(this.cid);
+    return yaml.load(await this.getFile(this.cid));
```
ianhe8x, line 29: same

packages/node/package.json
```diff
@@ -36,6 +36,7 @@
     "merkle-tools": "^1.4.1",
     "parse-json": "^5.2.0",
     "pg": "^8.7.1",
+    "pino": "^6.11.3",
```
ianhe8x, line 39: pino is an indirect dependency via `@subql/common`, any reason we need to explicitly import it now?

packages/node/src/configure/SubqueryProject.ts
```diff
@@ -0,0 +1,56 @@
...
+import {
+  loadProjectFromManifest0_0_1,
+  loadProjectFromManifest0_2_0,
+} from '../utils/project';
+
+export type SubqlProjectDs = SubqlDatasource & {
+  mapping: SubqlDatasource['mapping'] & { entryScript: string };
+};
+
+export class SubqueryProject {
```
ianhe8x, line 23: When a util function dependes on some type in a module you know that something is wrong.  The dependency order shouldn't be like this.

ianhe8x: I suggest you pull loadProjectFromManifest0_0_1 and loadProjectFromManifest0_2_0 into here.
and `loadProjectFromManifest` to be renamed to `SubqueryProject.create()`

packages/node/src/indexer/api.service.spec.ts
```diff
@@ -45,18 +46,17 @@ const testNetwork: ProjectNetworkV0_0_1 = {
...
+    dataSources: [],
+    id: 'test',
+    root: './',
+    schema: new GraphQLSchema({}),
+  };
 }
 
 describe('ApiService', () => {
-  it('read custom types from project manifest', async () => {
+  it.skip('read custom types from project manifest', async () => {
```
ianhe8x, line 59: will you re-enable this test?

packages/node/src/utils/project.ts
```diff
@@ -62,3 +85,159 @@ export function isCustomHandler<K extends string, F>(
...
+  const res = content ?? (await reader.getFile(inputFile));
+  const name = blake2AsHex(inputFile).substr(2, 10);
+  const fileName = ext ? `${name}${ext}` : name;
+  const outputPath = path.resolve(root, fileName);
+  await fs.promises.writeFile(outputPath, res as string);
+  return outputPath;
+}
+
+export async function updateDataSources(
+  _dataSources:
```
ianhe8x, line 105: I suggest you make separate `updateDataSources` for each manifest versions, and put them together so next time we add a new version we know what need to implement.

packages/node/src/utils/project.ts
```diff
@@ -62,3 +85,159 @@ export function isCustomHandler<K extends string, F>(
 ): handler is SubqlCustomHandler<K, F> {
   return !isBaseHandler(handler);
 }
+
+async function syncRemoteFileToLocal(
```
ianhe8x, line 89: very confuse, why you want reader and content both as parameters?
`content: string` should cover all cases

ianhe8x: And it's not generic as you tried to make it looks like. It will be easier you just make it `syncIPFSFileToLocal`, and you can use the ipfs hash as filename directly.

packages/node/src/utils/substrate.test.ts
```diff
@@ -21,7 +21,7 @@ describe('substrate utils', () => {
 
   afterAll(() => api?.disconnect());
 
-  it('query range of blocks', async () => {
+  it.skip('query range of blocks', async () => {
```
ianhe8x, line 24: why they are all skipped?

ianhe8x (approved):

PR #521, opened 2021-10-08
Github release script and workflow

lachlan1m:


ianhe8x:
Can you add some guides how to do a release after this PR? Do we need to tag the version or the workflow will handle that?

.github/workflows/release.yml
```diff
@@ -113,3 +113,40 @@ jobs:
         run: echo "Changes exist in validator" && yarn npm publish --access public
         env:
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+
+      #Add to github releases
+      - name: Release Cli
```
ianhe8x, line 118: Can you make it `Publish Node` -> `Release Node` -> `Publish Common` -> `Release Common` ...
so that even if there's a step fails, for example `Publish Common`, the release of Node is still posted.

PR #531, opened 2021-10-14
custom datasource processors continued

stwiname:
Changes:

- Add tests for project manifest with custom ds and fix relevant issues
- Add tests for ds service and fix relevant issues
- Add test for fetch service with custom ds
- Add build command in repo root
- Fix sandbox globals not being reset if there are errors

Todo:

- [x] Increase code reuse with sandbox and ds processor

### DIFF
--- a/packages/common/src/project/models.ts
+++ b/packages/common/src/project/models.ts
@@ -29,6 +29,7 @@ import {
   IsString,
   IsObject,
   ValidateNested,
+  IsDefined,
 } from 'class-validator';
 
 export class BlockFilter implements SubqlBlockFilter {
@@ -104,6 +105,13 @@ export class EventHandler {
   handler: string;
 }
 
+export class CustomHandler implements SubqlCustomHandler {
+  @IsString()
+  kind: string;
+  @IsString()
+  handler: string;
+}
+
 export class Mapping implements SubqlMapping {
   @Transform((handlers: SubqlHandler[]) => {
     return handlers.map((handler) => {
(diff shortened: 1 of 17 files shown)


packages/node/src/indexer/ds-processor.service.ts
```diff
@@ -84,18 +56,19 @@ export class DsProcessorService {
   getDsProcessor<D extends string, T extends SubqlNetworkFilter>(
```
stwiname, line 56: I'm not sure if we should be checking that the ds is part of the project data sources from the constructor

packages/contract-processors/src/jsonfy.ts
```diff
@@ -13,9 +13,19 @@ import {
...
   validate(ds: JsonfyDatasource): void {
+    if (ds.kind !== this.kind) {
+      throw new Error('ds kind doesnt match processor');
+    }
+
+    for (const handler of ds.mapping.handlers) {
+      if (!(handler.kind in this.handlerProcessors)) {
+        throw new Error(`ds kind ${handler.kind} not one of ${Object.keys(this.handlerProcessors).join(', ')}`);
+      }
+    }
```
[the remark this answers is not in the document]
ianhe8x: Agree.

packages/node/src/indexer/ds-processor.service.ts
ianhe8x: Don't get the full context of your question. If the check is async, we can't do it in constructor though.

packages/common/src/project/models.ts
```diff
@@ -157,9 +174,11 @@ export class CustomDataSourceBase<
 > implements SubqlCustomDatasource<K, T, M>
 {
   @IsString()
+  @IsDefined()
```
ianhe8x, line 177: I think `@IsString()` covers `@IsDefined()`

packages/node/src/indexer/ds-processor.service.ts
stwiname: Its possible to do something like this:

```ts
const project = {
  ...rest,
  dataSources: [ds1, ds2],
}

const ds3 = ...;

new DsProcessorService(project).getDsProcessor(ds3);
```

ds3 has nothing to do with the project

ianhe8x (approved):
lgtm

PR #547, opened 2021-10-20
Add Moonbeam custom datasource processor

stwiname:
Demo project that records erc20 transactions for moonriver SolarBeam token: https://github.com/subquery/subql-starter/tree/moonriver-ds

Todo:
- [ ] Add processor for moonbeam call processing
- [ ] Convert substrate hashes to evm hashes

Other relevant changes:

- Improve error messages
- Fix `assets` validation/parsing for custom datasources


packages/contract-processors/src/moonbeam.ts
```diff
@@ -0,0 +1,150 @@
...
+function hexStringEq(a: string, b: string): boolean {
+  if (!isHexString(a) || !isHexString(b)) {
+    throw new Error('Inputs are not hex strings');
+  }
+  return hexStripZeros(a) === hexStripZeros(b);
+}
+
+const EventProcessor: SecondLayerHandlerProcessor<SubqlHandlerKind.Event, MoonbeamEventFilter, MoonbeamEvent> = {
+  baseFilter: [{module: 'evm', method: 'Log'}],
+  baseHandlerKind: 'substrate/EventHandler' as any /*SubqlHandlerKind.Event*/,
```
stwiname, line 85: Work around for now as importing `@subql/types` causes issues with the sandbox scope.

packages/common/src/project/models.ts
```diff
@@ -185,7 +185,7 @@ export class CustomDataSourceBase<
   startBlock?: number;
   @Type(() => FileReferenceImpl)
   @ValidateNested({each: true})
-  assets: {[p: string]: CustomDataSourceAsset};
+  assets: Map<string, CustomDataSourceAsset>;
```
ianhe8x, line 188:
```suggestion
  assets: Record<string, CustomDataSourceAsset>;
```

packages/contract-processors/src/moonbeam.ts
```diff
@@ -0,0 +1,150 @@
...
+  transformer(original: SubstrateEvent, ds: MoonbeamDatasource): Log {
+    const [eventData] = original.event.data;
+
+    const log = {
+      ...(eventData.toJSON() as unknown as RawEvent),
```
ianhe8x, line 90: Might not be the best way, though I'm also not sure user's expectation. for example for a uint256, is it string or BN users are expected to see here

ianhe8x (changes requested):
Do you have an example project i can play with?

stwiname:
> Do you have an example project i can play with?

https://github.com/subquery/subql-starter/tree/moonriver-ds

stwiname:
> Please also update the release and pre-release pipeline to include the new package.

Do you think its worth adding webpack to contract-processors to fix the `@subql/types` require issue? I'm happy to do both in another PR

packages/contract-processors/src/moonbeam.ts
```diff
@@ -0,0 +1,278 @@
...
+}
+
+export interface MoonbeamCallFilter {
+  from?: string;
+  to?: string;
+  methodId?: string;
+}
+
+export type MoonbeamEvent = Log;
+export type MoonbeamCall = Omit<TransactionResponse, 'wait'>;
```
stwiname, line 46: Currently there is some data missing that is present in the [TransactionReceipt](https://docs.ethers.io/v5/api/providers/types/#providers-TransactionReceipt). Not sure if its needed right now or can be added later

packages/common/src/project/models.ts
stwiname: Doing record/dictionary doesn't seem to be supported https://github.com/typestack/class-transformer/issues/31

I've updated the test for appropriate behaviour. 

Although the assets are never used by the Moonbeam processor so we could probably drop the field altogether

packages/contract-processors/src/moonbeam.ts
ianhe8x: What's the missing data? I believe we can add them later. Maybe after talking with Moonbeam's team to collect feedbacks from them.

ianhe8x:
type generation seems another missing pieces, probably in a separated ticket.

packages/node/src/indexer/indexer.manager.ts
```diff
@@ -345,13 +350,15 @@ export class IndexerManager {
     for (const handler of ds.mapping.handlers) {
       const processor = plugin.handlerProcessors[handler.kind];
       if (isBlockHandlerProcessor(processor)) {
-        const transformedOutput = processor.transformer(block, ds);
+        const transformedOutput = await processor.transformer(
+          block,
+          ds,
+          this.api,
```
stwiname, line 356: This exposes the whole api to the ds-sandbox which is currently needed to get the eth block hash. A user could easily modify the ds-processor and effectively have full use of the api in here then pass what they need to their project handlers.

 Not sure how to address. We could just not provide the ethereum block hash until we offer an ethereum wrapper around the api

packages/contract-processors/src/moonbeam.ts
ianhe8x: If you know what the type is, then just forcefully convert into it.

PR #532, opened 2021-10-15
Support types mapping include float

jiqiang90:
Closes #530

PR #551, opened 2021-10-21
Add support for enums and doc comments

raycar5:
This PR adds support for enums as well as using doc comments on both entities and fields.

A few catches:

- Enums don't work inside JSON fields.
- [EnumName!]! gets transformed into [EnumName]! at the postgraphile layer and I haven't found a way around it.

In it's current state it's good enough for our use-case.


ianhe8x:
Thanks for contributing, we will play with it locally and come back to you.

jiqiang90:
Hi @raycar5 thank you for the contribution. I'm start work on testing this feature, can you provide a simple subquery example with generated enum, demonstrate how to use the enum in the mapping function etc? Thank you

jiqiang90:
@raycar5 We really appreciate your contribution, and we will adopt it.  And we have some concerns here for the risks of the data, but we will continue to improve on the current implementation. 

- The interface for generated enum should use number for value instead of string , so this can be matching the implementation on the database layer.
- Also the order of the member inside the enum is important , if the order changed we should consider it is not consistent. 
We expect `codegen` generate interface be like:
```
export enum TestEnum {
    Bar ,
    Foo ,
}
```
In another pr  https://github.com/subquery/subql/pull/532 we have tidy up the types in our packages,  soon after merge yours,  we will integrate it to our changes.

raycar5:
Looking at [the postgres docs](https://www.postgresql.org/docs/9.1/datatype-enum.html) indeed it seems enum order matters so the matching should be order sensitive.

However I do not understand what you mean by "should use number for value instead of string , so this can be matching the implementation on the database layer", the database layer uses strings to interact with enum fields.

If I understand your comment correctly, you're saying that you'll pick up where I left off right? So I shouldn't push the fix for the enum order checking.

ianhe8x:
Yes,  jiqiang90 will finish the rest along with the matching fix. The consideration of using number is for shortening the hashcode for enum fields. Depends on if pg accept number for enum field, we will change the enum to number or just do that in the `toHashcode()`

ianhe8x (approved):

Issue #526, opened 2021-10-12
merge metadata from query and node


jiqiang90:
Currently , query service fetches indexer metadata by `--indexer=http://localhost:3000` and expose them

We can query indexer metadata like 

<img width="967" alt="Screen Shot 2021-10-13 at 2 57 16 PM" src="https://user-images.githubusercontent.com/10431657/137053641-d5c0860c-5e43-4ee2-8d68-858864716721.png">

But in node, we also created another metadata entity to store `blockOffset` for mmr service.
https://github.com/subquery/subql/blob/5f7d1f152b9ca138a62ac7d7acc71015dd3f1c01/packages/node/src/indexer/entities/Metadata.entity.ts#L6

Now, It is better we can merge these indexer metadata into `medadata` entity in node. Therefore, even indexer service is down, we still can query indexer metadata from db, like  `lastProcessedHeight`.

Rather than fetch meta from indexer endpoint,
https://github.com/subquery/subql/blob/5f7d1f152b9ca138a62ac7d7acc71015dd3f1c01/packages/query/src/graphql/plugins/GetMetadataPlugin.ts#L33

Now we need new method to query it directly from the `_metadata` table from db. 
The `_metadata` is a key-value table, ideally the resolver will query all data in this table and map each key to match `metadata` type

https://github.com/subquery/subql/blob/5f7d1f152b9ca138a62ac7d7acc71015dd3f1c01/packages/query/src/graphql/plugins/GetMetadataPlugin.ts#L12

PR #555, opened 2021-10-22
merge metadata from query and node

lachlan1m:
#526


packages/node/src/meta/meta.module.ts
```diff
@@ -57,8 +61,20 @@ import { MetaService } from './meta.service';
       name: 'subql_indexer_skip_dictionary_count',
       help: 'The number of times indexer been skip use dictionary',
     }),
+    {
+      provide: MetaService,
+      useFactory: async (
+        indexerManager: IndexerManager,
+        storeService: StoreService,
+      ) => {
+        await indexerManager.start();
```
lachlan1m, line 70: I need to find a better way to do this. Because currently indexerManager.start() is being called twice.

packages/query/src/graphql/plugins/GetMetadataPlugin.ts
```diff
@@ -25,27 +25,65 @@ const metaCache = {
...
+  try {
+    health = await fetch(new URL(`health`, indexerUrl));
+    metaCache.indexerHealthy = !!health.ok;
+  } catch (e) {
+    metaCache.indexerHealthy = false;
+    console.warn(`Failed to fetch indexer health, `, e.message);
+  }
+}
+
+function fetchFromTable(rows): Metadata {
```
lachlan1m, line 50: Is this ok @ianhe8x, It will prevent wrong field from being added and wrong value types.

packages/node/src/meta/meta.service.ts
```diff
@@ -47,6 +53,17 @@ export class MetaService {
...
   }
 
+  @Interval(UPDATE_HEIGHT_INTERVAL)
+  async checkTargetHeight() {
+    await Promise.all([
+      this.storeService.setMetadata(
+        'lastProcessedHeight',
+        this.lastProcessedHeight,
+      ),
+      this.storeService.setMetadata('targetHeight', this.targetHeight),
```
lachlan1m, line 63: @ianhe8x I added targetHeight here, otherwise stored in db every 5 seconds. Is this interval fine?

packages/node/src/indexer/store.service.ts
```diff
@@ -167,9 +174,30 @@ export class StoreService {
...
+        value: value,
+      },
+    });
+
+    return result;
+  }
+
+  async setMetadata(key: string, value: any): Promise<void> {
+    if (this.sequelize.isDefined(`_metadata`)) {
+      const model = this.sequelize.model('_metadata');
```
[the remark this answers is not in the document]
lachlan1m: Is this still needed, I removed my other metadata method that uses this.sequelize.model('_metadata').
If so should I create a class private property for this.sequelize.model('_poi'); where it is referenced in setPoi()

packages/node/src/indexer/indexer.manager.ts
```diff
@@ -171,17 +175,37 @@ export class IndexerManager {
...
+    if (count < 5) {
+      //block offset should only been create once, never update.
+      //if change offset will require re-index and re-sync poi
       const offsetValue = (this.getStartBlockFromDataSources() - 1).toString();
-      await metadataRepo.create({
-        key: 'blockOffset',
-        value: offsetValue,
-      });
+
+      await Promise.all([
```
lachlan1m, line 198: I could have made a method that takes a list of metadata key/value pairs to be stored. Should I save this for the next pull request adding batch calls?.

ianhe8x: I'm not so worried about batch calls.
There's still something wrong about the logic here. For `blockOffset` and `genesisHash`, they are set when empty, but don't use `count` because you must know which one of them are empty and not updating the other one.
For `indexerNodeVersion`, `chain` & `specName` they can be overwritten to different value, and you can compare first and only update when value is different.

packages/node/src/indexer/store.service.ts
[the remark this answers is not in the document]
ianhe8x: even though, `setMetadata` will be called multiple times and we don't want `this.sequelize.model('_metadata')` called that many times.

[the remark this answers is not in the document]
ianhe8x: Yes, pls do the same to setPoi

packages/node/src/meta/meta.service.ts
ianhe8x: lgtm

packages/query/src/graphql/plugins/GetMetadataPlugin.ts
ianhe8x: not it doesn't. Typescript is not strong type language, it doesn't check field or value type for you.

packages/query/src/graphql/plugins/GetMetadataPlugin.ts
```diff
@@ -45,69 +104,21 @@ export const MetadataPlugin = makeExtendSchemaPlugin((build, options) => {
...
-          const result = await context.pgClient.query(`SELECT EXISTS (
-                SELECT FROM information_schema.tables 
-                WHERE  table_schema = '${schemaName}'
-                AND    table_name   = '_metadata'
-                );`);
-
-          if (result.rows[0].exists) {
+        _metadata: async (_parentObject, _args, context, _info): Promise<Metadata> => {
+          if (metadataTableExists) {
             const {rows} = await context.pgClient.query(`select * from ${schemaName}._metadata`);
```
ianhe8x, line 109: why don't move this line into `fetchFromTable()`

packages/node/src/indexer/indexer.module.ts
```diff
@@ -19,7 +21,18 @@ import { StoreService } from './store.service';
...
     StoreService,
-    ApiService,
+    {
+      provide: ApiService,
+      useFactory: async (
+        project: SubqueryProject,
+        eventEmitter: EventEmitter2,
+      ) => {
+        const apiService = new ApiService(project, eventEmitter);
+        await apiService.init();
```
ianhe8x, line 31: make sure you remove the other place `apiService.init()` is called

packages/node/src/indexer/indexer.manager.ts
```diff
@@ -176,13 +180,24 @@ export class IndexerManager {
...
+
+    const indexerVersion = await metadataRepo.findOne({
+      where: { key: 'indexerNodeVersion' },
+    });
+
+    if (!indexerVersion) {
+      this.storeService.setMetadata('indexerNodeVersion', packageVersion);
+    }
+
+    await this.apiService.emitStoreMetadata();
```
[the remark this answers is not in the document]
lachlan1m: I cannot call this there neither meta service or store service are ready until after apiService.init()

packages/node/src/indexer/indexer.manager.ts
lachlan1m: I have updated this.

packages/query/src/graphql/plugins/GetMetadataPlugin.ts
```diff
@@ -25,27 +25,75 @@ const metaCache = {
...
+  };
+
+  const {rows} = await context.pgClient.query(`select * from ${schemaName}._metadata`);
+
+  for (const row of rows) {
+    const {key, value} = row;
+
+    if (Object.prototype.hasOwnProperty.call(metadata, key)) {
+      const field = metadata[key];
+      const typedValue = value as typeof field;
```
lachlan1m, line 70: I think this is wrong

packages/node/src/indexer/store.service.ts
```diff
@@ -174,20 +174,22 @@ export class StoreService {
...
-  async setMetadata(key: string, value: any): Promise<void> {
-    if (this.sequelize.isDefined(`_metadata`)) {
-      const model = this.sequelize.model('_metadata');
-      await model.upsert({ key, value });
-    }
+  async setMetadata(
+    key: string,
+    value: string | number | boolean,
+  ): Promise<void> {
+    const model = this.metaDataRepo;
```
ianhe8x, line 181: the variable assignment isn't necessary, just use `this.poiRepo` will be fine.

packages/node/src/indexer/indexer.manager.ts
```diff
@@ -174,6 +174,8 @@ export class IndexerManager {
   }
 
   private async ensureMetadata(schema: string) {
+    type KeyValue = { [key: string]: string | number | boolean };
```
ianhe8x, line 177: type definition shouldn't be inside a function. Move it to top level (out of class)

packages/node/src/indexer/indexer.manager.ts
```diff
@@ -198,15 +195,15 @@ export class IndexerManager {
...
+    entries.forEach((entry) => {
       const { key, value } = entry;
       keyValue[key] = value;
     });
 
-    //block offset and genesisHash should only been create once, never update
-    //if offset is changed, will require re-index and re-sync poi.
+    //blockOffset and genesisHash should only been create once, never update
+    //if blockOffset is changed, will require re-index and re-sync poi.
     if (!Object.prototype.hasOwnProperty.call(keyValue, 'blockOffset')) {
```
ianhe8x, line 207: `hasOwnProperty` only check if the property exists, but not the value of it.
You can use `!keyValue.blockOffset` for checking both the property itself and the value

packages/node/src/indexer/indexer.manager.ts
```diff
@@ -198,15 +195,15 @@ export class IndexerManager {
       raw: true,
     });
 
-    const keyValue: KeyValue = {};
+    const keyValue: { [key: string]: string | number | boolean } = {};
```
ianhe8x, line 198:
```
const keyValue: { [key: string]: string | number | boolean } = entries.reduct((acc, {key, value}) => {
  acc[key] = value;
  return acc;
})
```

packages/query/src/graphql/plugins/GetMetadataPlugin.ts
```diff
@@ -49,28 +49,31 @@ async function fetchFromApi(): Promise<void> {
...
-    lastProcessedHeight: undefined,
-    lastProcessedTimestamp: undefined,
-    targetHeight: undefined,
-    chain: undefined,
-    specName: undefined,
-    genesisHash: undefined,
-    indexerHealthy: undefined,
-    indexerNodeVersion: undefined,
-    queryNodeVersion: undefined,
+    lastProcessedHeight: 0,
```
ianhe8x, line 52: Why do you make this change? The gql schema for Metadata allows these fields to be undefined.

lachlan1m: this way I can type check what value I expect it to be when I pull it from the database.

ianhe8x: Not a good way. You can create a map to link a metadata field with the expected type in that case.

packages/query/src/graphql/plugins/GetMetadataPlugin.ts
```diff
@@ -47,20 +60,13 @@ async function fetchFromApi(): Promise<void> {
...
-    queryNodeVersion: '',
-  };
+async function fetchFromTable(pgClient: any, schemaName: string): Promise<Metadata> {
+  const metadata = {} as Metadata;
+
+  const keys = Object.keys(METADATA_TYPES);
+  const formattedKeys = `'${keys.join("','")}'`;
 
-  const {rows} = await context.pgClient.query(`select key, value from ${schemaName}._metadata`);
+  const {rows} = await pgClient.query(`select key, value from ${schemaName}._metadata WHERE key IN (${formattedKeys})`);
```
ianhe8x, line 69: use Parameterized query https://node-postgres.com/api/client

lachlan1m: managed to get it to work with `WHERE key = ANY`

packages/node/src/indexer/indexer.manager.ts
lachlan1m: `   const keyValue: { [key: string]: string | number | boolean } = {};
    entries.forEach((o) => (keyValue[o.key] = o.value));`

packages/query/src/graphql/graphql.test.ts
```diff
@@ -0,0 +1,147 @@
+// Copyright 2020-2021 OnFinality Limited authors & contributors
+// SPDX-License-Identifier: Apache-2.0
+
+import {ApolloServer, gql} from 'apollo-server-express';
+import {Pool} from 'pg';
+import {getPostGraphileBuilder} from 'postgraphile-core';
+import {plugins} from './plugins';
+
+describe('GraphqlModule', () => {
+  const pool = new Pool({
```
ianhe8x, line 10: You can read DB params from env, the same way the main application does.

packages/query/src/graphql/graphql.test.ts
```diff
@@ -0,0 +1,151 @@
...
+  afterEach(async () => {
+    await pool.query(`DROP TABLE subquery_1._metadata`);
+  });
+
+  afterAll((done) => {
+    pool.close();
+    done();
+  });
+
+  it('can query all metadata fields from database', async () => {
```
ianhe8x, line 71: maybe another test case with some more useless records in the metadata table

packages/node/src/indexer/indexer.manager.ts
```diff
@@ -193,10 +193,14 @@ export class IndexerManager {
         ],
       },
       raw: true,
-    });
+    })) as any;
```
ianhe8x, line 196: don't use any, and if you should remove `raw: true` actually.

packages/node/src/indexer/indexer.manager.ts
```diff
@@ -193,10 +193,14 @@ export class IndexerManager {
...
 
-    const keyValue: { [key: string]: string | number | boolean } = {};
-    entries.forEach((o) => (keyValue[o.key] = o.value));
+    const keyValue = entries.reduce(
+      (obj: any, e: { key: any; value: boolean | number | string }) => {
```
ianhe8x, line 199: avoid using `any`

packages/node/src/indexer/indexer.manager.ts
```diff
@@ -193,10 +193,14 @@ export class IndexerManager {
...
+        return { ...obj, [e.key]: e.value };
```
ianhe8x, line 200: avoid copying object

ianhe8x (changes requested):

packages/query/src/graphql/graphql.test.ts
```diff
@@ -4,21 +4,23 @@
 import {ApolloServer, gql} from 'apollo-server-express';
 import {Pool} from 'pg';
 import {getPostGraphileBuilder} from 'postgraphile-core';
+import {Config} from '../configure';
```
ianhe8x, line 7: 👍

packages/query/src/graphql/graphql.test.ts
lachlan1m: Added, although it only tests from the api. I have also checked to make sure useless records aren't being pulled out of db (after query).

packages/node/src/indexer/indexer.manager.ts
lachlan1m: ok I fixed and used a mapping function instead.

packages/node/src/indexer/indexer.manager.ts
```diff
@@ -127,26 +129,28 @@ export class IndexerManager {
...
     this.eventEmitter.emit(IndexerEvent.BlockLastProcessed, {
       height: blockHeight,
       timestamp: Date.now(),
     });
   }
 
   async start(): Promise<void> {
-    await this.dsProcessorService.validateCustomDs();
-    await this.apiService.init();
+    this.dsProcessorService.validateCustomDs();
```
ianhe8x, line 140: why remove `await`

ianhe8x (approved):

ianhe8x:
Congrats @lachlan1m

PR #593, opened 2021-11-01
Moonbeam EVM improvements

stwiname:
Changes:
- Add tests for `transformer` function
- Fix `MoonbeamEvent` logIndex being incorrect
- Hoist `address`/`to` field from handler filter to datasource (@jiqiang90 @lachlan1m will be relevant to docs)

Before:
```yaml
  - kind: substrate/Moonbeam
    startBlock: 752073
    processor:
      file: './node_modules/@subql/contract-processors/dist/moonbeam.js'
    # Must be a key of assets
    abi: erc20
    assets:
      erc20:
        file: './erc20.abi.json'
    mapping:
      file: './dist/index.js'
      handlers:
        - handler: handleMoonriverEvent
          kind: substrate/MoonbeamEvent
          filter:
            address: '0x6bd193ee6d2104f14f94e2ca6efefae561a4334b'
            topics:
              - Transfer(address indexed from,address indexed to,u256 value)
```

After:
```yaml
  - kind: substrate/Moonbeam
    startBlock: 752073
    processor:
      file: './node_modules/@subql/contract-processors/dist/moonbeam.js'
    # Must be a key of assets
    abi: erc20
    assets:
      erc20:
        file: './erc20.abi.json'

    # Contract address to filter, if NULL should be for contract creation
    address: '0x6bd193ee6d2104f14f94e2ca6efefae561a4334b'
    mapping:
      file: './dist/index.js'
      handlers:
        - handler: handleMoonriverEvent
          kind: substrate/MoonbeamEvent
          filter:
            topics:
              - Transfer(address indexed from,address indexed to,u256 value)
```

TODO:

- Test filtering contract creation filters


packages/common/src/project/models.ts
```diff
@@ -195,4 +196,7 @@ export class CustomDataSourceBase<
   @IsOptional()
   @IsString()
   abi: string;
+  @IsOptional()
+  @IsEthereumAddress()
+  address: string;
```
stwiname, line 201: @ianhe8x would it make sense to nest this under a `filter` key? Or maybe change abi to `abi: { name: 'erc20'; address: '0x...'}`

ianhe8x: I would say parallel to abi，so now is fine. Depends on where is is used, if you add a extra filtering step for as, then under filter looks better

packages/contract-processors/src/moonbeam.ts
```diff
@@ -33,13 +33,11 @@ type TopicFilter = string | string[] | null | undefined;
 export type MoonbeamDatasource = SubqlCustomDatasource<'substrate/Moonbeam'>;
 
 export interface MoonbeamEventFilter {
-  address?: string;
   topics?: [TopicFilter, TopicFilter, TopicFilter, TopicFilter];
 }
 
 export interface MoonbeamCallFilter {
   from?: string;
-  to?: string;
```
ianhe8x, line 42: `to` might be used for non-contract calls like transfers.

packages/contract-processors/src/moonbeam.ts
```diff
@@ -187,6 +179,14 @@ const EventProcessor: SecondLayerHandlerProcessor<SubqlHandlerKind.Event, Moonbe
   ): Promise<MoonbeamEvent> {
     const [eventData] = original.event.data;
 
+    const baseFilter = Array.isArray(EventProcessor.baseFilter)
```
ianhe8x, line 182:
```suggestion
    this.baseFilter
```

packages/types/src/project.ts
```diff
@@ -136,6 +136,7 @@ export interface SubqlCustomDatasource<
   assets: Map<string, CustomDataSourceAsset>;
   processor: FileReference;
   abi?: string; // Should be a key of assets
+  address?: string;
```
ianhe8x, line 139: for `abi` and `address` here, they are defined by CustomProcessor not aware by us.
i suggest 
```
export interface SubqlCustomDatasource<
  K extends string = string,
  T extends SubqlNetworkFilter = SubqlNetworkFilter,
  M extends SubqlMapping = SubqlMapping<SubqlCustomHandler>
> extends ISubqlDatasource<M, T> {
  kind: K;
  assets: Map<string, CustomDataSourceAsset>;
  processor: FileReference;
  [more: string]: unknown;
}
```

package.json
```diff
@@ -22,6 +22,7 @@
     "husky": "^6.0.0",
     "jest": "^26.6.3",
     "lint-staged": ">=10",
+    "moonbeam-types-bundle": "^1.2.7",
```
ianhe8x, line 25: is it used anywhere?

ianhe8x:
The other thing we discussed is to re-order filter step in front of transform.

packages/contract-processors/src/moonbeam.ts
stwiname: Yes, you can still use the address field for this use case as well

package.json
stwiname: Yes, by the new tests, not sure why its in the top level. I will move

packages/types/src/project.ts
stwiname: Its not possible to do this with class validator

ianhe8x: because of `forbidNonWhitelisted: true` or even without that?

packages/contract-processors/src/moonbeam.ts
```diff
@@ -211,21 +211,24 @@ const EventProcessor: SecondLayerHandlerProcessor<SubqlHandlerKind.Event, Moonbe
 
     return log;
   },
-  filterProcessor(filter: MoonbeamEventFilter, input: MoonbeamEvent, ds: SubqlCustomDatasource): boolean {
-    if (ds.address && !stringNormalizedEq(ds.address, input.address)) {
+  filterProcessor(filter: MoonbeamEventFilter | undefined, input: SubstrateEvent, ds: SubqlCustomDatasource): boolean {
+    const [eventData] = input.event.data;
+    const rawEvent = eventData.toJSON() as unknown as RawEvent;
```
ianhe8x, line 216: You  can use type `EthLog` from EthLog rather than toJSON

packages/contract-processors/src/moonbeam.ts
```diff
@@ -299,21 +302,35 @@ const CallProcessor: SecondLayerHandlerProcessor<SubqlHandlerKind.Call, Moonbeam
...
+      const [tx] = input.extrinsic.method.args as [EthTransaction];
+
+      // if `to` is null then we handle contract creation
+      if ((ds.address && !stringNormalizedEq(ds.address, to)) || (ds.address === null && !tx.action.isCreate)) {
+        return false;
+      }
 
-    if (filter.function && input.data.indexOf(functionToSighash(filter.function)) !== 0) {
+      const rawTx = tx.toJSON() as unknown as RawTransaction;
+      if (filter?.function && rawTx.input.indexOf(functionToSighash(filter.function)) !== 0) {
```
ianhe8x, line 325:
```suggestion
      if (filter?.function && tx.input.indexOf(functionToSighash(filter.function)) !== 0) {
```

packages/node/src/indexer/indexer.manager.ts
```diff
@@ -353,59 +356,40 @@ export class IndexerManager {
...
   ): Promise<void> {
     const plugin = this.dsProcessorService.getDsProcessor(ds);
+    const assets = await this.dsProcessorService.getAssets(ds);
+
+    const processData = async <K extends SubqlHandlerKind>(
+      processor: SecondLayerHandlerProcessor<K, unknown, unknown>,
+      handler: SubqlCustomHandler<string, Record<string, unknown>>,
+      filteredData: RuntimeHandlerInputMap[K][],
+    ): Promise<void> => {
+      const transformedData = await Promise.all(
```
ianhe8x, line 366: 👍

ianhe8x:
moonbeam-types-bundle raised minimum typescript version. Rebase main branch should fix it.

ianhe8x (approved):

PR #600, opened 2021-11-02
Mmr Init check

### LINKED ISSUE #574: improve MMR check
We need to check if current filebased db mmr is ahead of poi table

- check poi table last poi record that with mmr value
- check in filebased db mmr this record is exist, and value are same, if not throw error.

jiqiang90:
Closes #574 

Todo:

- [x] force-clean , remove the filebased db

### DIFF
--- a/packages/node/src/indexer/mmr.service.ts
+++ b/packages/node/src/indexer/mmr.service.ts
@@ -33,7 +33,7 @@ export class MmrService implements OnApplicationShutdown {
   private metadataRepo: MetadataRepo;
   private fileBasedMmr: MMR;
   private poiRepo: PoiRepo;
-  private nextMmrHeight: number;
+  private nextMmrBlockHeight: number;
 
   constructor(
     protected nodeConfig: NodeConfig,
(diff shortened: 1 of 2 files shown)


jiqiang90:
The purpose of this fix is to improve handle some extreme situations encountered during MMR calculation and to deal with these situations with min cost.

**Premise:** 

We always want to use the MMR value stored in the poi table as the reference value, which means we expect the MMR value in the poi table should be correct. Once written, it cannot be modified. 

And the file-based DB MMR can always be used to reproduce the poi MMR , and the value should remain the same. The MMRs in the file based db, it is allowed to reset/recalculate. 

We allow file-based db height behind the poi table or higher than the poi table, but the value at a certain height must be consistent with that in the poi table, otherwise, an error will be thrown.

**When working:**

-  Under normal circumstances, we will see that file db MMR is trying to catch up with the height of poi, so the height of file db generally lags behind. When catching up,  file-based db might wait for poi new records. 
-  Abnormal circumstances should consider:
    - file db MMR height is higher than the MMR height in poi table.
    - Poi table has not been recorded yet, but file db MMR already exists.

Therefore, in order to deal with this abnormal scenario, we propose a method to reset/delete the file db MMR. First, we need to extract from the poi table the height that the file db should start to calculate when it starts, which may be the last record with MMR in the poi table, or the first record without it MMR in the poi table, Or we should consider starting from 0 when there is no record.


**Discussion:** 

**Scenario 1:**

Assume the height of file based db mmr is 30000

The situation in the poi table is as follows

| id | hash | mmr |
|-------|-------|--------|
| 27575 | xxxxx | 0x1230 |
| 28175 | yyyyy | 0x1240 |
| 28637 | zzzzz | |

In this scenario, when starting mmr service, we should start from height 28175 , using file db mmr compare whether the value of 28175 height mmr matches. We use `resetFileBasedMmr(28175)`


**Scenario 2:**

Assume the height of file based db mmr is 30000

The situation in the poi table is as follows

| id | hash | mmr |
|-------|-------|--------|
| 27575 | xxxxx | |
| 28175 | yyyyy | |
| 28637 | zzzzz | |

In this scenario, we should start at 27575 when starting mmr. (or it should remove the file db complete, as we can not do validation here)



**Scenario 3:**

Assume the height of file based db mmr is 30000, but the poi table is not ready or created yet, we will use `resetFileBasedMmr(0)`

packages/node/src/indexer/mmr.service.ts
```diff
@@ -50,10 +50,27 @@ export class MmrService implements OnApplicationShutdown {
...
+    const poiWithoutMmr = await this.getFirstPoiWithoutMmr();
+    // Found the first Poi with null mmr .
+    const latestPoiWithMmr = await this.getLatestPoiWithMmr();
+
+    const poiStartHeight = latestPoiWithMmr?.id
+      ? latestPoiWithMmr?.id
+      : poiWithoutMmr.id;
+
+    if (poiStartHeight && poiStartHeight < this.nextMmrBlockHeight) {
+      await this.resetFileBasedMmr(poiStartHeight);
```
ianhe8x, line 67:
```suggestion
      await this.resetFileBasedMmr(poiStartHeight-this.blockOffset);
```

packages/node/src/indexer/mmr.service.ts
```diff
@@ -75,18 +92,17 @@ export class MmrService implements OnApplicationShutdown {
...
       );
     }
-    const blockLeafHeight = poiBlock.id - this.blockOffset - 1; //leaf index alway -1
-    const mmrNextLeafIndex = await this.fileBasedMmr.getLeafLength();
-    if (blockLeafHeight === mmrNextLeafIndex) {
-      //next mmr append leaf index should be same as block height
-      await this.fileBasedMmr.append(newLeaf, blockLeafHeight);
-      mmrRoot = await this.fileBasedMmr.getRoot(blockLeafHeight);
-    } else if (blockLeafHeight < mmrNextLeafIndex) {
+    const blockHeight = poiBlock.id - this.blockOffset;
```
ianhe8x, line 95: Can we use a different name for the height that starts from offset? so we don't get confused.
`mmrLeafIndex` or something else

packages/node/src/indexer/mmr.service.ts
```diff
@@ -75,18 +92,17 @@ export class MmrService implements OnApplicationShutdown {
...
+
+    if (blockHeight === this.nextMmrBlockHeight) {
+      await this.fileBasedMmr.append(newLeaf, blockHeight - 1);
+      mmrRoot = await this.fileBasedMmr.getRoot(blockHeight - 1);
+    } else if (blockHeight < this.nextMmrBlockHeight) {
```
ianhe8x, line 100: the blockHeight here is actually a leafIndex, so it shouldn't be used to compare with a blockHeight (includes offset)

ianhe8x (approved):

Issue #556, opened 2021-10-22
Include LICENSE in the npmjs releases


ianhe8x:
Currently LICENSE is put in the root folder and not included in sub packages' releases

PR #607, opened 2021-11-04
add license to each package

lachlan1m:
#556


ianhe8x (approved):
That works~ originally i was thinking copy the root licence file not duplicating it. but anyway.

PR #619, opened 2021-11-10
force-clean

wqsz7xn:
obviously we don't want this all in `main.ts` but not entirely sure where to put it; all other options are small configuration options so don't have dedicated files


packages/node/src/yargs.ts
```diff
@@ -29,6 +29,11 @@ export function getYargsOption() {
       demandOption: false,
       describe: 'Use local mode',
     },
+    'force-clean': {
```
stwiname, line 32: We probably want a command so its like `subql-node force-clean --subquery-name=app`

ianhe8x: i'm ok with what it is now, basically means clean data and start indexing. If we make it a subcommand, it will only be used to clean data.

wqsz7xn: good points, i was tossing up which way to implement but ended up going with how it is now because during development it's very likely that you're going to want to clean the project and then start indexing immediately (make changes to schema and then index and test on the playground)

jiqiang90:
> obviously we don't want this all in `main.ts` but not entirely sure where to put it; all other options are small configuration options so don't have dedicated files

I think rather than remove record at the very beginning, we need to validate the project. 
We can move this step to before the `apiService.init()` but after validate the custom data source, or within the `ensureProject`
https://github.com/subquery/subql/blob/main/packages/node/src/indexer/indexer.manager.ts#L141

ianhe8x (changes requested):
pls check tests, not sure why it stuck there

wqsz7xn:
My bad, didn't ping anyone when I made changes, @ianhe8x, this good?

ianhe8x (approved):

Issue #513, opened 2021-10-05
Limit query record number & depth


jiqiang90:
In our current implementation, user can query up to 100 records when have conditions like`first`, `last`,`last`... inside the filter.
https://github.com/subquery/subql/blob/c31d3e12c68517d9d40b7a05316740129dd88e2e/packages/query/src/graphql/plugins/PgConnectionArgFirstLastBeforeAfter.ts#L9

However, when without the condition filter, it seems the user allow to query an unlimited number of records, which potentially can break down our query service. We need to set this limit with/without any filter. 

For depth: 

https://www.graphile.org/postgraphile/production/#limiting-graphql-query-depth

ianhe8x:
@zhex

wqsz7xn:
assigning myself to this

ianhe8x:
@wqsz7xn  We have a customer actually do fetch whole table data, to not break their usage, I'm think adding a flag `--unsafe`.

wqsz7xn:
would it be a hack just to clamp if there are no limits? I'm still looking through the docs because i'm not very familiar w/ graphql 
```typescript
if(!after && !before && !first && !last && !offset) {
    queryBuilder.limit(MAX_RECORD_COUNT);
}
```
does work though

jiqiang90:
Feature added, close this now

jiqiang90:
Feature added, close this now

PR #621, opened 2021-11-11
Limit query record size and depth

wqsz7xn:
implements: #513
todo:
- [x] `--unsafe` flag to override (or just two different flags for depth and allowable record return size)
- [ ] add some tests for the configuration under `--unsafe` flag

discussion points
- what's a good default for depth? hardcoded to 10 at the moment
- https://github.com/stems/graphql-depth-limit


wqsz7xn:
not entirely sure on how to test this; ideas welcome. wasn't able to find tests of this sort across the repo

ianhe8x:
@wqsz7xn https://explorer.subquery.network/subquery/OnFinality-io/validator-threshold-kusama
```
query {
    validatorThresholds  {
        totalCount
        nodes {
            id
            startBlock
            timestamp
            validatorWithLeastBond
        }
    }
}
```
instead of return all records, we want to limit it to 100

packages/query/src/graphql/plugins/PgConnectionArgFirstLastBeforeAfter.ts
```diff
@@ -34,7 +39,10 @@ export default (builder) => {
 
       addArgDataGenerator(function connectionFirstLastBeforeAfter({after, before, first, last, offset}) {
         return {
-          pgQuery: (queryBuilder) => {
+          pgQuery: (queryBuilder: QueryBuilder) => {
+            if (!unsafe) {
+              queryBuilder.limit(MAX_RECORD_COUNT);
```
ianhe8x, line 44: be careful don't conflict with the rest of codes, where `queryBuilder.first(first);` is set.

ianhe8x:
> not entirely sure on how to test this; ideas welcome. wasn't able to find tests of this sort across the repo

Yes, you will need some research how to write test for your changes.

wqsz7xn:
treat this as a draft, it's very much not completed yet. having some trouble properly querying. I can create dummy rows of key value data but can't actually fetch it like I can in a normal graphql query. I'm sure this is my graphql knowledge letting me down

wqsz7xn:
@ianhe8x 
What should be the behaviour when a query has a first limit that goes above the 100 limit? Do you want me to clamp that query to the 100 items or do you want me to throw an error? I think the latter it more suitable

wqsz7xn:
Just need to test query depth and check that unsafe mode is unbounded in tests now. Shouldn't be too hard

ianhe8x:
Why you only asking for when a query has `first` limit, if we throw error, i think we should do the same for query without first

wqsz7xn:
I was just about to mention that, a query without anything (first/last) should fail throw an error if we are in safe mode?

ianhe8x:
It's not actually silent, user can compare the result size with totalCount. I think not throwing is alright.

ianhe8x:
But you remind me that we should test if they can query next page using the endCursor from the result.

packages/query/src/graphql/plugins/PgConnectionArgFirstLastBeforeAfter.ts
```diff
@@ -41,11 +41,13 @@ export default (builder) => {
         return {
           pgQuery: (queryBuilder: QueryBuilder) => {
             if (!unsafe) {
-              queryBuilder.limit(MAX_RECORD_COUNT);
+              queryBuilder.limit(MAX_ITEM_COUNT);
```
ianhe8x, line 44: what's the difference between `queryBuilder.limit` and `queryBuilder.first`?

wqsz7xn: > It's not actually silent, user can compare the result size with totalCount. I think not throwing is alright.
Alright i'll change it back so that it clamps instead

wqsz7xn:
> It's not actually silent, user can compare the result size with totalCount. I think not throwing is alright.

Alright i'll change it back so that it clamps instead

packages/query/src/graphql/plugins/PgConnectionArgFirstLastBeforeAfter.ts
wqsz7xn: i shouldn't be using `limit` here I think. I have to check what graphql defaults to for an unspecified first/last

wqsz7xn:
This was pretty poorly thought of by me, i think I could have been using complexity this whole time instead which would eliminate unforeseen circumstances/security holes as well
https://www.howtographql.com/advanced/4-security/

wqsz7xn:
@ianhe8x 
I'm honestly really stuck on this and I don't think i'm getting anywhere. I don't think this solution in this pull request is particular sturdy and there are probably a range of other ways to get around the limiting that i've put into place. Defence against this is hard when you have to think of all possible scenarios that could be attacked while your opposition (the attackers) only have to think of one you haven't thought of before. I think a far better solution to this problem would be to introduce some form of query complexity. There's a lot out there but I have come across this plug and play module that seems to do atleast part of the job. 
https://github.com/4Catalyzer/graphql-validation-complexity
The problem that i'm facing with this module so far is that I haven't found a  way to scale it based on [first/last] (if there is even a pagination limiter attached to the query) which we need because without first/last the queries are hugely expensive

wqsz7xn:
Going to split this into two tasks because I'm basically struggling to get the depth limit testing working. For now we will just implement the query entity limits and in a later pull we will do query depth limiting.
This pull uses https://www.npmjs.com/package/graphql-depth-limit as a plugin for depth limiting but I was just unable to properly test it

PR #629, opened 2021-11-15
add unsafe flag for subql/node

lachlan1m:
I have tested to make sure this flag works and with it you can for example make http requests. Although It was not so straight forward to make http requests in subql test project.

### DIFF
--- a/packages/node/src/indexer/sandbox.service.ts
+++ b/packages/node/src/indexer/sandbox.service.ts
@@ -13,8 +13,11 @@ import { SubqueryProject } from '../configure/project.model';
 import { getLogger } from '../utils/logger';
 import { getProjectEntry } from '../utils/project';
 import { timeout } from '../utils/promise';
+import { getYargsOption } from '../yargs';
 import { ApiService } from './api.service';
 import { StoreService } from './store.service';
+
+const { argv } = getYargsOption();
 
 export interface SandboxOption {
   store?: Store;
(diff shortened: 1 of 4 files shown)


jamesbayly:
Can you also update the docs for this change?

packages/node/src/indexer/sandbox.service.ts
```diff
@@ -28,7 +31,9 @@ const DEFAULT_OPTION: NodeVMOptions = {
   wasm: false,
```
ianhe8x, line 31:
```suggestion
  wasm: argv.unsafe,
```

packages/node/src/yargs.ts
```diff
@@ -34,6 +34,11 @@ export function getYargsOption() {
       demandOption: false,
       describe: 'Force clean the database, dropping project schemas and tables',
     },
+    unsafe: {
+      type: 'boolean',
+      demandOption: false,
+      describe: 'Allows usage of any built-in module for sandbox service',
```
ianhe8x, line 40:
```suggestion
      describe: 'Allows usage of any built-in module within the sandbox',
```

packages/node/src/main.ts
```diff
@@ -21,6 +21,13 @@ async function bootstrap() {
     const indexerManager = app.get(IndexerManager);
     await indexerManager.start();
     await app.listen(port);
+
+    if (argv('unsafe')) {
```
ianhe8x, line 25: move it before `const app = await NestFactory.create`

ianhe8x (approved):

PR #628, opened 2021-11-14
touchup docs

wqsz7xn:
- moved english to their own `en` folder
- added force-clean to english docs
- touched up cli/node to have proper capitalisation for descriptions

PR #636, opened 2021-11-17
add hero courses link in de, zh, vi, clean up docs directory

lachlan1m:
Hero course files I added are english, however better than 404 (and a lot of the other foreign language pages are still english anyway).


wqsz7xn:
lgtm, i didn't think properly about moving the english files to `en`, we either need to fix it such that all root requests are redirected to include `en/` or just revert (or undo) my commit which moves them to `en` and just get rid of it. I tried playing about with routers but nothing seems to be working for me. The following is a good issue. This might be a bit weird with crowdin translations as well
https://github.com/vuejs/vuepress/issues/363

wqsz7xn:
https://vuepress-community.netlify.app/en/plugins/redirect/
this seems to do what we are looking for, i'll give this a try

wqsz7xn:
above plugin doesn't seem to do what I think it's describing. existing hyperlinks will still be broken as in it's not redirecting `http://localhost:8080/quickstart/quickstart/#preparation` --> `http://localhost:8080/en/quickstart/quickstart/#preparation` by default

wqsz7xn:
we have decided to remove the english folder that I created in https://github.com/subquery/subql/pull/628
reason being is that existing hyperlinks to the docs were not resolving even with the root redirect plugin (see discussion in #628), i suspect that the plugin I tried was just for keeping the structure folder structure clean instead of redirecting existing links as well. Note that we have to get crowdin translations in as well so not messing with the folder structure is less of a pain. Let's please move over to https://github.com/subquery/subql-docs after this after we have merge existing doc PR changes

jamesbayly (approved):

ianhe8x (approved):

ianhe8x:
I probably won't test locally, so merge if you guys feel fine with it

PR #644, opened 2021-11-18
limit query size

wqsz7xn:
limits the size of the queries that can be made against subql/query. this doesn't include the depth stuff and i'll put that in a different pr later


packages/query/src/graphql/graphql.test.ts
```diff
@@ -10,9 +10,9 @@ import {plugins} from './plugins';
 describe('GraphqlModule', () => {
   const dbSchema = 'subquery_1';
 
-  const config = new Config({});
+  const config: Config = new Config({});
```
ianhe8x, line 13: It's fine without the explicit type def, it looks like Java.

ianhe8x (approved):

PR #657, opened 2021-11-23
Rework block processing

stwiname:
Rewrites the BlockQueue and replaces block processing to remove any set timeouts (via `delay` function).

This seems to have a large performance increase. Running locally I've gone from <1 bps to ~7bps and I've seen as high as 13bps.

Further improvements could be achieved by using a fat dictionary to reduce wrapping blocks or introduce some form of lazy loading so only what is used is processed.

### DIFF
--- a/packages/node/src/indexer/BlockedQueue.ts
+++ b/packages/node/src/indexer/BlockedQueue.ts
@@ -40,6 +40,7 @@ export class AutoQueue<T> {
   private pendingPromise = false;
   private queue = new Queue<Action<T>>();
   private _capacity: number;
+  private _abort = false;
 
   private eventEmitter = new EventEmitter2();
 
@@ -83,8 +84,13 @@ export class AutoQueue<T> {
     });
   }
 
-  private async take(): Promise<void> {
+  async take(): Promise<void> {
     if (this.pendingPromise) return;
+    if (this._abort) {
+      // Reset so it can be restarted
+      this._abort = false;
+      return;
+    }
 
     const action = this.queue.take();
 
@@ -103,6 +109,10 @@ export class AutoQueue<T> {
       this.pendingPromise = false;
       void this.take();
     }
+  }
+
+  abort(): void {
+    this._abort = true;
   }
 
   on(
--- a/packages/node/src/indexer/BlockedQueue.ts
+++ b/packages/node/src/indexer/BlockedQueue.ts
@@ -1,59 +1,7 @@
 // Copyright 2020-2021 OnFinality Limited authors & contributors
 // SPDX-License-Identifier: Apache-2.0
 
-import { delay } from '../utils/promise';
-
-export class BlockedQueue<T> {
-  private _queue: T[] = [];
-  private _maxSize: number;
-
-  constructor(size: number) {
-    this._maxSize = size;
-  }
-
-  get size(): number {
-    return this._queue.length;
-  }
-
-  get freeSize(): number {
-    return this._maxSize - this._queue.length;
-  }
-
-  put(item: T): void {
-    if (this._queue.length >= this._maxSize) {
-      throw new Error('BlockedQueue exceed max size');
-    }
-    this._queue.push(item);
-  }
-
-  putAll(items: T[]): void {
-    if (this._queue.length + items.length > this._maxSize) {
-      throw new Error('BlockedQueue exceed max size');
-    }
-    this._queue.push(...items);
-  }
-
-  async take(): Promise<T> {
-    while (!this.size) {
-      await delay(0.1);
-    }
-    return this._queue.shift();
-  }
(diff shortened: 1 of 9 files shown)


ianhe8x:
The performance gain mainly comes from parallelization, but issue here is we want to guarantee the block processing order.

stwiname:
> The performance gain mainly comes from parallelization, but issue here is we want to guarantee the block processing order.

Blocks are still processed in order. It just doesn't read well [here](https://github.com/subquery/subql/pull/657/files#diff-fcb06f77cc4b787f2ed3c116347e39144575cd2ce8fb07af6b6ea11ef0ac2a05R306) because it's just filling the queue with batch size blocks then waiting for them all to be processed.

Further work is required to make fetching blocks and processing blocks happen independent of each other

ianhe8x:
My mistake, your implementation is correct, well done.

packages/node/src/indexer/BlockedQueue.ts
```diff
@@ -54,3 +54,78 @@ export class BlockedQueue<T> {
...
+  async put(item: () => T): Promise<T> {
+    return this.putMany([item])[0];
+  }
+
+  putMany(tasks: Array<() => T>): Promise<T>[] {
+    return tasks.map((task, index) => {
+      return new Promise((resolve, reject) => {
+        this.queue.put({ task, resolve, reject });
+        if (tasks.length - 1 === index) {
+          void this.take();
```
ianhe8x, line 104: It seems the return value is dropped, and not used anywhere

packages/node/src/indexer/BlockedQueue.ts
```diff
@@ -54,3 +54,78 @@ export class BlockedQueue<T> {
...
+        }
+      });
+    });
+  }
+
+  async take(): Promise<boolean> {
```
ianhe8x, line 110: Do i understand it correct, that when a batch of task is pushed into the queue, `take()` will be called same times initially, but except the first call, you expect the others will return false and exit early. The first success call will then recursively call itself until all tasks in the queue are executed.

stwiname: `take` gets called again to ensure that we are processing items in the queue because it stops being recursive if the queue is emptied. It only needs to be called a single time for a batch in order to start processing again.

Its based on this answer https://stackoverflow.com/a/63208885/7492768 which is well worth the read

packages/node/src/indexer/BlockedQueue.ts
stwiname: I have dropped the return value from `take`

packages/node/src/indexer/fetch.service.ts
```diff
@@ -259,108 +228,146 @@ export class FetchService implements OnApplicationShutdown {
...
+  private async queueBlocks(
+    processor: (value: BlockContent) => Promise<void> | void,
+  ): Promise<void> {
+    let bufferBlocks = await this.nextBlocks();
 
-      const bufferBlocks = await this.blockNumberBuffer.takeAll(takeCount);
-      const metadataChanged = await this.fetchMeta(
-        bufferBlocks[bufferBlocks.length - 1],
+    if (!bufferBlocks.length) {
+      const wait = 3;
```
stwiname, line 287: I'm not sure if we want this to be configurable or wait batch size when up to date with the chain.

ianhe8x:
I tested with two versions, but didn't find obvious difference. I think it's expected actually, if the performance gain comes from removing delay when the queue is empty, since most of time the queue is full.
If we can keep the interface untouched, we can compare bare queue r/w operations directly which might be fairer.

BTW, i noticed when running this branch, fetch block starts from 2 not 1, which is likely a bug.
`<fetch> INFO fetch block [2,101], total 100 blocks`

stwiname:
> I think it's expected actually, if the performance gain comes from removing delay when the queue is empty, since most of time the queue is full.

Yeah this is true to some extent. But the usage of `while(...) { await delay(...) }` in a couple of places floods the event loop causing other things to slow down. And it's gone from 2 queues (block processing, block buffer) to a single queue for both, so they are not each waiting while the other is doing stuff.


 > If we can keep the interface untouched, we can compare bare queue r/w operations directly which might be fairer.

I can do this but i think it would be a lot more work and be less readable. I think we would be better to focus on a general benchmark solution. That way we could test other changes and their effect on performance.

ianhe8x:
have another trial with the moonbeam dictionary, 
image

stwiname:
> have another trial with the moonbeam dictionary, the speed difference is subtle. Considering the scale of changes (and still have unsolved issues with this PR), it doesn't look like worth to continue. 

I see a decent speed up still, but i think it can be quite dependent on time it takes to fetch blocks. I'm keen to keep working on this, ideally finding a good way to benchmark. The issue in the screenshot should be trivial to fix.

Issue #650, opened 2021-11-22
add file based log for subql/query


zhex:
the query service will have performance issues when the coming requests are growing. to avoid the performance issue, we should support writing logs into the file.

related changes:

- logger lib in the common package
- query service launch argument to support a given destination, the default could be stdout

lachlan1m:
#667

PR #667, opened 2021-11-25
Add subql/query log to file

lachlan1m:
#650

### DIFF
--- a/packages/common/src/logger/logger.ts
+++ b/packages/common/src/logger/logger.ts
@@ -9,14 +9,14 @@ export interface LoggerOption {
   outputFormat?: 'json' | 'colored';
   level?: string;
   nestedKey?: string;
-  toFile?: boolean;
+  logFileDir?: string;
 }
 
 export class Logger {
   private pino: Pino.Logger;
   private childLoggers: {[category: string]: Pino.Logger} = {};
 
-  constructor({level: logLevel = 'info', nestedKey, outputFormat, toFile = false}: LoggerOption) {
+  constructor({level: logLevel = 'info', logFileDir, nestedKey, outputFormat}: LoggerOption) {
     const options = {
       messageKey: 'message',
       timestamp: () => `,"timestamp":"${new Date().toISOString()}"`,
@@ -75,8 +75,8 @@ export class Logger {
       },
     } as Pino.LoggerOptions;
 
-    if (toFile) {
-      this.pino = Pino(options, Pino.destination('./logger.log'));
+    if (logFileDir) {
+      this.pino = Pino(options, Pino.destination(logFileDir));
     } else {
       this.pino = Pino(options);
     }
--- a/packages/common/src/logger/logger.ts
+++ b/packages/common/src/logger/logger.ts
@@ -9,14 +9,15 @@ export interface LoggerOption {
   outputFormat?: 'json' | 'colored';
   level?: string;
   nestedKey?: string;
+  logToFile?: boolean;
 }
 
 export class Logger {
   private pino: Pino.Logger;
   private childLoggers: {[category: string]: Pino.Logger} = {};
 
-  constructor({level: logLevel = 'info', nestedKey, outputFormat}: LoggerOption) {
-    this.pino = Pino({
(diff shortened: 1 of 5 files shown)


packages/common/src/logger/logger.ts
```diff
@@ -72,7 +73,13 @@ export class Logger {
           return Object.prototype.toString.apply(input) === '[object Object]';
         }
       },
-    });
+    } as Pino.LoggerOptions;
+
+    if (toFile) {
+      this.pino = Pino(options, Pino.destination('./logger.log'));
```
lachlan1m, line 79: Left log file name as nondescript incase want to use in another package.

zhex: log files should be rotated

lachlan1m: I've updated so you can specify log file name and dir.

ianhe8x:
@lachlan1m can you check this https://github.com/ExpediaGroup/pino-rotating-file ?

packages/common/src/logger/logger.ts
```diff
@@ -72,7 +76,25 @@ export class Logger {
...
+      }
+      this.pino = Pino(
+        options,
+        rotator.getStream({
+          filename: `${logDirectory}/%DATE%.log`,
+          frequency: 'daily',
+          max_logs: '7d',
+          size: '1G',
+          verbose: false,
+        })
```
lachlan1m, line 93: @zhex let me know if you want this changed at all.

packages/common/src/logger/logger.ts
```diff
@@ -1,22 +1,25 @@
 // Copyright 2020-2021 OnFinality Limited authors & contributors
 // SPDX-License-Identifier: Apache-2.0
 
+import fs from 'fs';
 import {stringify} from 'flatted';
 import Pino, {LevelWithSilent} from 'pino';
 import {colorizeLevel, ctx} from './colors';
 
+const rotator = require('file-stream-rotator');
```
ianhe8x, line 9: use import

ianhe8x (approved):
I'm good with this. @zhex How do you want to config the log behavior?

packages/common/src/logger/logger.ts
lachlan1m: changes to a different package with better typescript support.

zhex:
I saw we can set the log dir which is really helpful. all the other settings we can leave it as default.

zhex (approved):

packages/common/src/logger/logger.ts
```diff
@@ -72,7 +74,25 @@ export class Logger {
...
+    } as Pino.LoggerOptions;
+
+    if (logDirectory) {
+      if (!fs.existsSync(logDirectory)) {
+        fs.mkdirSync(logDirectory);
+      }
+      // will add concatenate timestamp to 'file.log'
+      this.pino = Pino(
+        options,
+        createStream('file.log', {
```
ianhe8x, line 86: @lachlan1m what log file's name looks like, why it is hardcoded as 'file.log' ?

packages/common/src/logger/logger.ts
```diff
@@ -72,7 +74,25 @@ export class Logger {
...
+          interval: '1d',
+          maxFiles: 7,
+          maxSize: '1G',
+          path: `${logDirectory}`,
```
ianhe8x, line 90:
```suggestion
          path: logDirectory,
```

packages/query/src/yargs.ts
```diff
@@ -19,6 +19,11 @@ export function getYargsOption() {
       type: 'boolean',
       demandOption: false,
     },
+    'log-dir': {
```
ianhe8x, line 22: To avoid using a mandatory file.log as log name, I suggest use `log-path` to allow user specify name. You can extract the dir via `path.dirname(path.resolve(filepath))`
Then I suggest to add another flag 'log-rotate` to enable rotate. (default to false)

ianhe8x (changes requested):

packages/query/src/yargs.ts
lachlan1m: I have updated this.

ianhe8x (approved):

PR #668, opened 2021-11-25
Alter batchSize based on memory usage

lachlan1m:
TODO:

- [x] Implement being able to alter batchSize based on memory usage.
- [x] Make sure that the adjustments being made don't negatively impact performance more than necessary to prevent heap out of memory error.

### DIFF
--- a/packages/node/src/indexer/fetch.service.ts
+++ b/packages/node/src/indexer/fetch.service.ts
@@ -217,7 +217,6 @@ export class FetchService implements OnApplicationShutdown {
 
   @Interval(BLOCK_TIME_VARIANCE * 10000)
   getMemoryUsage() {
-    // I need to figure out what the threshold of memory usage should be.
     const formatMemoryUsage = (data) =>
       `${Math.round((data / 1024 / 1024) * 100) / 100} MB`;
 
@@ -256,7 +255,6 @@ export class FetchService implements OnApplicationShutdown {
       }
     }
 
-    //I should maybe consider not doing this based on interval
     console.log(memoryUsage);
     console.log(ratio);
   }
--- a/packages/node/src/indexer/fetch.service.ts
+++ b/packages/node/src/indexer/fetch.service.ts
@@ -92,7 +92,7 @@ export class FetchService implements OnApplicationShutdown {
   private parentSpecVersion: number;
   private useDictionary: boolean;
   private dictionaryQueryEntries?: DictionaryQueryEntry[];
-  private scaleBatchSize: number;
+  private batchSizeScale: number;
 
   constructor(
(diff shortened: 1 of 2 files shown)


ianhe8x:
What's the usual heapUsed:heapTotal ratio? How did you decide the low/high threshold?

ianhe8x:
Looks good, i suggest adding a minimum batchSize just in case the heap used isn't shrunken for some reasons.

packages/node/src/indexer/fetch.service.ts
```diff
@@ -213,6 +215,50 @@ export class FetchService implements OnApplicationShutdown {
     await this.getBestBlockHead();
   }
 
+  @Interval(BLOCK_TIME_VARIANCE * 10000)
+  getMemoryUsage() {
+    const formatMemoryUsage = (data) =>
```
ianhe8x, line 220: these declaration can be escalated

packages/node/src/indexer/fetch.service.ts
```diff
@@ -213,6 +215,50 @@ export class FetchService implements OnApplicationShutdown {
     await this.getBestBlockHead();
   }
 
+  @Interval(BLOCK_TIME_VARIANCE * 10000)
```
ianhe8x, line 218: The interval isn't related to block time logically. If you think 50s is a reasonable value, just make a const and assign 50,000 to it.

stwiname:
Out of curiosity, what sort of batch size does it tend to and with how much memory?

lachlan1m:
> What's the usual heapUsed:heapTotal ratio? How did you decide the low/high threshold?

It is usually quite high in the 0.8 - 0.95% range (heapUsed / heapTotal). This is how I came up with the threshold values.

lachlan1m:
> Out of curiosity, what sort of batch size does it tend to and with how much memory?

It tended jump around quite a lot because memory usage can change sporadically. So what I've done is lengthen the interval in which will check memory and increase the amount scales down batch size by to 10% (if it gets above 95% heap usage).

lachlan1m:
I've also added a flag which can be used to disable this feature in yargs (--scale-batch-size).

packages/node/src/yargs.ts
```diff
@@ -39,6 +39,12 @@ export function getYargsOption() {
       describe: 'Batch size of blocks to fetch in one round',
       type: 'number',
     },
+    'scale-batch-size': {
+      type: 'boolean',
+      demandOption: false,
+      describe: 'scale batch size based on memory usage',
+      default: true,
```
ianhe8x, line 46: default to false

ianhe8x (approved):

Issue #671, opened 2021-11-27
log the error before throwing for incorrect genesis hash


jiqiang90:
https://github.com/subquery/subql/blob/874816754d45062fb08366e6218bb89dbf470d2c/packages/node/src/indexer/api.service.ts#L71

PR #675, opened 2021-11-28
Log API error

wqsz7xn:
- [x] Log API error, probably can do this for a few more around but unsure which ones are important enough to log
- https://github.com/subquery/subql/issues/671

### DIFF
--- a/packages/node/src/indexer/api.service.ts
+++ b/packages/node/src/indexer/api.service.ts
@@ -74,8 +74,6 @@ export class ApiService implements OnApplicationShutdown {
       const err = new Error(`Network genesisHash doesn't match expected genesisHash. expected="${network.genesisHash}" actual="${this.networkMeta.genesisHash}`);
       logger.error(err, err.message);
       throw err;
-      logger.error(err);
-      throw new Error(err);
     }
 
     return this;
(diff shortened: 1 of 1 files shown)


packages/node/src/indexer/api.service.ts
```diff
@@ -68,9 +71,9 @@ export class ApiService implements OnApplicationShutdown {
       network.genesisHash &&
       network.genesisHash !== this.networkMeta.genesisHash
     ) {
-      throw new Error(
-        `Network genesisHash doesn't match expected genesisHash. expected="${network.genesisHash}" actual="${this.networkMeta.genesisHash}`,
-      );
+      const err = `Network genesisHash doesn't match expected genesisHash. expected="${network.genesisHash}" actual="${this.networkMeta.genesisHash}`;
```
ianhe8x, line 74:
```suggestion
      const err = new Error(`Network genesisHash doesn't match expected genesisHash. expected="${network.genesisHash}" actual="${this.networkMeta.genesisHash}`);
      logger.error(err, err.message);
      throw err;
```

ianhe8x: This way it logs error's stack when --output-fmt=json is enabled

ianhe8x:
@wqsz7xn failed at code style check

wqsz7xn:
oh my bad

wqsz7xn:
Not sure why the CI tests are failing, not related to this though

ianhe8x (approved):

PR #676, opened 2021-11-28
Exit the node cleanly when genesis hash doesn't match

stwiname:


stwiname:
Duplicate of https://github.com/subquery/subql/pull/675

PR #679, opened 2021-11-29
Disable shadow api in unsafe mode

yrong:


ianhe8x:
The behaviors of with and withoug --unsafe are totally different. 
I suggest you can inject api as global.unsafeApi when --unsafe is enabled.

yrong:
> The behaviors of with and withoug --unsafe are totally different. I suggest you can inject api as global.unsafeApi when --unsafe is enabled.

@ianhe8x some refact accordingly

ianhe8x (approved):

PR #683, opened 2021-11-30
Depreciate subqueries table 

wqsz7xn:
Resubmitting a new version of this for review
- depreciate subqueries table in favour of project metadata table
- will automatically convert subqueries entry to metadata entry, works for existing projects
- adds `--schema` flag, use this as primary directive for schema name/location
- reordered some of the functions in indexer in order of calling, makes more sense

### DIFF
--- a/packages/common/src/query/types.ts
+++ b/packages/common/src/query/types.ts
@@ -2,6 +2,8 @@
 // SPDX-License-Identifier: Apache-2.0
 
 export type MetaData = {
+  name: string;
+  schema: string;
   lastProcessedHeight: number;
   lastProcessedTimestamp: number;
   targetHeight: number;
--- a/packages/common/src/query/types.ts
+++ b/packages/common/src/query/types.ts
@@ -2,8 +2,6 @@
 // SPDX-License-Identifier: Apache-2.0
 
 export type MetaData = {
-  name: string;
-  schema: string;
   lastProcessedHeight: number;
   lastProcessedTimestamp: number;
   targetHeight: number;
--- a/packages/node/src/configure/configure.module.ts
+++ b/packages/node/src/configure/configure.module.ts
@@ -56,7 +56,14 @@ export class ConfigureModule {
     } else {
       if (!argv.subquery) {
         logger.error(
-          'subquery path is missing neither in cli options nor in config file',
+          'Subquery path is missing neither in cli options nor in config file',
+        );
+        yargsOptions.showHelp();
+        process.exit(1);
+      }
+      if (argv.subquery && argv.schema) {
+        logger.error(
+          'Conflicting flags for specifying subquery schema and path',
         );
         yargsOptions.showHelp();
         process.exit(1);
--- a/packages/node/src/configure/configure.module.ts
+++ b/packages/node/src/configure/configure.module.ts
@@ -61,12 +61,10 @@ export class ConfigureModule {
         yargsOptions.showHelp();
         process.exit(1);
       }
-      if (argv.subquery && argv.schema) {
-        logger.error(
-          'Conflicting flags for specifying subquery schema and path',
+      if (argv['subquery-name']) {
+        logger.info(
+          'Note that argument --subquery-name has been deprecated in favour of --schema',
         );
-        yargsOptions.showHelp();
-        process.exit(1);
       }
       assert(argv.subquery, 'subquery path is missing');
       config = new NodeConfig(defaultSubqueryName(yargsToIConfig(argv)));
--- a/packages/node/src/configure/configure.module.ts
+++ b/packages/node/src/configure/configure.module.ts
@@ -11,6 +11,8 @@ import { getYargsOption } from '../yargs';
 import { IConfig, MinConfig, NodeConfig } from './NodeConfig';
 import { SubqueryProject } from './project.model';
 
+const logger = getLogger('configure');
+
 const YargsNameMapping = {
   local: 'localMode',
 };
@@ -42,7 +44,33 @@ function defaultSubqueryName(config: Partial<IConfig>): MinConfig {
   } as MinConfig;
 }
 
-const logger = getLogger('configure');
+// Check if a subquery name is a valid schema name
+function validDbSchemaName(name: string): boolean {
+  if (name.length === 0) {
+    return false;
+  } else {
+    const regexp = new RegExp('[a-zA-Z_]');
+    const flag0 = !name.startsWith('pg_'); // Reserved identifier
+    const flag1 = regexp.test(name.charAt(0)); // Prefix clamp
+    const flag2 = Buffer.byteLength(name, 'utf-8') <= 63; // 63 byte limit
+    if (!flag0) {
+      logger.error(
+        `Invalid schema name '${name}', schema name must not be prefixed with 'pg_'`,
+      );
+    }
+    if (!flag1) {
+      logger.error(
+        `Invalid schema name '${name}', schema name must start with a letter or underscore`,
+      );
+    }
+    if (!flag2) {
+      logger.error(
+        `Invalid schema name '${name}', schema name length must be less than 63 bytes`,
+      );
+    }
+    return flag0 && flag1 && flag2;
+  }
+}
 
 @Global()
 @Module({})
@@ -66,8 +94,13 @@ export class ConfigureModule {
           'Note that argument --subquery-name has been deprecated in favour of --schema',
         );
       }
+
       assert(argv.subquery, 'subquery path is missing');
       config = new NodeConfig(defaultSubqueryName(yargsToIConfig(argv)));
+    }
+
+    if (!validDbSchemaName(config.dbSchema)) {
+      process.exit(1);
     }
 
     if (config.debug) {
--- a/packages/node/src/configure/configure.module.ts
+++ b/packages/node/src/configure/configure.module.ts
@@ -49,10 +49,10 @@ function validDbSchemaName(name: string): boolean {
   if (name.length === 0) {
     return false;
   } else {
-    const regexp = new RegExp('[a-zA-Z_]');
+    name = name.toLowerCase();
+    const regexp = new RegExp('^[a-zA-Z_][a-zA-Z0-9_\\-\\/]{0,62}$');
     const flag0 = !name.startsWith('pg_'); // Reserved identifier
-    const flag1 = regexp.test(name.charAt(0)); // Prefix clamp
-    const flag2 = Buffer.byteLength(name, 'utf-8') <= 63; // 63 byte limit
+    const flag1 = regexp.test(name); // <= Valid characters, less than 63 bytes
     if (!flag0) {
       logger.error(
         `Invalid schema name '${name}', schema name must not be prefixed with 'pg_'`,
@@ -60,15 +60,11 @@ function validDbSchemaName(name: string): boolean {
     }
     if (!flag1) {
       logger.error(
-        `Invalid schema name '${name}', schema name must start with a letter or underscore`,
+        `Invalid schema name '${name}', schema name must start with a letter or underscore, 
+         be less than 63 bytes and must contain only valid alphanumeric characters (can include characters '_-/')`,
       );
     }
-    if (!flag2) {
-      logger.error(
-        `Invalid schema name '${name}', schema name length must be less than 63 bytes`,
-      );
-    }
-    return flag0 && flag1 && flag2;
+    return flag0 && flag1;
   }
 }
 
--- a/packages/node/src/configure/configure.module.ts
+++ b/packages/node/src/configure/configure.module.ts
@@ -45,7 +45,7 @@ function defaultSubqueryName(config: Partial<IConfig>): MinConfig {
 }
 
 // Check if a subquery name is a valid schema name
-function validDbSchemaName(name: string): boolean {
+export function validDbSchemaName(name: string): boolean {
   if (name.length === 0) {
     return false;
   } else {
--- a/packages/node/src/indexer/indexer.manager.ts
+++ b/packages/node/src/indexer/indexer.manager.ts
@@ -183,15 +183,12 @@ export class IndexerManager {
     if (!schema) {
       schema = await this.createProjectSchema();
     } else {
+      console.log(schema);
+      console.log(`old name: ${schema}`);
+      console.log(`new name: ${this.nodeConfig.subqueryName}`);
       if (isSubqueries) {
         await this.sequelize.query(
-          `ALTER SCHEMA :oldname RENAME TO :newname `,
-          {
-            replacements: {
-              oldname: schema,
-              newname: `"${this.nodeConfig.subqueryName}"`,
-            },
-          },
+          `ALTER SCHEMA "${schema}" RENAME TO "${this.nodeConfig.subqueryName}"`,
         );
         schema = this.nodeConfig.subqueryName;
       }
@@ -207,7 +204,7 @@ export class IndexerManager {
           await this.sequelize.query(
             ` DELETE
               FROM public.subqueries
-              where name = :name`,
+              WHERE name = :name`,
             {
               replacements: { name: this.nodeConfig.subqueryName },
               type: QueryTypes.DELETE,
--- a/packages/node/src/indexer/indexer.manager.ts
+++ b/packages/node/src/indexer/indexer.manager.ts
@@ -304,7 +304,7 @@ export class IndexerManager {
       if (project) {
         await metadataRepo.upsert({
           key: 'lastProcessedHeight',
-          value: project.nextBlockHeight,
+          value: project.nextBlockHeight - 1,
         });
       } else {
         await metadataRepo.upsert({ key: 'lastProcessedHeight', value: 0 });
--- a/packages/node/src/indexer/indexer.manager.ts
+++ b/packages/node/src/indexer/indexer.manager.ts
@@ -228,7 +228,12 @@ export class IndexerManager {
         },
       );
       if (result.length === 0) {
-        schema = undefined;
+        if (argv.schema) {
+          logger.error(`Was not able to find argument schema ${argv.schema}`);
+          process.exit(1);
+        } else {
+          schema = undefined;
+        }
       } else {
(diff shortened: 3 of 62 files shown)


packages/common/src/query/types.ts
```diff
@@ -2,6 +2,8 @@
 // SPDX-License-Identifier: Apache-2.0
 
 export type MetaData = {
+  name: string;
+  schema: string;
```
ianhe8x, line 6: I don't think we need to move name and schema to _metadata

packages/node/src/indexer/indexer.manager.ts
```diff
@@ -104,9 +104,15 @@ export class IndexerManager {
...
         }
       }
-      this.subqueryState.nextBlockHeight =
-        block.block.header.number.toNumber() + 1;
-      await this.subqueryState.save({ transaction: tx });
+
+      await this.metadataRepo.upsert(
+        {
+          key: 'lastProcessedHeight',
+          value: block.block.header.number.toNumber() + 1,
```
ianhe8x, line 111:
```suggestion
          value: block.block.header.number.toNumber(),
```

packages/node/src/indexer/indexer.manager.ts
```diff
@@ -169,30 +178,128 @@ export class IndexerManager {
...
+      const result = await this.sequelize.query(
+        `select db_schema from public.subqueries where name = :name`,
+        {
+          replacements: { name: schema },
+          type: QueryTypes.SELECT,
+        },
       );
-      process.exit(1);
+      if (result.length === 0) {
+        schema = undefined;
```
jiqiang90, line 232: if can not find this schema in db, we should use `argv.schema` or project name as its schema name

packages/node/src/indexer/indexer.manager.ts
```diff
@@ -169,30 +178,128 @@ export class IndexerManager {
...
+    return schema;
+  }
+
+  private async getProjectSchema(): Promise<string> {
+    let schema = argv.schema;
+    if (!schema) {
+      schema = this.nodeConfig.subqueryName;
+    }
+    const schemas = (await this.sequelize.showAllSchemas(
+      undefined,
```
ianhe8x, line 221:
```suggestion
    this.sequelize.showAllSchemas()
```

packages/node/src/indexer/indexer.manager.ts
```diff
@@ -169,30 +178,128 @@ export class IndexerManager {
...
+
+  private async createProjectSchema(): Promise<string> {
+    let schema: string;
+    if (this.nodeConfig.localMode) {
+      // create tables in default schema if local mode is enabled
+      schema = DEFAULT_DB_SCHEMA;
     } else {
-      return Math.min(...startBlocksList);
+      const suffix = await this.nextSubquerySchemaSuffix();
+      schema = `subquery_${suffix}`;
```
jiqiang90, line 247: We are no longer use this format, use `argv.schema` or project name

packages/node/src/indexer/indexer.manager.ts
jiqiang90, line 232: if can not find this schema in db, we should use argv.schema or project name as its schema name

packages/node/src/indexer/indexer.manager.ts
```diff
@@ -340,20 +389,23 @@ export class IndexerManager {
...
+  private async filterDataSources(): Promise<SubqlDatasource[]> {
     let filteredDs = this.getDataSourcesForSpecName();
     if (filteredDs.length === 0) {
       logger.error(
         `Did not find any dataSource match with network specName ${this.api.runtimeVersion.specName}`,
       );
       process.exit(1);
     }
+    const lastProcessedHeight = (
+      await this.metadataRepo.findOne({ where: { key: 'lastProcessedHeight' } })
```
ianhe8x, line 401: You can maintain a copy in the mem.

packages/node/src/indexer/indexer.manager.ts
jiqiang90, line 111: why + 1 here? For `nextBlockHeight` it should be  lastProcessedHeight + 1

packages/node/src/indexer/indexer.manager.ts
```diff
@@ -140,25 +146,28 @@ export class IndexerManager {
...
+        this.mmrService.init(schema),
       ]);
     }
 
+    const lastProcessedHeight = (
+      await this.metadataRepo.findOne({ where: { key: 'lastProcessedHeight' } })
+    ).value;
     void this.fetchService
-      .startLoop(this.subqueryState.nextBlockHeight)
+      .startLoop(lastProcessedHeight as number)
```
jiqiang90, line 164: this should be lastProcessedHeight + 1

packages/node/src/indexer/indexer.manager.ts
```diff
@@ -211,108 +318,50 @@ export class IndexerManager {
...
+    }
+
     if (!keyValue.blockOffset) {
       const offsetValue = (this.getStartBlockFromDataSources() - 1).toString();
-      await this.storeService.setMetadata('blockOffset', offsetValue);
+      await metadataRepo.upsert({ key: 'blockOffset', value: offsetValue });
+    }
+
+    if (!keyValue.lastProcessedHeight) {
+      await metadataRepo.upsert({ key: 'lastProcessedHeight', value: 1 });
```
ianhe8x, line 342:
```suggestion
      await metadataRepo.upsert({ key: 'lastProcessedHeight', value: 0 });
```

packages/node/src/indexer/indexer.manager.ts
```diff
@@ -169,30 +178,128 @@ export class IndexerManager {
...
+  private async getProjectSchema(): Promise<string> {
+    let schema = argv.schema;
+    if (!schema) {
+      schema = this.nodeConfig.subqueryName;
+    }
+    const schemas = (await this.sequelize.showAllSchemas(
+      undefined,
+    )) as unknown as string[];
+    if (!schemas.includes(schema)) {
+      const result = await this.sequelize.query(
```
ianhe8x, line 224: why don't use `this.subqueryRepo`?

packages/node/src/indexer/indexer.manager.ts
```diff
@@ -169,30 +178,128 @@ export class IndexerManager {
...
+  }
+
+  private async createProjectSchema(): Promise<string> {
+    let schema: string;
+    if (this.nodeConfig.localMode) {
+      // create tables in default schema if local mode is enabled
+      schema = DEFAULT_DB_SCHEMA;
     } else {
-      return Math.min(...startBlocksList);
+      const suffix = await this.nextSubquerySchemaSuffix();
```
ianhe8x, line 246: After this PR, we don't generate schema for project but use what was given when launch the node.

packages/node/src/indexer/indexer.manager.ts
```diff
@@ -340,20 +389,23 @@ export class IndexerManager {
...
         `Did not find any dataSource match with network specName ${this.api.runtimeVersion.specName}`,
       );
       process.exit(1);
     }
+    const lastProcessedHeight = (
+      await this.metadataRepo.findOne({ where: { key: 'lastProcessedHeight' } })
+    ).value;
     filteredDs = filteredDs.filter(
-      (ds) => ds.startBlock <= this.subqueryState.nextBlockHeight,
+      (ds) => ds.startBlock <= lastProcessedHeight,
```
jiqiang90, line 404: lastProcessedHeight + 1

ianhe8x (changes requested):

packages/common/src/query/types.ts
jiqiang90, line 6: I wonder if this `schema` has been used anywhere

packages/node/src/indexer/indexer.manager.ts
wqsz7xn: argument for logging in this function isn't optional, explicit undefined required

packages/node/src/indexer/indexer.manager.ts
wqsz7xn: subqueryRepo is sourced from `subqueries` aiak, i've removed it from the implementation

packages/node/src/indexer/indexer.manager.ts
wqsz7xn: this method is to fetch a schema for the project if it exists, otherwise it will return undefined, if I make this change it will mean that I can no longer detect if a project must be created using `createProjectSchema`

wqsz7xn:
@ianhe8x @jiqiang90 
I remembered the reason why I had put the `name` field into the metadata, if we have an old project underneath the schema `subquery_1` currently my implementation will take that row from the `subqueries` table, create a `_metadata` table and then delete that row so we needed some way of tracking the name of the project if the schema name isn't the project name.

What i'm going to do instead is just move old project schemas to the project name

packages/node/src/indexer/indexer.manager.ts
```diff
@@ -349,11 +384,11 @@ export class IndexerManager {
       process.exit(1);
     }
     filteredDs = filteredDs.filter(
-      (ds) => ds.startBlock <= this.subqueryState.nextBlockHeight,
+      (ds) => ds.startBlock <= lastProcessedHeight + 1,
     );
     if (filteredDs.length === 0) {
       logger.error(
-        `Your start block is greater than the current indexed block height in your database. Either change your startBlock (project.yaml) to <= ${this.subqueryState.nextBlockHeight} or delete your database and start again from the currently specified startBlock`,
+        `Your start block is greater than the current indexed block height in your database. Either change your startBlock (project.yaml) to <= ${lastProcessedHeight} or delete your database and start again from the currently specified startBlock`,
```
ianhe8x, line 391:
```suggestion
        `Your start block is greater than the current indexed block height in your database. Either change your startBlock (project.yaml) to <= ${lastProcessedHeight+1} or delete your database and start again from the currently specified startBlock`,
```

packages/query/src/graphql/plugins/GetMetadataPlugin.ts
```diff
@@ -12,6 +12,8 @@ const {version: packageVersion} = require('../../../package.json');
 const indexerUrl = argv('indexer') as string | undefined;
 
 const METADATA_TYPES = {
+  name: 'string',
```
ianhe8x, line 15: no need to expose them in query service

wqsz7xn: oops, let me reread through the diff, there's probably a bit like this

packages/node/src/indexer/indexer.manager.ts
ianhe8x: Then you can keep subqueryRepo there.

packages/node/src/configure/configure.module.ts
```diff
@@ -56,7 +56,14 @@ export class ConfigureModule {
     } else {
       if (!argv.subquery) {
         logger.error(
-          'subquery path is missing neither in cli options nor in config file',
+          'Subquery path is missing neither in cli options nor in config file',
+        );
+        yargsOptions.showHelp();
+        process.exit(1);
+      }
+      if (argv.subquery && argv.schema) {
```
ianhe8x, line 64: I think it is allowed, --schema overrides --subquery and --subquery-name

actually we can mark --subquery-name as deprecated. and encourage people to use --schema

ianhe8x:
Will need to make sure the schema names are valid. check section 4.1.1 in https://www.postgresql.org/docs/9.2/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS
@wqsz7xn

packages/node/src/indexer/indexer.manager.ts
```diff
@@ -211,136 +293,56 @@ export class IndexerManager {
...
       const offsetValue = (this.getStartBlockFromDataSources() - 1).toString();
-      await this.storeService.setMetadata('blockOffset', offsetValue);
+      await metadataRepo.upsert({ key: 'blockOffset', value: offsetValue });
+    }
+
+    if (!keyValue.lastProcessedHeight) {
+      if (project) {
+        await metadataRepo.upsert({
+          key: 'lastProcessedHeight',
+          value: project.nextBlockHeight,
```
jiqiang90, line 309: I think this is the scenario when first time subqueryRepo have a nextBlockHeight but metadata table don't have lastProcessdHeight yet, and you try to upsert a value.

But I think this is very dangerous, let me give you an example.

Currently in subqueries table nextBlockHeight is 1000, and you have upsert this 1000 for lastProcessdHeight in meta, what if indexing encounter error at block 1000 and failed, when you re-start index, it will read from meta and start from height 1000+1 = 1001 , so the block 1000 has been skipped.

I rather not upsert the last processed height in here, only update this value when indexd(processed) completed.

ianhe8x:
Some tests are welcome for this PR, since it changed a lot which I don't think are covered by any existing tests. I might write some for you to begin with.

packages/node/src/indexer/indexer.manager.ts
wqsz7xn: > I think this is the scenario when first time subqueryRepo have a nextBlockHeight but metadata table don't have lastProcessdHeight yet, and you try to upsert a value.

That's right

> I rather not upsert the last processed height in here, only update this value when indexd(processed) completed.

The problem is that I need a block number to start indexing from and this is based on `lastProcessedHeight`. I'll try to look into how `main` handles this at the moment, I think it's with two different metadata entries.

wqsz7xn: @jiqiang90 
I'm slightly unconvinced that this is a problem. If a block fails to be indexed then the transaction rollsback which should also roll back the `nextBlockHeight` meaning that `nextBlockHeight` is only incremented on successfully indexed blocks.
https://github.com/subquery/subql/blob/a27be97028d64459d7fe895e71bc33198ad688a6/packages/node/src/indexer/indexer.manager.ts#L109-L129

ianhe8x: I think @jiqiang90 was right about
1) even you want to upsert it to a value it should be 
```
           value: project.nextBlockHeight - 1, 
```
2) for the first block to be processed, nextBlockHeight = startBlock from user's project.yaml, which doesn't mean startBlock - 1 was processed.

packages/node/src/configure/configure.module.ts
```diff
@@ -42,7 +44,33 @@ function defaultSubqueryName(config: Partial<IConfig>): MinConfig {
   } as MinConfig;
 }
 
-const logger = getLogger('configure');
+// Check if a subquery name is a valid schema name
+function validDbSchemaName(name: string): boolean {
+  if (name.length === 0) {
+    return false;
+  } else {
+    const regexp = new RegExp('[a-zA-Z_]');
```
ianhe8x, line 52: Can we make it more strict than the actual postgres's rule. like `/^[a-zA-Z_][a-zA-Z0-9_\-\/]{0,62}$/`

packages/node/src/configure/configure.module.ts
```diff
@@ -42,7 +44,33 @@ function defaultSubqueryName(config: Partial<IConfig>): MinConfig {
   } as MinConfig;
 }
 
-const logger = getLogger('configure');
+// Check if a subquery name is a valid schema name
+function validDbSchemaName(name: string): boolean {
```
ianhe8x, line 48: schema name is case insensitive, `toLowerCase()` before continue

packages/node/src/indexer/indexer.manager.ts
wqsz7xn: added a check on the `start` function that attempts to use the find the lastProcessedHeight and then falls back to the `nextBlockHeight` when that doesn't exit (exit otherwise). `lastProcessedHeight init has been removed from the `ensureMetadata` function as well.

packages/node/src/configure/configure.module.ts
wqsz7xn: Used your regex as I'm happy with it

packages/node/src/configure/configure.module.ts
wqsz7xn: done

packages/node/src/indexer/indexer.manager.ts
```diff
@@ -162,15 +163,33 @@ export class IndexerManager {
...
-    const lastProcessedHeight = (
-      await this.metadataRepo.findOne({ where: { key: 'lastProcessedHeight' } })
-    ).value as number;
-    void this.fetchService.startLoop(lastProcessedHeight + 1).catch((err) => {
+    let startHeight;
+    const lastProcessedHeight = await this.metadataRepo.findOne({
+      where: { key: 'lastProcessedHeight' },
+    });
+    if (lastProcessedHeight !== null) {
+      startHeight = (lastProcessedHeight.value as number) + 1;
```
ianhe8x, line 171: `       startHeight = Number(lastProcessedHeight.value) + 1;`
in case it is a string.

packages/node/src/configure/configure.module.ts
```diff
@@ -49,26 +49,22 @@ function validDbSchemaName(name: string): boolean {
   if (name.length === 0) {
     return false;
   } else {
-    const regexp = new RegExp('[a-zA-Z_]');
+    name = name.toLowerCase();
+    const regexp = new RegExp('^[a-zA-Z_][a-zA-Z0-9_\\-\\/]{0,62}$');
```
ianhe8x, line 53: can you add some tests to cover this?

wqsz7xn: will do

ianhe8x (approved):

xblad:
Hi guys, could it possible break a hello world example from `subql init --starter myProject`? :smiley_cat: 
My graphql engine keep saying that project with name app doesn't exist (when trying to spin up a nest.js). I've checked created postgres database in docker and it has no schema `app` whatsoever, only `public` with couple of tables. Which module is responsible for `app` schema creation? Should it be `subquery-node`?

wqsz7xn:
@xblad 
I offered a fix on the discord with an explanation: https://discord.com/channels/796198414798028831/809621232361668609/925571758143180860

PR #684, opened 2021-11-30
Cute codegen

wqsz7xn:
- make the codegen use cli-ux
- up to you if you want this in or if you want the previous console.log banner, just noticed that the rest of the ui was using cli-ux

### DIFF
--- a/packages/cli/src/commands/codegen.ts
+++ b/packages/cli/src/commands/codegen.ts
@@ -16,10 +16,6 @@ export default class Codegen extends Command {
 
   async run(): Promise<void> {
     const {flags} = this.parse(Codegen);
-    this.log('===============================');
-    this.log('---------Subql Codegen---------');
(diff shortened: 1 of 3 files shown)


packages/cli/src/commands/codegen.ts
```diff
@@ -16,10 +16,6 @@ export default class Codegen extends Command {
 
   async run(): Promise<void> {
     const {flags} = this.parse(Codegen);
-    this.log('===============================');
```
ianhe8x, line 19: Can you replace this via cli-ux?

ianhe8x:
Can you put some screenshots here?

packages/cli/src/commands/codegen.ts
wqsz7xn: I'll see if I can do that

PR #687, opened 2021-12-01
Now poi will skip create for null transaction

jiqiang90:


packages/node/src/indexer/indexer.manager.ts
```diff
@@ -47,6 +48,7 @@ import { BlockContent } from './types';
 const { version: packageVersion } = require('../../package.json');
 
 const DEFAULT_DB_SCHEMA = 'public';
+const NULL_MERKEL_ROOT = hexToU8a('0x00');
```
ianhe8x, line 51: it is already defined somewhere right? can you group them together into a consts.ts

ianhe8x (approved):

PR #690, opened 2021-12-02
Support block 0

shunjizhan:
## Change
Changed default start block to 0, and make indexer just ignore this block since the genesis block data is different format (and don't need to index). This will enable subql-node to run with a fresh substrate node which is started with `--instant-sealing` flag.

## Test
- `yarn test` all passed
- tested locally with `--instant-sealing` mandala node, and subql-node won't crash, just wait there listening to new block, when new block comes in, it will index new blocks.
- tested locally with mandala node that has transactions, still able to index all blocks.

### DIFF
--- a/packages/node/src/configure/project.model.ts
+++ b/packages/node/src/configure/project.model.ts
@@ -41,9 +41,9 @@ export class SubqueryProject {
     this._path = path;
 
     manifest.dataSources?.forEach(function (dataSource) {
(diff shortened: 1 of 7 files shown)


jiqiang90:
I think we should keep the default start block height to 1.  

Change the start block height will impact on the project currently running with Proof-of-index service, its block offset value and calculated MMR value.

shunjizhan:
sounds good, I can just use some local workarounds

PR #697, opened 2021-12-08
Fix indexer for new projects

wqsz7xn:
Made a mistake where I forgot to seed the initial `lastProcessedHeight` when creating a new project. Was causing `nextBlockHeight` error

### DIFF
--- a/packages/node/src/indexer/indexer.manager.ts
+++ b/packages/node/src/indexer/indexer.manager.ts
@@ -173,13 +173,11 @@ export class IndexerManager {
       const project = await this.subqueryRepo.findOne({
         where: { name: this.nodeConfig.subqueryName },
       });
-      assert(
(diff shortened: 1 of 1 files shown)


packages/node/src/indexer/indexer.manager.ts
```diff
@@ -325,6 +323,15 @@ export class IndexerManager {
 
     // blockOffset and genesisHash should only have been created once, never updated.
     // If blockOffset is changed, will require re-index and re-sync poi.
+
+    // Project is new and requries a starting block
+    if (!project && !keyValue.lastProcessedHeight) {
```
ianhe8x, line 328: As we discussed, for a new project, mark the startBlock-1 as processed is weird since it is against the truth.

ianhe8x (approved):

PR #698, opened 2021-12-10
Add support for loading chaintypes from js

lachlan1m:
- [X] Can load js chaintypes file
- [x] Writing tests


packages/common/src/project/load.ts
```diff
@@ -3,42 +3,55 @@
...
+import {NodeVM, VMScript} from '@subql/x-vm2';
 import {plainToClass} from 'class-transformer';
 import {validateSync} from 'class-validator';
 import yaml from 'js-yaml';
 import {ChainTypes} from './models';
 import {ProjectManifestVersioned, VersionedProjectManifest} from './versioned';
 
-export function loadFromJsonOrYaml(file: string): unknown {
-  const {ext} = path.parse(file);
+export function loadFromFile(filePath: string): unknown {
```
ianhe8x, line 13: Doesn't feel right here. Common is shared between subql-node and subql-query, but apparently subql-query doesn't need to use sandbox to load the js chain type.

ianhe8x (changes requested):

packages/common/src/project/load.ts
ianhe8x: What `loadFromFile()` here concerns is read a file and return its content. It is subql-node to decide how to use the file content, in this case if it's a .js, subql-node need to interpret it via sandbox.
So you can see  the dependency of `@subql/x-vm2` is always in subql-node, shouldn't be extended to common.

lachlan1m: The reason I did this is validator/src/readers/local-reader.ts used `loadFromJsonOrYaml()`

lachlan1m: The reason I didn't do it this way is that validator/src/readers/local-reader.ts also uses `loadFromFile()`. It may not be necessary for validator to be able to do this. Let me know.

ianhe8x: The validator is not an essential component, if that's the case we can skip validating chaintypes for js like chaintype.

ianhe8x: let's leave it in common for now. But you need to add more restrict rules for the sandbox.

lachlan1m: sure, just wanted to get some feedback.

packages/common/src/project/load.ts
```diff
@@ -19,10 +19,15 @@ export function loadFromFile(filePath: string): unknown {
 
   if (ext === '.js') {
     const vm = new NodeVM({
+      console: 'inherit',
+      wasm: false,
+      sandbox: {},
       require: {
```
ianhe8x, line 25: add `builtin: ['assert', 'buffer', 'crypto', 'util', 'path'],`

packages/common/src/project/load.ts
```diff
@@ -3,42 +3,60 @@
...
+      sandbox: {},
+      require: {
+        external: true,
+        context: 'sandbox',
+      },
+      wrapper: 'commonjs',
+      sourceExtensions: ['js', 'cjs'],
+    });
+
+    const script = new VMScript(`module.exports = require('./${base}');`, filePath);
```
ianhe8x, line 33: you can assume `base` starts with `./`

ianhe8x, line 33:
```suggestion
    const script = new VMScript(`module.exports.default = require('./${base}').default;`, filePath);
```
to ignore non-default exports.

packages/common/src/project/load.ts
```diff
@@ -3,42 +3,60 @@
...
+
+    return vm.run(script) as unknown;
```
ianhe8x, line 35: you can add some logic here to check the keys are in our whitelist.

packages/common/src/project/load.ts
```diff
@@ -3,42 +3,63 @@
...
+      console: 'inherit',
+      wasm: false,
+      sandbox: {},
+      require: {
+        external: true,
+        builtin: ['assert', 'buffer', 'crypto', 'util', 'path'],
+        context: 'sandbox',
+      },
+      wrapper: 'commonjs',
+      compiler: (code: string, filename: string) => ts.transpile(code, {}, filename),
```
lachlan1m, line 33: This may be a bad idea, It does work though and is less work for the user.

packages/common/src/project/load.ts
lachlan1m: I have `const script = new VMScript(`module.exports = require('${filePath}').default;`, filePath).compile();` and a test which will test non-default exports

lachlan1m: const script = new VMScript(`module.exports = require('${filePath}').default;`, filePath);
It will ignore non-default exports (have written tests).

packages/common/src/project/load.ts
lachlan1m: This logic exists in a parseChainTypes() and I have written a test to check that an exception is thrown if extra keys are added.

jiqiang90 (approved):

ianhe8x (approved):

Issue #695, opened 2021-12-07
adding request payload into log in query service


zhex:
now the query service only writes `request completed` message into the log, the graphql payload is not included. it is really hard to debug in a production environment.  we can add payload into log data to improve the debugging experience.

ianhe8x:
related to #633

ianhe8x:
closed by #633

Issue #633, opened 2021-11-15
Improve subql/query logging


wqsz7xn:
Logging can be significantly improved to help both people running a playground and statistics for hosted services. Ideally we would like to have the following exposed when at debug logging

- project key
- source IP
- referral domain
- user agent
- request duration (ms)
- response size (bytes)
- response success (bool)
- query methods: string[]

Proof of concept:
![image](https://user-images.githubusercontent.com/67990210/141866163-3a6bd2ef-664f-4ea3-8887-b9a98adb9e2d.png)

Found here:
https://github.com/wqsz7xn/subql/tree/improved-logging-poc
https://www.apollographql.com/docs/apollo-server/integrations/plugins/

ianhe8x:
We have a `--profiler` flag to help developer identify the bottleneck. Don't feel we should enable that by default, it could flood the console outputs.

wqsz7xn:
Profiler information is available to the subql/node only right? We need some profiling for fulfilling queries from subql/query as well. This would all be hidden behind a debug logging level so i don't think there's a concern about flooding logs

ianhe8x:
This is useful. Please continue with a PR. Can you merge it with the current `request completed`? Also can you handle `--output-fmt` flag, we usually use json for backend log processing, plain text for local dev.

ianhe8x:
Closed by #742

PR #703, opened 2021-12-12
Improve query logging

lachlan1m:
For each graphql request a single json blob will be logged containing any useful information.

### DIFF
--- a/packages/query/src/graphql/plugins/logGraphql.ts
+++ b/packages/query/src/graphql/plugins/logGraphql.ts
@@ -33,7 +33,11 @@ export const LogGraphqlPlugin: ApolloServerPlugin = {
   ): Promise<GraphQLRequestListener<MyApolloContext>> {
     const start = performance.now();
 
-    //IntrospectionQuery clutters logs and isn't useful
+    if (requestContext.request.operationName) {
+      logger.debug(`operation: ${requestContext.request.operationName}`);
+    }
+
+    //IntrospectionQuery payload clutters logs and isn't useful
     if (requestContext.request.operationName !== 'IntrospectionQuery') {
       logger.debug('graphql payload: \n' + requestContext.request.query);
     }
--- a/packages/query/src/graphql/plugins/logGraphql.ts
+++ b/packages/query/src/graphql/plugins/logGraphql.ts
@@ -10,14 +10,9 @@ import {getLogger} from '../../utils/logger';
 const logger = getLogger('graphql');
 
 const getSizeInBytes = (obj: Record<string, unknown>) => {
-  let str: string;
-  if (typeof obj === 'string') {
-    str = obj;
-  } else {
-    str = JSON.stringify(obj);
-  }
-  const bytes = new TextEncoder().encode(str).length;
-  return bytes;
+  console.log(obj);
+  const str = JSON.stringify(obj);
+  return new TextEncoder().encode(str).length;
 };
 
 const logSizeInKilobytes = (description: string, obj: Record<string, unknown>) => {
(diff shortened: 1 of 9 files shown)


lachlan1m:
#695 #633

packages/query/src/graphql/plugins/logGraphql.ts
```diff
@@ -0,0 +1,71 @@
...
+import {ApolloServerPlugin} from 'apollo-server-plugin-base';
+import {GraphQLRequestListener} from 'apollo-server-plugin-base/src/index';
+import {GraphQLRequestContext} from 'apollo-server-types';
+import {getLogger} from '../../utils/logger';
+
+const logger = getLogger('graphql');
+
+const getSizeInBytes = (obj: Record<string, unknown>) => {
+  let str: string;
+  if (typeof obj === 'string') {
```
zhex, line 14: the obj is Record type, is it still possible to be string?

lachlan1m: made a mistake here, forgot to update.

packages/query/src/graphql/plugins/logGraphql.ts
```diff
@@ -0,0 +1,71 @@
...
+                //Warn as no metadata field should be null
+                logger.warn(`${info.parentType.name}.${info.fieldName} returned null`);
+              }
+            };
+          },
+        };
+      },
+      async willSendResponse({response}) {
+        if (!response.errors) {
+          logger.debug(logSizeInKilobytes('response size:', response.data));
```
ianhe8x, line 60: I don' think we want multiple lines for one request. Can you put everything about one request in one line?
and maybe format it as json so backend can more easily consume it. @zhex which formate do you prefer?

zhex: in the production, the debug level will be turned off, so it is good to add response data in the existing info log payload

lachlan1m: I have 1 line for graphql payload (doesn't convert nicely to json) and 1 json object for response data.

packages/query/src/graphql/plugins/logGraphqlPlugin.ts
```diff
@@ -6,26 +6,26 @@ import {ApolloServerPlugin} from 'apollo-server-plugin-base';
...
-  const str = JSON.stringify(obj);
-  return new TextEncoder().encode(str).length;
-};
+const logger = getLogger('graphql');
 
 const getSizeInKilobytes = (obj: Record<string, unknown>) => {
-  const bytes = getSizeInBytes(obj);
-  const kb = (bytes / 1000).toFixed(2);
-  return kb;
+  const str = JSON.stringify(obj);
```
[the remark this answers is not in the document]
ianhe8x: I think the better place to get the response size is from express response object. Though the down side is you will have to refactor the code.

packages/query/src/graphql/plugins/logGraphqlPlugin.ts
```diff
@@ -0,0 +1,74 @@
...
+    if (httpHeaders) {
+      payload['x-forwarded-for'] = httpHeaders['x-forwarded-for'] ?? '127.0.0.1';
+      payload['user-agent'] = httpHeaders['user-agent'] ?? 'unknown';
+      payload.referer = httpHeaders.referer ?? 'unknown';
+    }
+
+    return {
+      async willSendResponse({response}) {
+        payload['response-time'] = `${Math.round(performance.now() - start)}`;
+        payload['response-payload-size'] = getSizeInKilobytes(response.data);
```
ianhe8x, line 56: try this `response.http.headers.get('Content-Length')` though i'm not sure if it is set in this stage.

packages/query/src/graphql/plugins/logGraphqlPlugin.ts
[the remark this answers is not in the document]
ianhe8x: I saw in line 67 the response is stringified again, if `JSON.stringify` is inevitable, at least reduce it to once.

packages/query/src/graphql/plugins/logGraphqlPlugin.ts
lachlan1m: My solution so far is to use response.http.headers.set(graphqlData), and then use ExpressPinoLogger with custom serializers.

Issue #716, opened 2021-12-21
`subql build` entry use export in package.json


jiqiang90:
currently `subql build` will only build from `src/index.ts` by default,  and we not support any alternative entries. 
We suggest to use `export` from package.json as the build entry.

lachlan1m:
#753

PR #718, opened 2021-12-22
Add cli build exports field alternative

lachlan1m:
#716


packages/cli/src/commands/build.ts
```diff
@@ -35,10 +35,21 @@ export default class Build extends Command {
 
     // Get the output location from the project package.json main field
     const pjson = JSON.parse(readFileSync(path.join(directory, 'package.json')).toString());
+
+    let buildPath = 'src/index.ts';
```
ianhe8x, line 39: read pjson.main for default build path.

ianhe8x: what is the type of buildPath, it could be a array now, because with this PR we support multiple entries.

Issue #726, opened 2021-12-29
Expand health service


wqsz7xn:
Currently the health service endpoint is very primitive, only supporting a [catch all internal server error for its endpoint](https://github.com/subquery/subql/blob/3553be9593bf29bda68271896f77af31198f4e7b/packages/node/src/meta/health.controller.ts). This health service should be expanded and be made more accurate (ie: less than a 10 block time window as is currently implemented below and with proper http status codes) https://github.com/subquery/subql/blob/cb7ebc814db13469472c54a658282ae5119d3029/packages/node/src/meta/health.service.ts#L71-L85

I personally want to be able to check the health service to see if the node is in a startup state or indexing state so that I can add a healthcheck within a docker image so that the query service can start after the indexer is properly started up and **available** (has created its schemas and is indexing _versus_ has just started up and schemas not created (if it's a new project)). 

Currently the following[ docker image ](https://gist.github.com/wqsz7xn/83348e1f84a9e5789b6b725859c82621)against a starter project will start the query service **immediately** after the node has started. As described above this is not good because the query service shows some errors and will crash and restart. Then after a while query service will successfully start (after the node service has properly gone through its start sequence). In order to achieve this we must be returning json objects and then docker side decode with something like `jq`

ianhe8x:
Looks like the health check in docker-compose works like readiness probe in Kubernetes, while the health check we implemented is liveness probe in Kubernetes.
We can do that in the new `/ready` rpc.

ianhe8x:
Closed by #728

PR #728, opened 2021-12-29
Ready endpoint

wqsz7xn:
A ready endpoint separate from the health endpoint which indicates that all schemas have been created and the node is ready for the query service to begin indexing against created db schemas. Includes additional docker include of `curl` so that query docker image can get status of endpoint for a healthcheck before starting itself within docker-compose.

Partially implements #726 for functionality

Todo:
- [ ] Modify the starter projects `docker-compose.yml` using the [following healthcheck](https://gist.github.com/wqsz7xn/2afe714f587b9f19c976859ece629e05) to ensure proper timing of starting docker images. This pr should be merged before making this modification though

### DIFF
--- a/packages/node/src/meta/ready.service.ts
+++ b/packages/node/src/meta/ready.service.ts
@@ -7,18 +7,18 @@ import { EventPayload, IndexerEvent } from '../indexer/events';
 
 @Injectable()
 export class ReadyService {
-  private ready: boolean;
+  private _ready: boolean;
 
   constructor() {
-    this.ready = false;
+    this._ready = false;
   }
 
   @OnEvent(IndexerEvent.Ready)
   handleReady({ value }: EventPayload<boolean>): void {
-    this.ready = value;
+    this._ready = value;
   }
 
-  get getReady() {
-    return this.ready;
+  get ready() {
+    return this._ready;
   }
 }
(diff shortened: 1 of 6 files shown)


ianhe8x (approved):

packages/node/src/meta/ready.service.ts
```diff
@@ -0,0 +1,24 @@
...
+  constructor() {
+    this.ready = false;
+  }
+
+  @OnEvent(IndexerEvent.Ready)
+  handleReady({ value }: EventPayload<boolean>): void {
+    this.ready = value;
+  }
+
+  get getReady() {
```
ianhe8x, line 21:
```suggestion
  get ready() {
```
getReady suggests it is a method rather a property

ianhe8x (changes requested):

packages/node/src/meta/ready.service.ts
wqsz7xn: this is a duplicate identifier because we already have a property in this named `ready` and thus can't have a getter with the same name. I considered using underscore notation but we don't use it elsewhere in the codebase

ianhe8x: underscore looks good to me

ianhe8x (approved):

PR #729, opened 2021-12-29
remove stable version, bump prerelease

wqsz7xn:
while https://github.com/subquery/subql/pull/719 hasn't been merged we still have to do this D:

### DIFF
--- a/packages/node/package.json
+++ b/packages/node/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@subql/node",
-  "version": "0.26.1-0",
+  "version": "0.26.1-1",
   "description": "",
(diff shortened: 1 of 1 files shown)


packages/node/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@subql/node",
-  "version": "0.26.1-0",
+  "version": "0.26.1-1",
```
ianhe8x, line 3: don't bump here, let the pipeline do this

ianhe8x (changes requested):

ianhe8x (approved):

Issue #727, opened 2021-12-29
Spec validation error is eaten and only visible with --debug


ianhe8x:
When there's type mismatch in network.chaintypes, no error printed to console and the whole app just crashed, unless with --debug enabled.

The problem here is the validation of chaintypes is done in the SubqueryProject's `get chainTypes()`, the better place is `ProjectManifestVersioned -> validate()`

wqsz7xn:
I don't think that we should move this to validate as it seems geared towards the project yaml we could expand it though? Could just add the raw parse to the try block

```typescript
      let rawChainTypes: unknown;
      let parsedChainTypes: ChainTypes;
      try {
        rawChainTypes = loadChainTypes(
          path.join(this._path, impl.network.chaintypes.file),
          this.path,
        );
        parsedChainTypes = parseChainTypes(rawChainTypes);
      } catch (e) {
        logger.error(`failed to load chaintypes file, ${e}`);
      }
      return parsedChainTypes;
    }
```

Which nicely displays the error:
```
2021-12-31T02:06:51.725Z <configure> ERROR failed to load chaintypes file, Error: failed to parse chain types.
```

wqsz7xn:
Fixed in above pull, although we may want to move the validation logic as described in the issue in the future

PR #732, opened 2022-01-04
Display chaintypes error

wqsz7xn:
Display a swallowed chaintypes error that was only visible from `--debug`. Quick fix for https://github.com/subquery/subql/issues/727. Although doesn't move the validation logic to `ProjectManifestVersioned -> validate()`.

In order to fix this further down the line we should consider adding validation as a prebuild step to the starter project so that invalid chain types result in a build error. Enhancement ticket about that here https://github.com/subquery/subql-starter/issues/38

### DIFF
--- a/packages/node/src/configure/project.model.ts
+++ b/packages/node/src/configure/project.model.ts
@@ -11,6 +11,7 @@ import {
   manifestIsV0_0_1,
   manifestIsV0_2_0,
   loadChainTypes,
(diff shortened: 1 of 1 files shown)


ianhe8x (approved):

ianhe8x:
Good for now. @jiqiang90 when you work on the other refactory task, don't forget rebase this commit.

Issue #731, opened 2022-01-04
check manifest network and metadata stored genesis hash


jiqiang90:
We have encounter this error when user have a deployment running , and user changed network endpoint with a same spec name, but with different finalized head height and different genesis hash (we can see this as different new network)

**The update for deployment**: https://github.com/parallel-finance/moonbeam-query-service/commit/69dda9e254558d8ec3e061b540997c3da84a9a3c

**Error:** <img width="1066" alt="Screen Shot 2022-01-04 at 9 28 03 PM" src="https://user-images.githubusercontent.com/10431657/148036513-3e56b4ee-0559-47d5-866a-834601631bb2.png">

**Cause:** 
This error is due to the previous deployment metadata last processed height is 8356730, and when user update deployment the new endpoint current finalized height height is 247853,  which 8356730 is not exist in their network yet. 

**TODO:** 
However it should not even try to fetch block 8356730, as it should fail network and metadata validation,  even network spec name are same, their genesis hash must match to make sure this the same network,  it did't throw error after user updates, we need to look into this.

PR #735, opened 2022-01-05
Detect changed project metadata

wqsz7xn:
Fixes: https://github.com/subquery/subql/issues/731
Unhandled cases where the project has run and has stored metadata. Project then changes manifest entry (genesis hash, spec version or chain) and the project will not detect the changes made and attempt (and fail at) indexing. This pull now creates a detection for these metadata fields and exits the node with an assert. 

 Tested with recreating the scenario in issue linked

### DIFF
--- a/packages/node/src/indexer/indexer.manager.ts
+++ b/packages/node/src/indexer/indexer.manager.ts
@@ -342,27 +342,17 @@ export class IndexerManager {
       }
     } else {
       assert(
-        this.project.network === keyValue.genesisHash,
+        this.project.network.genesisHash === keyValue.genesisHash,
         'Specified project manifest genesis hash does not match database stored genesis hash, consider cleaning project schema using --force-clean',
       );
     }
 
     if (keyValue.chain !== chain) {
       await metadataRepo.upsert({ key: 'chain', value: chain });
-    } else {
-      assert(
-        chain === keyValue.chain,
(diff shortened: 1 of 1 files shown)


packages/node/src/indexer/indexer.manager.ts
```diff
@@ -340,14 +340,29 @@ export class IndexerManager {
       } else {
         await metadataRepo.upsert({ key: 'genesisHash', value: genesisHash });
       }
+    } else {
+      assert(
+        this.project.network === keyValue.genesisHash,
```
ianhe8x, line 345:
```suggestion
        project.networkGenesis === keyValue.genesisHash,
```

packages/node/src/indexer/indexer.manager.ts
```diff
@@ -340,14 +340,29 @@ export class IndexerManager {
...
+        'Specified project manifest genesis hash does not match database stored genesis hash, consider cleaning project schema using --force-clean',
+      );
     }
 
     if (keyValue.chain !== chain) {
       await metadataRepo.upsert({ key: 'chain', value: chain });
+    } else {
+      assert(
+        chain === keyValue.chain,
```
ianhe8x, line 354: Only need to do the check for genesisHash, because chain and specName could be different after a chain upgrade.

ianhe8x (approved):

Issue #734, opened 2022-01-04
add row estimate in _metadata


ianhe8x:
In UI we show a total entities number which is fetched from graphql endpoint by adding totalCount of each entity.
For a large table doing this might slow down db significantly. 
One workaround is use count estimate provided by postgres in this case.

to be more specific:
```
type _Metadata {
        lastProcessedHeight: Int
        lastProcessedTimestamp: Date
        targetHeight: Int
        chain: String
        specName: String
        genesisHash: String
        indexerHealthy: Boolean
        indexerNodeVersion: String
        queryNodeVersion: String
        rowCountEstimate: JSONObject # {"table1": xxx, "table2": yyy}
      }


```

links: https://wiki.postgresql.org/wiki/Count_estimate

wqsz7xn:
Added in #736

PR #736, opened 2022-01-05
Row count estimate

wqsz7xn:
Adds a row count estimate to the metadata plugin for query service. Although it uses a custom type rather than a generic json object.
Implements: https://github.com/subquery/subql/issues/734
Example:
```gql
query {
  _metadata {
    rowCountEstimate {
      table,
      estimate
    }
  }
}
```
```
{
  "data": {
    "_metadata": {
      "rowCountEstimate": [
        {
          "table": "starter_entities",
          "estimate": 195220
        },
        {
          "table": "_metadata",
          "estimate": 9
        }
      ]
    }
  }
}
```

### DIFF
--- a/packages/query/src/graphql/plugins/GetMetadataPlugin.ts
+++ b/packages/query/src/graphql/plugins/GetMetadataPlugin.ts
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import {URL} from 'url';
-import {MetaData} from '@subql/common';
+import {MetaData, TableEstimate} from '@subql/common';
 import {makeExtendSchemaPlugin, gql} from 'graphile-utils';
 import fetch, {Response} from 'node-fetch';
 import {setAsyncInterval} from '../../utils/asyncInterval';
@@ -73,7 +73,7 @@ async function fetchFromTable(pgClient: any, schemaName: string): Promise<MetaDa
 
   metadata.queryNodeVersion = packageVersion;
 
-  const xs = await pgClient
+  const tableEstimates: [TableEstimate] = await pgClient
     .query(
       `select relname, reltuples::bigint as estimate from pg_class where relname in (select table_name from information_schema.tables where table_schema = '${schemaName}')`
     )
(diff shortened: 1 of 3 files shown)


packages/query/src/graphql/plugins/GetMetadataPlugin.ts
```diff
@@ -72,6 +73,17 @@ async function fetchFromTable(pgClient: any, schemaName: string): Promise<MetaDa
 
   metadata.queryNodeVersion = packageVersion;
 
+  const tableEstimates: [TableEstimate] = await pgClient
+    .query(
+      `select relname, reltuples::bigint as estimate from pg_class where relname in (select table_name from information_schema.tables where table_schema = '${schemaName}')`
```
ianhe8x, line 78: use query parameter

packages/query/src/graphql/plugins/GetMetadataPlugin.ts
```diff
@@ -72,6 +73,17 @@ async function fetchFromTable(pgClient: any, schemaName: string): Promise<MetaDa
...
+    )
+    .then(({rows}) => rows.map(({estimate, relname}) => ({table: relname, estimate: estimate})))
```
ianhe8x, line 80: if you set alias in the sql, you can save this map loop.

packages/query/src/graphql/plugins/GetMetadataPlugin.ts
```diff
@@ -73,22 +73,27 @@ async function fetchFromTable(pgClient: any, schemaName: string): Promise<MetaDa
...
 
-  metadata.rowCountEstimate = tableEstimates;
+  metadata.rowCountEstimate = tableEstimates.rows;
 
   return metadata;
 }
 
 export const GetMetadataPlugin = makeExtendSchemaPlugin((build, options) => {
-  const schemaName = options.pgSchemas;
+  const [schemaName] = options.pgSchemas;
```
ianhe8x, line 96: There wasn't any error before?

ianhe8x (approved):

packages/query/src/graphql/plugins/GetMetadataPlugin.ts
wqsz7xn: Nope. It gets cast to `string` fine but it was really `[string]`. Can't trust the types with a whole lot of `any` being used here, but the plugin needs it so no avoiding.

wqsz7xn:
Seems my chaintypes error pull that has been merged is causing problems with the CI. I'll fix this by either removing the process exit and moving it to the validate or mocking process exit in that test.

ianhe8x:
@wqsz7xn I can merge this one and you fix it in another PR.

PR #738, opened 2022-01-06
Fix CI: Revert chaintypes change and move check to API module

wqsz7xn:
Move chaintypes process exit to api module so that jest doesn't die


ianhe8x (approved):
It still feels not right, that we make a getter failible, though since Jay is refactoring this part, I will merge this for now.

wqsz7xn:
Agreed; getters shouldn't be doing this sort of processing and it needs to be split off into other functions. Perhaps add validation in the constructor or something?

PR #739, opened 2022-01-09
Change node default port

wqsz7xn:
- Changes the default port of the node service to `3001` and keeps the default port of the query service as `3000`
- Reason for decision above is that people will be used to going to localhost:3000 in their browsers so we don't want to change this
- Adds a log for the URL of the playground as a shortcut
- Add strict options to `getYargsOptions` so that invalid flags are not swallowed
- **I need to change the docker images on the templates so that they support the new ports so let me put that up before this is merged**

### DIFF
--- a/packages/common/src/project/utils.ts
+++ b/packages/common/src/project/utils.ts
@@ -11,6 +11,8 @@ import {
   SubqlRuntimeDatasource,
 } from '@subql/types';
 
+import detectPort from 'detect-port';
+
 export function isBlockHandlerProcessor<T extends SubqlNetworkFilter, E>(
   hp: SecondLayerHandlerProcessor<SubqlHandlerKind, T, unknown>
 ): hp is SecondLayerHandlerProcessor<SubqlHandlerKind.Block, T, E> {
@@ -35,4 +37,26 @@ export function isCustomDs<F extends SubqlNetworkFilter>(ds: SubqlDatasource): d
 
 export function isRuntimeDs(ds: SubqlDatasource): ds is SubqlRuntimeDatasource {
   return ds.kind === SubqlDatasourceKind.Runtime;
+}
+
+export async function findAvailablePort(startPort: number, range = 10): Promise<number> {
+  for (let p = startPort; p < startPort + range; p++) {
+    const candidatePort = await detectPort(p)
+      .then((_port) => {
+        if (p === _port) {
+          return p;
+        } else {
+          return null;
+        }
+      })
+      .catch((err) => {
+        console.error(err);
+      });
+
+    if (candidatePort) {
+      return candidatePort;
+    }
+  }
+
+  return null;
 }
--- a/packages/common/src/project/utils.ts
+++ b/packages/common/src/project/utils.ts
@@ -41,11 +41,13 @@ export function isRuntimeDs(ds: SubqlDatasource): ds is SubqlRuntimeDatasource {
 
 export async function findAvailablePort(startPort: number, range = 10): Promise<number> {
   for (let port = startPort; port <= startPort + range; port++) {
-    const _port = await detectPort(port).catch(() => {
+    try {
+      const _port = await detectPort(port);
+      if (_port === port) {
+        return port;
+      }
+    } catch (e) {
       return null;
-    });
-    if (_port === port) {
-      return port;
     }
   }
 
--- a/packages/common/src/project/utils.ts
+++ b/packages/common/src/project/utils.ts
@@ -40,21 +40,12 @@ export function isRuntimeDs(ds: SubqlDatasource): ds is SubqlRuntimeDatasource {
 }
 
 export async function findAvailablePort(startPort: number, range = 10): Promise<number> {
-  for (let p = startPort; p < startPort + range; p++) {
-    const candidatePort = await detectPort(p)
-      .then((_port) => {
-        if (p === _port) {
-          return p;
-        } else {
-          return null;
-        }
-      })
-      .catch((err) => {
-        console.error(err);
-      });
-
-    if (candidatePort) {
-      return candidatePort;
+  for (let port = startPort; port <= startPort + range; port++) {
+    const _port = await detectPort(port).catch(() => {
+      return null;
+    });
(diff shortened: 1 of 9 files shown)


ianhe8x:
@wqsz7xn I didn't notice that you have added strictOptions to subql-node. I have a concern that this might break something from hosted service, because with this option, they have to handle the args in different subql version in a precious manner now.

@zhex

wqsz7xn:
@ianhe8x @zhex 
I thought `strictOptions` would be good to have because sometimes when typing out a longish command you make a mistake and it gets swallowed and starts (with the bad configuration). But I didn't know that hosted uses custom options. Let me know if you want to revert the strict options stuff, we can maybe workaround with a warning or just drop this all together

zhex:
I think both changes will affect hosted service. all the existing deployments in the hosted service are running on the default port, which means we need to force set a port after this version is released. The hosted service need to support different versions in one template, `restrictOptions` might cause some legacy options stop working.

wqsz7xn:
`strictOptions` has been removed and the default port of 3000 added re-added a fallback for query. Is there anything that I can add on my side to help the hosted team transition the new node port?

ianhe8x:
@wqsz7xn how about instead of change the port to 3001, we leave it 3000 unchanged but detect when there's port conflict, we increase 1 until finding an available port.

wqsz7xn:
> how about instead of change the port to 3001, we leave it 3000 unchanged but detect when there's port conflict, we increase 1 until finding an available port.

Now that I think about it this is a bit weird and would mean that ports are different across environments, also weird that we are port scanning on one service but not the other (we could do both). How about we just use some other unbound port? We have a envvar in place if people want to change the port themselves.

ianhe8x:
create-react-app use this approach. https://github.com/facebook/create-react-app/blob/f0a837c1f07ebd963ddbba2c2937d04fc1b79d40/packages/react-dev-utils/WebpackDevServerUtils.js

This will only be used when no port is specified via env or args.

wqsz7xn:
We have decided against doing it the createreactapp

packages/common/src/project/utils.ts
```diff
@@ -36,3 +38,25 @@ export function isCustomDs<F extends SubqlNetworkFilter>(ds: SubqlDatasource): d
...
+    const candidatePort = await detectPort(p)
+      .then((_port) => {
+        if (p === _port) {
+          return p;
+        } else {
+          return null;
+        }
+      })
+      .catch((err) => {
+        console.error(err);
```
ianhe8x, line 53: clean up console.error, or print only when --debug is used.

packages/common/src/project/utils.ts
```diff
@@ -36,3 +38,25 @@ export function isCustomDs<F extends SubqlNetworkFilter>(ds: SubqlDatasource): d
 export function isRuntimeDs(ds: SubqlDatasource): ds is SubqlRuntimeDatasource {
   return ds.kind === SubqlDatasourceKind.Runtime;
 }
+
+export async function findAvailablePort(startPort: number, range = 10): Promise<number> {
+  for (let p = startPort; p < startPort + range; p++) {
+    const candidatePort = await detectPort(p)
+      .then((_port) => {
```
ianhe8x, line 45: don't combine `await` with `then`

packages/node/src/main.ts
```diff
@@ -11,7 +12,18 @@ const logger = getLogger('subql-node');
 
 async function bootstrap() {
   const debug = argv('debug');
-  const port = argv('port') as number;
+  const candidatePort = argv('port') as number;
+  const port = await findAvailablePort(candidatePort);
```
ianhe8x, line 16: I think it would be more expected if we only do port scan when --port is not given (default port is used).

packages/query/src/main.ts
```diff
@@ -12,5 +13,18 @@ void (async () => {
   const app = await NestFactory.create(AppModule, {
     logger: new NestLogger(),
   });
-  await app.listen(process.env.PORT ?? argv.port ?? 3000);
+  const candidatePort = parseInt(process.env.PORT) ?? argv.port;
+  const port = await findAvailablePort(candidatePort);
```
ianhe8x, line 17: same here

packages/node/src/main.ts
wqsz7xn: moved default port outside of yargs so that we can detect if user is using flag

packages/query/src/main.ts
wqsz7xn: same as above

ianhe8x (approved):

packages/query/src/graphql/graphql.module.ts
```diff
@@ -84,6 +86,10 @@ export class GraphqlModule implements OnModuleInit, OnModuleDestroy {
       cors: true,
     });
 
+    if (this.config.get('playground')) {
+      logger.info('Started playground at http://localhost:3000');
```
ianhe8x, line 90: Print the actual used port.

ianhe8x (changes requested):

packages/query/src/graphql/graphql.module.ts
wqsz7xn: oh yeah

packages/common/src/project/utils.ts
```diff
@@ -36,3 +38,16 @@ export function isCustomDs<F extends SubqlNetworkFilter>(ds: SubqlDatasource): d
 export function isRuntimeDs(ds: SubqlDatasource): ds is SubqlRuntimeDatasource {
   return ds.kind === SubqlDatasourceKind.Runtime;
 }
+
+export async function findAvailablePort(startPort: number, range = 10): Promise<number> {
+  for (let port = startPort; port <= startPort + range; port++) {
+    const _port = await detectPort(port).catch(() => {
```
ianhe8x, line 44:
```suggestion
try{
    const _port = await detectPort(port)
    ...
    } catch(e) {
    return null;
    }
```

ianhe8x (approved):

PR #740, opened 2022-01-10
Edit interval metadata

wqsz7xn:
- Changes the data that is updated during the interval, `lastProcessedHeight` and `lastProcessedTimestamp` must be tied to a transaction otherwise it can lead to an inconsistent db state. The `targetHeight` remains tied to an interval as this is the finalized target height of the chain that you are indexing. Db inconsistency is currently created in the following way:
```
block is processed --> interval event updates metadata --> block is rolled back but interval event updates are not
```
- Should fix a critical issue that we are facing with the moonbeam dictionary project

### DIFF
--- a/packages/node/src/indexer/indexer.manager.ts
+++ b/packages/node/src/indexer/indexer.manager.ts
@@ -108,11 +108,9 @@ export class IndexerManager {
         }
       }
 
-      this.storeService.setMetadata(
-        'lastProcessedHeight',
-        block.block.header.number.toNumber(),
-        { transaction: tx },
-      );
+      this.storeService.setMetadata('lastProcessedHeight', blockHeight, {
+        transaction: tx,
+      });
       this.storeService.setMetadata('lastProcessedTimestamp', Date.now(), {
         transaction: tx,
       });
@@ -129,7 +127,7 @@ export class IndexerManager {
             this.project.path, //projectId // TODO, define projectId
           );
           poiBlockHash = poiBlock.hash;
-          await this.storeService.setPoi(tx, poiBlock);
+          await this.storeService.setPoi(poiBlock, { transaction: tx });
         }
       }
     } catch (e) {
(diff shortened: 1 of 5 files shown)


packages/node/src/indexer/indexer.manager.ts
```diff
@@ -108,13 +108,12 @@ export class IndexerManager {
...
       }
 
-      await this.metadataRepo.upsert(
-        {
-          key: 'lastProcessedHeight',
-          value: block.block.header.number.toNumber(),
-        },
-        { transaction: tx },
-      );
+      this.storeService.setMetadata('lastProcessedHeight', blockHeight, {
```
ianhe8x, line 111: `await` is missing, and can you create a batch update method for metadata?

packages/node/src/indexer/store.service.ts
```diff
@@ -234,6 +234,17 @@ export class StoreService {
...
 
+  async setMetadataBatch(
+    metadata: Record<string, string | number | boolean>,
+    options?: UpsertOptions<Metadata>,
+  ): Promise<void> {
+    const xs = [];
+    for (const [key, value] of Object.entries(metadata)) {
+      xs.push(this.setMetadata(key, value, options));
+    }
+    await Promise.all(xs);
```
ianhe8x, line 245: Can you make use of bulkCreate() to accelerate the insert further?

packages/node/src/indexer/indexer.manager.ts
```diff
@@ -108,12 +108,12 @@ export class IndexerManager {
...
-      });
-      this.storeService.setMetadata('lastProcessedTimestamp', Date.now(), {
-        transaction: tx,
-      });
+      const metadata: Record<string, string | number | boolean> = {
+        lastProcessedHeight: blockHeight,
+        lastProcessedTimestamp: Date.now(),
+      };
+
+      await this.storeService.setMetadataBatch(metadata, { transaction: tx });
```
ianhe8x, line 116: inline metadata object will look more neat

ianhe8x (approved):

PR #742, opened 2022-01-11
modify query logger config

lachlan1m:
For each graphql request a single json blob will be logged containing any useful information.


packages/query/src/utils/logger.ts
```diff
@@ -36,3 +38,44 @@ export class NestLogger implements LoggerService {
...
+
+export const PinoConfig = {
+  logger: getLogger('express'),
+  serializers: {
+    req(req) {
+      const body = req.raw.body;
+      if ('operationName' in body && body.query) {
+        // Logging IntrospectionQuery payload clutters logs and isn't useful
+        if (body.operationName !== 'IntrospectionQuery') {
+          req.payload = stringify(
```
ianhe8x, line 50: Why it need to be converted into gql rather than just use it straight away?

packages/query/src/utils/logger.ts
```diff
@@ -36,3 +38,44 @@ export class NestLogger implements LoggerService {
...
+      }
+      return req;
+    },
+    res(res) {
+      if (res.headers.stack) {
+        res.stack = res.headers.stack;
+        delete res.headers.stack;
+      }
+      if (res.headers.message) {
+        delete res.headers.message;
```
ianhe8x, line 65: Message in the headers is still not removed. check browser's network inspect

ianhe8x:
```
(node:43238) [PINODEP004] PinoWarning: bindings.serializers is deprecated, use options.serializers option instead
```

packages/query/src/utils/logger.ts
lachlan1m: removed

packages/query/src/graphql/plugins/logGraphqlPlugin.ts
```diff
@@ -0,0 +1,20 @@
...
+import {stringify} from 'flatted';
+
+/* eslint-disable @typescript-eslint/require-await */
+export const LogGraphqlPlugin: ApolloServerPlugin = {
+  async requestDidStart<MyApolloContext>(
+    _requestContext: GraphQLRequestContext<MyApolloContext>
+  ): Promise<GraphQLRequestListener<MyApolloContext>> {
+    return {
+      async didEncounterErrors(requestContext) {
+        requestContext.response.http.headers.set('stack', stringify(requestContext.errors));
```
ianhe8x, line 16: if it's not easy to remove `stack` from headers, I would rather not having it.

ianhe8x (approved):

PR #751, opened 2022-01-14
Spec version filter for cli init

wqsz7xn:
Adds a filter of templates for users chosen specVersion. If there are no templates for that specVersion (which should not be the case) then it falls back to a custom git remote & branch.

### DIFF
--- a/packages/cli/src/commands/init.ts
+++ b/packages/cli/src/commands/init.ts
@@ -1,6 +1,7 @@
 // Copyright 2020-2021 OnFinality Limited authors & contributors
 // SPDX-License-Identifier: Apache-2.0
 
+import assert from 'assert';
 import fs from 'fs';
 import path from 'path';
 import {URL} from 'url';
@@ -66,6 +67,7 @@ export default class Init extends Command {
     force: flags.boolean({char: 'f'}),
     location: flags.string({char: 'l', description: 'local folder to create the project in'}),
     'install-dependencies': flags.boolean({description: 'Install dependencies as well', default: false}),
+    starter: flags.boolean({required: false, default: false, description: 'Select subql-starter as template'}),
     npm: flags.boolean({description: 'Force using NPM instead of yarn, only works with `install-dependencies` flag'}),
     specVersion: flags.string({
       required: false,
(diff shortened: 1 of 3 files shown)


packages/cli/src/commands/init.ts
```diff
@@ -113,65 +113,75 @@ export default class Init extends Command {
...
-      const paddingWidth = candidateTemplates.map(({name}) => name.length).reduce((acc, xs) => Math.max(acc, xs)) + 5;
-
-      const templateDisplays = candidateTemplates.map(
-        ({description, name}) => `${name.padEnd(paddingWidth, ' ')}${chalk.gray(description)}`
-      );
-      templateDisplays.push(`${'Other'.padEnd(paddingWidth, ' ')}${chalk.gray('Enter a custom git endpoint')}`);
+      const networks = templates
+        .map(({network}) => network)
+        .filter((n, i, self) => {
+          return i === self.indexOf(n);
```
ianhe8x, line 126: is this a trick to do unique? I'm not too confident with it, can you just use unique from lodash?

packages/cli/src/commands/init.ts
```diff
@@ -113,65 +113,75 @@ export default class Init extends Command {
...
+        });
+      networks.push('Other');
 
+      // Network
+      inquirer.registerPrompt('autocomplete', require('inquirer-autocomplete-prompt'));
```
ianhe8x, line 131: would be better the setup steps be moved to the beginning of the file, (right after imports).

packages/cli/src/commands/init.ts
wqsz7xn: Sure

wqsz7xn:
Let me put back in the `--starter` flag before merging this. I don't know where it disappeared. My bad

wqsz7xn:
Added back the `--starter` flag so that I don't have to change a bunch of docs

ianhe8x:
The template feature is a replacement for the original --starter. With --starter, it skips template selection. I think we should move on without --starter

ianhe8x (approved):

PR #765, opened 2022-01-20
Add group by in safe mode

lachlan1m:


packages/query/src/graphql/plugins/PgAggregationPlugin.ts
```diff
@@ -0,0 +1,107 @@
...
+
+import type {Plugin} from 'graphile-build';
+import type {PgAttribute, PgType, SQL, PgProc} from 'graphile-build-pg';
+import {makePluginByCombiningPlugins} from 'graphile-utils';
+
+import {argv} from '../../yargs';
+
+const unsafe = argv('unsafe') as boolean;
+
+interface AggregateGroupBySpec {
```
ianhe8x, line 23: you don't need to redefine the interfaces

ianhe8x (changes requested):

packages/query/src/graphql/plugins/PgAggregationPlugin.ts
lachlan1m: Yeah I missed that they were exported.

ianhe8x (approved):

PR #773, opened 2022-01-26
Add support for dynamic datasources

stwiname:
Allows creating new datasources from templates within a handler

### Changes

* Introduces new v0.2.1 project manifest version which adds templates property
* Injects a function called `createDynamicDatasource` into the sandbox to allow creating a new datasource dynamically from a template datasource in the project manifest. These dynamic datasources are stored in the project metadata table

Example project and branch https://github.com/subquery/query-registry-subquery-project/tree/dynamic-ds

### TODO

- [x] Setup codegen to provide better types for creating a dynamic datasource so that `createDynamicDatasource` isn't called directly
- [x] Make sure that a database transaction is used for creating dynamic datasources
- [x] Docs https://github.com/subquery/documentation/pull/48

### Limitations
- If a dynamic DS creates another dynamic DS in the same block it will not get run that block
- With EVM the abis are static and cant easily be updated if the contract changes before its deployed
- Dynamic datasources cant be removed, this could end up slowing performance

### DIFF
--- a/packages/node/src/indexer/dynamic-ds.service.ts
+++ b/packages/node/src/indexer/dynamic-ds.service.ts
@@ -3,16 +3,15 @@
 
 import { Injectable } from '@nestjs/common';
 import {
-  CustomDataSourceV0_2_0Impl,
+    CustomDataSourceV0_2_0Impl,
   isCustomDs,
   isRuntimeDs,
   RuntimeDataSourceV0_2_0Impl,
 } from '@subql/common';
-import { SubqlDatasource, SubqlDatasourceKind } from '@subql/types';
+import { SubqlDatasource } from '@subql/types';
 import { plainToClass } from 'class-transformer';
-import yaml from 'js-yaml';
-import { SubqlProjectDs, SubqueryProject } from '../configure/SubqueryProject';
 import { Transaction } from 'sequelize/types';
+import { SubqlProjectDs, SubqueryProject } from '../configure/SubqueryProject';
 import { getLogger } from '../utils/logger';
 import { DsProcessorService } from './ds-processor.service';
 import { StoreService } from './store.service';
@@ -21,17 +20,10 @@ const logger = getLogger('dynamic-ds');
 
 const METADATA_KEY = 'dynamicDatasources';
 
-function replacer(key: string, value: any): any {
-  if (value instanceof Map) {
-    const obj: Record<string, unknown> = {};
-    for (const key of value.keys()) {
-      obj[key] = value.get(key);
-    }
-
-    return obj;
-  } else {
-    return value;
-  }
+interface DatasourceParams {
+  templateName: string;
+  args?: Record<string, unknown>;
+  startBlock: number;
 }
 
 @Injectable()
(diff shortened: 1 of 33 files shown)


ianhe8x:
The first limitation seems critical, it could happen that people interact with the new created contract within same tx, even in our use case.

packages/node/src/indexer/dynamic-ds.service.ts
```diff
@@ -0,0 +1,130 @@
...
+
+    logger.info(`Creating new datasource from template: "${templateName}"`);
+
+    const ds: SubqlDatasource = { ...template, startBlock: currentBlock };
+    try {
+      if (isCustomDs(ds)) {
+        ds.processor.options = { ...ds.processor.options, ...args };
+        await this.dsProcessorService.validateCustomDs([ds]);
+      } else if (isRuntimeDs(ds)) {
+        // XXX add any modifications to the ds here
```
ianhe8x, line 80: It's not clear to me how runtime ds need to be dynamically created, what parameter requires to be dynamically set. Should we just forbid it for now?

packages/node/src/indexer/dynamic-ds.service.ts
```diff
@@ -0,0 +1,130 @@
...
+  private async saveDynamicDatasource(
+    ds: SubqlDatasource,
+    tx: Transaction,
+  ): Promise<void> {
+    const existing = await this.getDynamicDatasources();
+
+    // Need to convert Map objects to records
+    const dsObj = JSON.parse(JSON.stringify(ds, replacer));
+
+    await this.storeService.setMetadata(
```
ianhe8x, line 124: Can we only persist as minimal data as possible? name, args, startBlock

packages/node/src/indexer/dynamic-ds.service.ts
```diff
@@ -0,0 +1,130 @@
...
+    } catch (e) {
+      logger.error(`Unable to create dynamic datasource.\n ${e.message}`);
+      process.exit(1);
+    }
+
+    await this.saveDynamicDatasource(ds, tx);
+  }
+
+  async getDynamicDatasources(): Promise<SubqlDatasource[]> {
+    const results = await this.storeService.getMetadata(METADATA_KEY);
```
ianhe8x, line 94: I would prefer injecting metaRepo into here. `storeservice` is for preparing the store object.

stwiname:
> The first limitation seems critical, it could happen that people interact with the new created contract within same tx, even in our use case.

Maybe i didnt explain it well. If this happens within a single block: `project ds` -> `dynamic ds 1` -> `dynamic ds 2`. Then `dynamic ds 2` will be created but not run for that block, it will only run subsequent blocks

packages/node/src/indexer/dynamic-ds.service.ts
stwiname: I couldn't think of a use case for this or any specific parameters to set. But I don't see the harm in allowing this.

ianhe8x:
> > The first limitation seems critical, it could happen that people interact with the new created contract within same tx, even in our use case.
> 
> Maybe i didnt explain it well. If this happens within a single block: `project ds` -> `dynamic ds 1` -> `dynamic ds 2`. Then `dynamic ds 2` will be created but not run for that block, it will only run subsequent blocks

It's me miss reading. That's a rare case i guess, at least we don't use it that way. We can log it a separate issue.

ianhe8x (commented):
Overall looks really good, I might spend some time play with it locally today.

jiqiang90 (commented):
We also have to upgrade `polkadot/api` and `moonbeam-types-bundle` version in contract-processor, otherwise bunch of types errors will throw when user project build

stwiname:
@jiqiang90 can you please have another look when you get the chance. I've rebased and had to make some changes to match with the IPFS support.

ianhe8x (approved):

jiqiang90 (approved):

PR #778, opened 2022-01-27
Add authToken to subql publish ipfs-http header

### LINKED ISSUE #776: subql-cli publish support default ipfs server
subql-cli need to put access token in the header for authorizaion, so they can access IPFS server

```
Authorization: Bearer xxxxxxxxxxxxxxx
```
user may store the token in local file system,and cli can read the token from file.

wqsz7xn:
Implements:
- https://github.com/subquery/subql/issues/776

Overview:
- Adds a yaml file for secrets and a flag for a secret authToken
- This yaml doesn't have a spec because it's likely to remain small and I think we would want breaking changes for changes to this yaml. Not using a normal file so that we have the possibility of expanding if we need to later

Closes #776

### DIFF
--- a/packages/cli/src/commands/publish.ts
+++ b/packages/cli/src/commands/publish.ts
@@ -14,7 +14,7 @@ class AuthorizationSpec {
   secret: string;
 }
 
-const ACCESS_TOKEN_PATH = path.resolve(process.env.HOME, 'SUBQL_ACCESS_TOKEN');
+const ACCESS_TOKEN_PATH = path.resolve(process.env.HOME, '.SUBQL_ACCESS_TOKEN');
 
 export default class Publish extends Command {
   static description = 'Upload this SubQuery project to IPFS';
--- a/packages/cli/src/commands/publish.ts
+++ b/packages/cli/src/commands/publish.ts
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import {lstatSync, readFileSync} from 'fs';
+import * as fs from 'fs';
 import path from 'path';
 import {Command, flags} from '@oclif/command';
 import yaml from 'js-yaml';
@@ -14,7 +15,7 @@ class AuthorizationSpec {
   secret: string;
 }
 
-const ACCESS_TOKEN_PATH = path.resolve(process.env.HOME, '.SUBQL_ACCESS_TOKEN');
+const ACCESS_TOKEN_PATH = path.resolve(process.env.HOME, '.subql/SUBQL_ACCESS_TOKEN');
 
 export default class Publish extends Command {
   static description = 'Upload this SubQuery project to IPFS';
@@ -22,7 +23,6 @@ export default class Publish extends Command {
   static flags = {
     location: flags.string({char: 'l', description: 'local folder'}),
     ipfs: flags.string({description: 'IPFS gateway endpoint', required: false}),
-    'access-token': flags.string({description: 'Secret authorization token', required: false}),
   };
 
   async run(): Promise<void> {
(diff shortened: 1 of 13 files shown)


packages/cli/src/controller/publish-controller.ts
```diff
@@ -3,66 +3,102 @@
...
+    ? ipfs.add(content, {pin: true, cidVersion: 0})
+    : Promise.reject(new Error('Ipfs server not provided'));
+  let contentBuffer: Uint8Array;
+  if (determineStringOrFsStream(content)) {
+    contentBuffer = new Blob([fs.readFileSync(content.path)]);
+  } else {
+    contentBuffer = new Blob([content]);
+  }
+  const clusterPromise = cluster.add(contentBuffer, {cidVersion: 0, rawLeaves: false});
+  const results = await Promise.allSettled([clientPromise, clusterPromise]);
```
stwiname, line 88: This might be better suited https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/any

stwiname (commented):
I think users might think the `secret` flag relates to the `ipfs` endpoint they specify when it would actually get sent to our cluster. It could be considered deceptive.

packages/cli/src/commands/publish.ts
```diff
@@ -32,9 +40,19 @@ export default class Publish extends Command {
...
     }
 
-    this.log('Uploading SupQuery project to IPFS');
-    const cid = await uploadToIpfs(flags.ipfs, directory);
+    let authToken: string;
+    if (flags.secret) {
+      authToken = flags.secret;
+    } else {
+      try {
+        authToken = (yaml.load(readFileSync(path.resolve(directory, 'auth.yaml'), 'utf8')) as AuthorizationSpec).secret;
```
ianhe8x, line 48: I don't think this is a project level thing. better store it in user's HOME

packages/cli/src/commands/publish.ts
```diff
@@ -1,18 +1,28 @@
...
+const ACCESS_TOKEN_PATH = path.resolve(process.env.HOME, '.SUBQL_ACCESS_TOKEN');
+
 export default class Publish extends Command {
   static description = 'Upload this SubQuery project to IPFS';
 
   static flags = {
     location: flags.string({char: 'l', description: 'local folder'}),
-    ipfs: flags.string({description: 'IPFS gateway endpoint', default: 'http://localhost:5001/api/v0'}),
+    ipfs: flags.string({description: 'IPFS gateway endpoint', required: false}),
+    'access-token': flags.string({description: 'Secret authorization token', required: false}),
```
ianhe8x, line 25: For env variable & flag, I think just one of them should be enough.

packages/cli/src/commands/publish.ts
```diff
@@ -1,18 +1,28 @@
...
 import {uploadToIpfs} from '../controller/publish-controller';
 import Build from './build';
 
+// Authorization spec is small enough not to have its own specVersion.
+// Additionally when authorization requirements change we would like this to be a breaking change
+class AuthorizationSpec {
+  secret: string;
+}
+
+const ACCESS_TOKEN_PATH = path.resolve(process.env.HOME, '.SUBQL_ACCESS_TOKEN');
```
ianhe8x, line 17: suggest to put it under `.subql` in case more files added

packages/cli/src/controller/publish-controller.ts
```diff
@@ -3,66 +3,114 @@
...
+        Authorization: `Bearer ${authToken}`,
+        'Content-Type': 'multipart/form-data',
+        ...bodyFormData.getHeaders(),
+      },
+      method: 'post',
+      url: IPFS_CLUSTER_ENDPOINT,
+      data: bodyFormData,
+    })
+  ).data as ClusterResponseData;
+  return result.cid['/'];
```
ianhe8x, line 113: - 
```suggestion
  return result.cid?.['/'];
```

ianhe8x (changes requested):

ianhe8x (approved):

PR #781, opened 2022-01-28
Introduce Terra Support

DeveloperInProgress:


packages/types/src/interfaces.ts
```diff
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import {Extrinsic, EventRecord, SignedBlock} from '@polkadot/types/interfaces';
+import {BlockInfo, EventsByType} from '@terra-money/terra.js';
```
ianhe8x, line 5: if you introduce terra sdk to types package, you will need to add dependency in `types/package.json`.
Though this is not the preferable way, since no projects will rely on both polkadotjs and terra.js together.
I suggest a seperate `@subql/types-terra` for terra subquery project only.

packages/types-terra/src/project.ts
```diff
@@ -0,0 +1,144 @@
+// Copyright 2020-2021 OnFinality Limited authors & contributors
+// SPDX-License-Identifier: Apache-2.0
+
+import {BlockInfo, EventsByType, LCDClient} from '@terra-money/terra.js';
```
stwiname, line 4: If we can limit to just type imports that would be ideal. Same applies to other files in terra-types

```suggestion
import types{BlockInfo, EventsByType, LCDClient} from '@terra-money/terra.js';
```

packages/node-terra/src/indexer/apiterra.service.ts
```diff
@@ -0,0 +1,55 @@
...
+    private eventEmitter: EventEmitter2,
+  ) {}
+
+  // eslint-disable-next-line @typescript-eslint/require-await
+  async init(): Promise<ApiTerraService> {
+    const { network } = this.project;
+    this.clientConfig = {
+      URL: network.endpoint,
+      chainID: network.chainId,
+    };
```
stwiname, line 28: If the URL doesn't match the chainId, what happens?

stwiname (commented):
Well done. 
Some other comments:
- Does POI work, it looks like there is some commented out code?
- Is `node-terra/src/x-vm2` still necessary?
- Whats the reason for having `types-terra`?  I think we want it to be under `types` as it will still be required by `common`
- It would be good to see some tests, specifically for the manifest changes and other code that has been changed for terra support

packages/common/src/project/types.ts
```diff
@@ -2,12 +2,13 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import {SubqlDatasource} from '@subql/types';
+import {SubqlTerraDatasource} from '@subql/types-terra';
```
ianhe8x, line 5: `@subql/common` is used by `@subql/node`, therefore by introducing `@subql/types-terra` it contaminate dependencies of  `@subql/node`

packages/common/src/project/versioned/v0_3_0/model.ts
```diff
@@ -0,0 +1,130 @@
...
+    },
+    keepDiscriminatorProperty: true,
+  })
+  dataSources: (RuntimeDataSourceV0_3_0 | CustomDatasourceV0_3_0)[];
+  @ValidateNested()
+  @Type(() => ProjectNetworkDeploymentV0_3_0)
+  network: ProjectNetworkDeploymentV0_3_0;
+}
+
+export class ProjectManifestV0_3_0Impl extends ProjectManifestBaseImpl implements ProjectManifestV0_3_0 {
```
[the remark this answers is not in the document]
DeveloperInProgress: @stwiname I was aiming for it, given that v0.2.0 already supports substrate. It would be neat to have separate manifest designs for terra and substrate chains right? It makes processing and validation look simple.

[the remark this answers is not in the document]
jamesbayly: Quite the opposite, we're aiming for one manifest design that works with all networks

packages/common/src/query/types.ts
```diff
@@ -12,3 +12,13 @@ export type MetaData = {
   indexerNodeVersion: string;
   queryNodeVersion: string;
 };
+
+export type TerraMetaData = {
```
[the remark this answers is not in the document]
DeveloperInProgress: Should I make `specName` and `genesisHash` optional?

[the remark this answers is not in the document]
stwiname: We need one of `genesisHash` or `chainId` to be provided. `genesisHash` should be an alias to `chainId` as thats more generic

stwiname:
@DeveloperInProgress. In order to avoid having multiple chain sdks in chain specific packages we would like to have packages as follows: 
* `@subql/common`: nothing chain specific
* `@subql/common-substrate`: specific classes for substrate
* `@subql/common-terra`: specific classes for terra

The CLI package can use all chain versions. `@subql/node` should use `@subql/common-substrate`, `@subql/node-terra` should use `@subql/common-terra` and `@subql/query` should use `@subql/common`.

DeveloperInProgress:
@stwiname . In which of these common packages should v0.3.0 manifest come under?  It is going to cover both substrate and terra

stwiname:
> 
> @stwiname . In which of these common packages should v0.3.0 manifest come under?  It is going to cover both substrate and terra
> 

The base class should live in common. That would include things common to any chain. And there would be the specific chain versions in each of the named packages

I would suggest reading again. There has been some changes in this area that might be harder to bring over if work continues.

packages/common-substrate/src/project/types.ts
```diff
@@ -0,0 +1,19 @@
...
+export interface IProjectManifest {
+  specVersion: string;
+  description: string;
+  repository: string;
+  dataSources: SubqlDatasource[];
+  toDeployment(): string;
+  validate(): void;
+}
+
+export interface ProjectNetworkConfig {
```
stwiname, line 15: These could live in `@subql/common`. Remember  `genesisHash` should just be an alias

packages/common-substrate/src/project/versioned/v0_3_0/model.ts
```diff
@@ -0,0 +1,139 @@
...
+import yaml from 'js-yaml';
+import {CustomDataSourceBase, Mapping, RuntimeDataSourceBase} from '../../models';
+import {
+  CustomDatasourceV0_3_0,
+  SubstrateProjectManifestV0_3_0,
+  RuntimeDataSourceV0_3_0,
+  SubqlMappingV0_3_0,
+} from './types';
+
+export class FileTypeV0_3_0 {
```
stwiname, line 23: There's no need to define a new class if the previous version is the same. We can reuse a lot of code. 

If you rebase you will see how v0.2.1 was done.

packages/common/package.json
```diff
@@ -15,6 +15,7 @@
   "dependencies": {
     "@polkadot/util": "^7.9.2",
     "@subql/types": "workspace:*",
+    "@subql/types-terra": "workspace:*",
```
stwiname, line 18: This shouldn't be a dependency

stwiname (commented):
Nice work. We're on the right track, I think it would be good to try and reuse as much code as possible, by moving nearly identical code out of common-substrate and common-terra and into the common package.

packages/common-substrate/src/project/versioned/ProjectManifestVersioned.ts
```diff
@@ -1,32 +1,44 @@
...
-export class ProjectManifestVersioned implements IProjectManifest {
+export function manifestIsV0_2_1(manifest: ISubstrateProjectManifest): manifest is ProjectManifestV0_2_1Impl {
+  return manifest.specVersion === '0.2.1';
+}
+
+export function manifestIsV0_3_0(manifest: ISubstrateProjectManifest): manifest is ProjectManifestV0_3_0Impl {
+  return manifest.specVersion === '0.3.0';
+}
+
+export class ProjectManifestVersioned implements ISubstrateProjectManifest {
```
stwiname, line 41: I think we can have a base class for this, the terra implementation is almost identical

packages/common-substrate/src/project/versioned/v0_2_0/model.ts
```diff
@@ -43,22 +43,40 @@ export class ProjectMappingV0_2_0 extends Mapping {
   file: string;
 }
 
+function validateObject(object: any, errorMessage = 'failed to validate object.'): void {
```
stwiname, line 46: This can be moved to `@subql/common`

packages/common-substrate/src/project/versioned/v0_2_0/model.ts
```diff
@@ -106,7 +124,7 @@ export class ProjectManifestV0_2_0Impl extends ProjectManifestBaseImpl implement
     keepDiscriminatorProperty: true,
   })
   dataSources: (RuntimeDataSourceV0_2_0 | CustomDatasourceV0_2_0)[];
-  private _deployment: DeploymentV0_2_0;
+  protected _deployment: DeploymentV0_2_0;
 
   toDeployment(): string {
```
stwiname, line 129: This function and `validate` can be moved to https://github.com/DeveloperInProgress/subql/blob/main/packages/common/src/project/versioned/base.ts#L6 and `ProjectManifestBaseImpl` can be made into an abstract class

stwiname (commented):
This is looking good. Its very hard to see the changes now though, a lot of recent changes from main are showing up in the diff.

Could you please update the CI pipelines for the new packages (in `.github` folder)

packages/common-substrate/src/project/versioned/v0_2_0/model.ts
DeveloperInProgress: shall I move it to `@subql/common/project/versioned/v0_2_0/model.ts`  or otherwise move it to an utility file such as `@subql/common/project/utils.ts`?

packages/common-substrate/src/project/versioned/ProjectManifestVersioned.ts
DeveloperInProgress: most of the identical functions uses `_impl` which is of type `ProjectManifestImpls` . How shall I create this type in base class?
https://github.com/DeveloperInProgress/subql/blob/472189516e3b8ac301f36f8e4cc10e7471677ef8/packages/common-substrate/src/project/versioned/ProjectManifestVersioned.ts#L23

stwiname: I don't think so, you can use generics and either have it extend the base class or the interface

stwiname:
@DeveloperInProgress 

Can we please remove the Api/LCD client being injected into the sandbox. There doesn't appear to be any way to limit queries to a point in time (the current block being indexed)

ianhe8x (approved):

PR #788, opened 2022-02-04
Add support for historical RPC methods that use BlockNumber

stwiname:
Once https://github.com/polkadot-js/api/pull/4540 is released all the rpc methods that have `isHistoric` params can be `BlockNumber` type. This adds a check for the type and inserts the current block number (rather than block hash)

### DIFF
--- a/packages/node/src/indexer/api.service.test.ts
+++ b/packages/node/src/indexer/api.service.test.ts
@@ -324,7 +324,8 @@ describe('ApiService', () => {
     await expect(patchedApi.query.system.events()).resolves.toHaveLength(2);
   });
 
-  it('can correctly call eth rpc methods', async () => {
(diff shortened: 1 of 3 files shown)


ianhe8x:
@jiqiang90 shouldn't these eth methods historical? then we can support them without any changes

stwiname:
> @jiqiang90 shouldn't these eth methods historical? then we can support them without any changes

The `isHistoric` prop doesn't seem to be on the params. And the methods take the block number rather than block hash

ianhe8x:
this is a chain specific thing, I think it's better to be handled in user project side. Apparently acala don't use that rpc.

stwiname:
I've made changes to polkadot.js https://github.com/polkadot-js/api/pull/4539

I will leave this PR open because we will still need to match the types (BlockHash or BlockNumber)

ianhe8x (approved):

PR #796, opened 2022-02-10
Add custom resolvers for dictionary projects

lachlan1m:


packages/query/src/graphql/plugins/PgDictionaryPlugin.ts
```diff
@@ -0,0 +1,147 @@
...
+      }
+    }
+  }
+};
+
+const PgDictionaryPlugin = makeExtendSchemaPlugin((build, options) => {
+  const [schemaName] = options.pgSchemas;
+  const {pgSql: sql} = build;
+
+  const eventsTableExists = build.pgIntrospectionResultsByKind.class.find(
```
lachlan1m, line 57: [ ] check on table columns as well

packages/query/src/graphql/plugins/PgDictionaryPlugin.ts
```diff
@@ -0,0 +1,147 @@
...
+import {makeExtendSchemaPlugin, gql} from 'graphile-utils';
+
+const SORT_ARGS = {
+  lessThan: '<',
+  lessThanOrEqualTo: '<=',
+  greaterThan: '>',
+  greaterThanOrEqualTo: '>=',
+};
+
+const generateArgs = (queryBuilder: QueryBuilder, args, sql) => {
```
lachlan1m, line 14: - [ ] I can make this cleaner

ianhe8x: You can use `builder.hook("GraphQLObjectType:fields:field:args",` like in here https://github.com/graphile-contrib/postgraphile-plugin-connection-filter/blob/fda1bc1e4474a3a64be10f3228c81b88560f6799/src/PgConnectionArgFilterPlugin.ts#L27

packages/query/src/graphql/plugins/PgDictionaryPlugin.ts
```diff
@@ -0,0 +1,71 @@
...
+        distinctExtrinics(on: String!): ExtrinsicsConnection!
+        distinctSpecVersions(on: String!): SpecVersionsConnection!
+      }
+    `,
+    resolvers: {
+      Query: {
+        distinctEvents: async (_parentObject, args, _context, info): Promise<any> => {
+          if (eventsTableExists) {
+            return info.graphile.selectGraphQLResultFromTable(
+              sql.fragment`(select distinct on (${sql.identifier(args.on)}) * from ${sql.identifier(
```
lachlan1m, line 36: sql needs to be updated

PR #800, opened 2022-02-11
Support node load and cli publish from a manifest file

### LINKED ISSUE #794: subql-node -f point to a project.yaml
in order to support multiple network

jiqiang90:
Closes #794

### DIFF
--- a/packages/cli/src/commands/publish.ts
+++ b/packages/cli/src/commands/publish.ts
@@ -25,6 +25,8 @@ export default class Publish extends Command {
     if (!lstatSync(directory).isDirectory()) {
       this.error('Argument `location` is not a valid directory');
     }
+    //TODO, now publish will only read from project.yaml, we might allow it point to different file
+    const manifestPath = path.resolve(directory, 'project.yaml');
 
     // Ensure that the project is built
     try {
@@ -49,7 +51,7 @@ export default class Publish extends Command {
     }
 
     this.log('Uploading SupQuery project to IPFS');
-    const cid = await uploadToIpfs(directory, authToken.trim(), flags.ipfs).catch((e) => this.error(e));
+    const cid = await uploadToIpfs(directory, authToken.trim(), manifestPath, flags.ipfs).catch((e) => this.error(e));
     this.log(`SubQuery Project uploaded to IPFS: ${cid}`);
   }
 }
--- a/packages/cli/src/commands/publish.ts
+++ b/packages/cli/src/commands/publish.ts
@@ -1,7 +1,7 @@
 // Copyright 2020-2022 OnFinality Limited authors & contributors
 // SPDX-License-Identifier: Apache-2.0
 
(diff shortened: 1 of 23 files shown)


ianhe8x:
Can you add tests for this feature?

jiqiang90:
Need to add SUBQL_ACCESS_TOKEN in $secret

packages/cli/src/commands/publish.ts
```diff
@@ -46,9 +46,7 @@ export default class Publish extends Command {
     }
```
ianhe8x, line 46: Seems need to throw here.

packages/common/src/project/readers/reader.ts
```diff
@@ -42,10 +39,10 @@ export class ReaderFactory {
       return new IPFSReader(locationWithoutSchema, options.ipfs);
```
ianhe8x, line 39: logic can be improved here. if `ipfs://` is provided but then is tested to be not valid, shouldn't we throw error?

packages/node/src/configure/configure.module.ts
```diff
@@ -17,6 +17,8 @@ const YargsNameMapping = {
   local: 'localMode',
 };
 
+const IPFS_REGEX = /ipfs:\/\//i;
```
ianhe8x, line 20:
```suggestion
const IPFS_REGEX = /^ipfs:\/\//i;
```

packages/node/src/configure/configure.module.ts
```diff
@@ -36,11 +38,13 @@ function yargsToIConfig(yargs: Args): Partial<IConfig> {
...
 
 function defaultSubqueryName(config: Partial<IConfig>): MinConfig {
+  const ipfsMatch = config.subquery.match(IPFS_REGEX);
   return {
     ...config,
     subqueryName:
-      config.subqueryName ??
-      last(path.resolve(config.subquery).split(path.sep)),
+      config.subqueryName ?? ipfsMatch
+        ? config.subquery.replace('ipfs://', '')
```
ianhe8x, line 46:
```suggestion
        ? config.subquery.replace(IPFS_REGEX, '')
```

packages/node/src/configure/configure.module.ts
```diff
@@ -112,12 +116,17 @@ export class ConfigureModule {
...
-          config.subquery,
-        );
+    let projectPath: string;
+
+    const ipfsMatch = argv.subquery.match(IPFS_REGEX);
+    if (ipfsMatch) {
+      projectPath = argv.subquery;
+    } else {
+      projectPath = path.resolve(
+        config.configDir && !argv.subquery ? config.configDir : '.',
```
ianhe8x, line 126: The origin implementation seems too complicated. argv.subquery should be mandatory then here can be simplified to `projectPath = argv.subquery`, `path.resolve()` can be done in localreader.

ianhe8x (approved):

PR #809, opened 2022-02-17
wip

roduquen:


packages/node/src/indexer/api.wrapper.ts
```diff
@@ -0,0 +1,217 @@
...
+import { ApiPromise } from '@polkadot/api';
+import { ApiInterfaceEvents, ApiOptions } from '@polkadot/api/types';
+import {
+  BlockHash,
+  SignedBlock,
+  RuntimeVersion,
+  Header,
+} from '@polkadot/types/interfaces';
+
+export class ApiWrapper {
```
stwiname, line 13: I think it would be good to extract an interface from this and have an implementation for Polkadot and for Algorand.

PR #812, opened 2022-02-20
Rename Moonbeam datasource processor

stwiname:
The datasource processor seems to apply to Frontier so its better to use Frontier as the name.

Also re-exports Frontier as Moonbeam for backwards compatibility.


ianhe8x:
Since the ds kinds are changed as well, it will break legacy projects. Re-exporting isn't enough. I think it need to be either we add the original moonbeam kind as an alias, or we just make it clear a breaking change (and bump version).

ianhe8x:
Regards naming, i feel we can make it more clear by changing `Frontier` to `FrontierEVM`

jamesbayly:
We should also update the sample project https://github.com/subquery/tutorials-moonriver-evm-starter

PR #822, opened 2022-02-24
Robin/algorand

roduquen:


packages/node/src/indexer/api.wrapper.ts
```diff
@@ -0,0 +1,196 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+import { ApiPromise } from '@polkadot/api';
+import { ApiInterfaceEvents, ApiOptions } from '@polkadot/api/types';
+import { BlockHash, RuntimeVersion } from '@polkadot/types/interfaces';
+import algosdk from 'algosdk';
+import { BlockContent } from './types';
+import { AlgorandBlock, AlgorandApi } from './typesAlgo';
+
+export class ApiWrapper {
```
stwiname, line 11: Hi @roduquen, I'm not sure if you saw the comment on the last PR.

It would be good to have an IApiWrapper interface and split the implementation into 2, for each chain

tdelabro: I think he saw it because he have been talking about this exact implementation. But it was kind of easier to start this way. It will be our next move.

Issue #823, opened 2022-02-24
"Unable to properly filter input" on blocks at a later date.


kenjicncr:
I'm getting an `Unable to properly filter input` after a certain number of blocks. 

The filter input is fine about 5-6 days ago. But anything after that throws this error.

Here's the link to the repo

https://github.com/kenjicncr/blockplace-subgraph

kenjicncr:
Did a little digging and the tx seems to be returning undefined. Here's what the input data looks like. Looks like i'm getting legacy transaction input?

```
[...]

PR #825, opened 2022-02-25
Fix moonbeam filters post EIP1559

stwiname:
Addresses https://github.com/subquery/subql/issues/823

### DIFF
--- a/packages/contract-processors/src/moonbeam.ts
+++ b/packages/contract-processors/src/moonbeam.ts
@@ -389,17 +389,23 @@ const CallProcessor: SecondLayerHandlerProcessor<
         return false;
       }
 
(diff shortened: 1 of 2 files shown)


packages/contract-processors/src/moonbeam.ts
```diff
@@ -389,17 +389,23 @@ const CallProcessor: SecondLayerHandlerProcessor<
         return false;
       }
 
-      const [tx] = input.extrinsic.method.args as [EthTransaction];
+      const [tx] = input.extrinsic.method.args as [TransactionV2 | EthTransaction];
```
stwiname, line 392: Id love to find a way to not cast here

ianhe8x (approved):

PR #844, opened 2022-03-10
Fix enum sort order

jiqiang90:


packages/node/src/indexer/store.service.ts
```diff
@@ -89,12 +89,14 @@ export class StoreService {
       // We shouldn't set the typename to e.name because it could potentially create SQL injection,
       // using a replacement at the type name location doesn't work.
       const enumTypeName = `${schema}_enum_${this.enumNameToHash(e.name)}`;
+      console.log(`enumTypeName : ${enumTypeName}`);
```
ianhe8x, line 92: Don't forget to clean up

ianhe8x (approved):
lgtm

PR #846, opened 2022-03-12
Add GraphQL subscription support

### LINKED ISSUE #237: Support graphql's subscription

spitko:
Resolves #237 

## Node package

Added trigger function send_notification() that sends notification using pg_notify to topic [schema].[name]

Notification payload (JSON string) consists of:
1. id (Entity ID),
2. mutation_type (INSERT/UPDATE/DELETE)
3. _entity (JSON-formatted entity)

Note that pg_notify has a payload size limit of 8000 bytes

## Query package

- Used pg-pubsub plugin to handle listening to Postgres notifications 
- Created a custom graphile plugin to add subscription fields to schema
  - Entity field is currently typed as JSON
- Used subscriptions-transport-ws for websocket server
  - this should be replaced with graphql-ws once [support is added to playground](https://github.com/graphql/graphql-playground/issues/1291)

### DIFF
--- a/packages/node/src/utils/sync-helper.ts
+++ b/packages/node/src/utils/sync-helper.ts
@@ -37,3 +37,34 @@ export function createUniqueIndexQuery(
     field,
   )}' on '${schema}.${table}' (${underscored(field)})`;
 }
+
+export const createSendNotificationTriggerFunction = `
+CREATE OR REPLACE FUNCTION send_notification()
+    RETURNS trigger AS $$
+DECLARE
+    row RECORD;
+BEGIN
+    IF (TG_OP = 'DELETE') THEN
+      row = OLD;
+    ELSE
+      row = NEW;
+    END IF;
+    PERFORM pg_notify(
+      CONCAT(TG_TABLE_SCHEMA, '.', TG_TABLE_NAME),
+      json_build_object(
+        'id', row.id,
+        'mutation_type', TG_OP,
+        '_entity', row)::text);
+    RETURN NULL;
+END;
+$$ LANGUAGE plpgsql;`;
+
+export function createNotifyTrigger(schema: string, table: string): string {
+  return `
+DROP TRIGGER IF EXISTS ${schema}_${table}_notify_trigger
+    ON ${schema}.${table};
+CREATE TRIGGER ${schema}_${table}_notify_trigger
+    AFTER INSERT OR UPDATE OR DELETE
+    ON ${schema}.${table}
+    FOR EACH ROW EXECUTE FUNCTION send_notification();`;
+}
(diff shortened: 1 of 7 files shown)


ianhe8x:
Thanks for contribution, we will look at it as soon as we can

packages/node/src/utils/sync-helper.ts
```diff
@@ -37,3 +37,34 @@ export function createUniqueIndexQuery(
...
+        'mutation_type', TG_OP,
+        '_entity', row)::text);
+    RETURN NULL;
+END;
+$$ LANGUAGE plpgsql;`;
+
+export function createNotifyTrigger(schema: string, table: string): string {
+  return `
+DROP TRIGGER IF EXISTS ${schema}_${table}_notify_trigger
+    ON ${schema}.${table};
```
jiqiang90, line 65: suggest wrap `${schema}` with double quote "" 

if we have project name in camelCase, like "dragonFly", when init the project your will get error like : 

```
<store> ERROR Having a problem when syncing schema SequelizeDatabaseError: schema "dragonfly" does not exist

// The query

DROP TRIGGER IF EXISTS dragonFly_starter_entities_notify_trigger
    ON dragonFly.starter_entities;
CREATE TRIGGER dragonFly_starter_entities_notify_trigger
    AFTER INSERT OR UPDATE OR DELETE
    ON dragonFly.starter_entities
    FOR EACH ROW EXECUTE FUNCTION send_notification(); 

```

We can fix it with double quote

```
DROP TRIGGER IF EXISTS dragonFly_starter_entities_notify_trigger
    ON "dragonFly".starter_entities;

```

packages/node/src/utils/sync-helper.ts
```diff
@@ -37,3 +37,34 @@ export function createUniqueIndexQuery(
...
+CREATE TRIGGER ${schema}_${table}_notify_trigger
+    AFTER INSERT OR UPDATE OR DELETE
+    ON ${schema}.${table}
```
jiqiang90, line 68: add double quote here

jiqiang90: Hi @spitko,  also can we add this feature under `--unsafe` mode ? Thank you 

Other LGTM

spitko:
Thanks for review @jiqiang90, just pushed fixes

spitko:
Pushed a fix to set entity as null in notification payload if payload size is over the 8000 byte limit

spitko:
Pushed a fix to set entity as null in notification payload if payload size is over the 8000 byte limit

jiqiang90:
Thanks @spitko, one more thing, can we add filter for id and mutation_type?

spitko:
> Thanks @spitko, one more thing, can we add filter for `id` and `mutation_type`?

Done

jamesbayly:
@spitko  I strongly suggest you also take a look at the [Historical Query Support](https://gitcoin.co/issue/subquery/grants/7/100028529)

ianhe8x (approved):

PR #848, opened 2022-03-14
Add contract handling for terra

DeveloperInProgress:


packages/node-terra/src/utils/terra-helper.ts
```diff
@@ -61,43 +108,102 @@ export async function fetchTerraBlocksArray(
...
   const txInfos = await getTxInfobyHashes(api, txHashes);
-  const txLogs = txInfos.map((txInfo) => txInfo.logs);
-  const txLogsFlat = ([] as TxLog[]).concat(...txLogs);
-  const events = txLogsFlat.map((txLog) => txLog.eventsByType);
-  return events;
+  const callInfos: TerraCall[] = [];
+  for (const txInfo of txInfos) {
+    for (const msg of txInfo.tx.body.messages) {
+      if (msg.toData()['@type'] === '/terra.wasm.v1beta1.MsgExecuteContract') {
+        const data = msg.toData() as MsgExecuteContract.Data;
```
stwiname, line 163: Should we be using https://github.com/terra-money/terra.js/blob/9e5f553/src/core/Msg.ts#L238 here.

packages/types-terra/src/interfaces.ts
```diff
@@ -26,5 +26,23 @@ export interface TerraBlock {
...
+export interface TerraCall {
+  from: string;
+  to: string;
+  fee: Fee.Data;
+  data: MsgExecuteContract.Data;
+  hash: string;
+  blockNumber: number;
+  blockHash: string;
+  timestamp: string;
+  signatures: string[];
```
stwiname, line 47: I think we can improve these types to be more like Substrate

```ts

interface TerraBlock {
  block: BlockInfo;
  txs: TxInfo[];
}

interface TerraCall {
  idx: number;
  block: TerraBlock;
  tx: TxInfo;
}

interface TerraEvent {
  idx: number; // To avoid doing https://github.com/DeveloperInProgress/terra-contract-starter/blob/main/src/mappings/mappingHandlers.ts#L17
  block: TerraBlock;
  call: TerraCall;
  event: TxLog; // Not sure this is the type we want
}

```

packages/node-terra/src/utils/terra-helper.ts
```diff
@@ -61,43 +108,102 @@ export async function fetchTerraBlocksArray(
...
   }
   const txInfos = await getTxInfobyHashes(api, txHashes);
-  const txLogs = txInfos.map((txInfo) => txInfo.logs);
-  const txLogsFlat = ([] as TxLog[]).concat(...txLogs);
-  const events = txLogsFlat.map((txLog) => txLog.eventsByType);
-  return events;
+  const callInfos: TerraCall[] = [];
+  for (const txInfo of txInfos) {
+    for (const msg of txInfo.tx.body.messages) {
+      if (msg.toData()['@type'] === '/terra.wasm.v1beta1.MsgExecuteContract') {
```
stwiname, line 162: What about other types? Can we support them?

Im trying to understand this better. How does a transaction have multiple messages?

From what I can tell we don't want to filter to be a specific type here, we should allow all types for calls. But we can only support filters for `/terra.wasm.v1beta1.MsgExecuteContract` for now

packages/cli/src/controller/codegen-controller.ts
```diff
@@ -190,16 +191,18 @@ export async function codegen(projectPath: string): Promise<void> {
   await prepareDirPath(modelDir, true);
   await prepareDirPath(interfacesPath, false);
 
+  const rawManifest = loadFromJsonOrYaml(path.join(projectPath, 'project.yaml'));
   let manifest;
-  try {
+  if ((rawManifest as any).dataSources[0].kind.split('/')[0] === 'substrate') {
```
ianhe8x, line 196: This looks very intrusive. Is there any abstraction can be done for it?

packages/node-terra/src/indexer/fetchterra.service.test.ts
```diff
@@ -83,7 +83,7 @@ describe('FetchService', () => {
     const loopPromise = fetchService.startLoop(6684553);
     // eslint-disable-next-line @typescript-eslint/require-await
     fetchService.register(async (content) => {
-      if (Number(content.block.block.header.height) === 6684563) {
+      if (Number(content.block.block.block.header.height) === 6684563) {
```
ianhe8x, line 86: Is this a mistake? Same thing in `fetchterra.service.ts`, make me think it is due to a global replace.

packages/node-terra/src/utils/terra-helper.ts
DeveloperInProgress: From what I understand all contract calls happens through MsgExecuteContract type messages.

Multiple messages seemed new to me too. But many of the transactions in terra had multiple messages and multiple logs corresponding to each to those messages

packages/types-terra/src/interfaces.ts
DeveloperInProgress: we could msg_index from logs as idx for TerraEvent. But, what shall we assign for idx in TerraCall?

DeveloperInProgress: Here is what I suggest:

```
export interface TerraBlock {
  block: BlockInfo;
  txs: TxInfo[];
}

export interface TerraEvent {
  event: EventsByType;
  block: TerraBlock;
  msgIndex: number;
  txInfo: TxInfo;
  log: TxLog;
}

export interface TerraCall {
  idx: number;
  data: MsgExecuteContract.Data;
  block: TerraBlock;
  tx: TxInfo;
}
```

In TerraEvent, msgIndex could be used along transaction hash to create an unique primary key. Also, provide the event data as EventsByType for easy access and also the corresponding TxLog for those who would like to deal with the raw log.

In TerraCall, should idx be a random generated number?

packages/node-terra/src/indexer/fetchterra.service.test.ts
DeveloperInProgress: No, at first TerraBlockContent was something like:

```
{
   block: BlockInfo
   ...
}
```

Now It's changed to:
```
{
   block: TerraBlock
   ...
}
```

Where TerraBlock is:

```
{
   block: BlockInfo,
   txs: TxInfo[]
}
```

packages/cli/src/controller/codegen-controller.ts
```diff
@@ -190,16 +191,19 @@ export async function codegen(projectPath: string): Promise<void> {
   await prepareDirPath(modelDir, true);
   await prepareDirPath(interfacesPath, false);
 
+  const rawManifest = loadFromJsonOrYaml(path.join(projectPath, 'project.yaml'));
   let manifest;
-  try {
+  const datasourceChain = (rawManifest as any).dataSources[0].kind.split('/')[0];
```
stwiname, line 196: Do we need these changes now that we dont have a contract/datasource processor?

packages/node-terra/src/utils/terra-helper.ts
```diff
@@ -30,29 +35,70 @@ export function filterEvents(
...
+async function getBlockByHeight(api: LCDClient, height: number) {
+  return api.tendermint.blockInfo(height).catch((e) => {
     logger.error(`failed to fetch Block ${height}`);
     throw e;
   });
 }
 
 export async function fetchTerraBlocksArray(
-  api: TerraClient,
+  api: LCDClient,
```
stwiname, line 101: Like in my previous comment, all these need to be changed back to `TerraClient`. There are various occurrences and `LCDClient` shouldn't be called directly

packages/types-terra/src/interfaces.ts
stwiname: By the looks of things we want another type of handler for `Messages`. Because `Blocks` have `Transactions` which have `Messages` which  have `Events`. We're currently skipping over the extra layer of depth.

We should make the following changes:

- Rename `Call` to `Transaction` to be consistent with Terra naming conventions
- Introduce `Message` handler type.
    - Filter can have a `type` which should match to these [types](https://github.com/terra-money/terra.js/blob/9e5f553/src/core/Msg.ts#L239)
    - Filter can also match any key value on the [Amino type](https://github.com/terra-money/terra.js/blob/9e5f553de3ff3e975eaaf91b1f06e45658b1a5e0/src/core/wasm/msgs/MsgExecuteContract.ts#L113) e.g. `- sender: terra1j66jatn3k50hjtg2xemnjm8s7y8dws9xqa5y8w`. We can keep validation to simple object validation for now
- `Event` filters can keep the `type` filter as well as adding Message filters. 
   ```yaml
    - type: transfer
    - messageFilter:
        - type: '/terra.wasm.v1beta1.MsgExecuteContract'
        - contract: terra1627ldjvxatt54ydd3ns6xaxtd68a2vtyu7kakj
   ```

Let me know if this doesn't make sense and I can explain further.

> In TerraCall, should idx be a random generated number?

We don't want to randomly generate numbers

> Also, provide the event data as EventsByType

I don't think this is very helpful except for in a block handler.

jamesbayly: Agree, the idx should be the ordered index of the event/call on the block if possible

stwiname:
Hi @DeveloperInProgress. I've cleaned up some of the code and I think this is very close to being ready.

The only thing missing is to filter the contract method (the keys in the `execute_msg`). I think we need to add a special case for this.

We can update the message filter params to have conditional validation like this:

https://github.com/typestack/class-validator#conditional-validation

```ts
export class TerraMessageFilter implements SubqlTerraMessageFilter {
  @IsString()
  type: string;
  @IsOptional()
  @IsObject()
  values?: {[key: string]: string};

  // Just an example, not tested
  @ValidateIf(o => o.type === '/terra.wasm.v1beta1.MsgExecuteContract')
  @IsOptional()
  @IsString()
  contractCall?: string;
}
```

And with this tx https://finder.terra.money/mainnet/tx/d8816c05fb5a87018d7a0aa6c4b3cfaadda623c668e88a382a7bb17987f589c8

the filter would be 

```yaml
     - handler: handleMessage
          kind: terra/MessageHandler
          filter:
            type: '/terra.wasm.v1beta1.MsgExecuteContract'
            contractCall: 'provide_liquidity'
            values: 
              contract: 'terra1j66jatn3k50hjtg2xemnjm8s7y8dws9xqa5y8w'
```

stwiname (approved):
Im happy with this as is now without dictionary support, we can do that in another PR. 

@ianhe8x did you want to give another review?

PR #849, opened 2022-03-14
Update polkadot to 7.12.1

jiqiang90:


ianhe8x:
@jiqiang90 Do you think we should skip the moonriver related tests or switch to other network so tests are not failing.

Issue #780, opened 2022-01-27
Handle all data in time sequence order rather than handler definition order


ianhe8x:
For example, the following case
```
# block x
call1
  -> event1
  -> event2
call2
  -> event3
  -> event4
# handlers
- callHandler
- eventHandler
```

Currently we process them in handler order, meaning `call1` -> `call2` -> `event1` -> ... -> `event4`

After we introduce dynamic datasource, things become even more complicated.
```
# block x
call1
  -> event1 (contract creation)
      -> event 5 (from the new created contract)
  -> event2 
call2
  -> event3
# handlers
- callHandler
- eventHandler
```
The order we implemented is `call1` -> `call2` -> `event1` -> `event2` -> `event3` -> `event5`

If we use time sequence order, it becomes `call1` -> `event1`  -> `event5` -> `event2` -> `call2` ->  `event3`, which seems a more reasonable order.

PR #853, opened 2022-03-19
Process block content in time secquence rather than ds/handler order

stwiname:
⚠️ This could break existing projects ⚠️ 

Addresses https://github.com/subquery/subql/issues/780

There is also a slight change with dynamic datasources. Any new datasources created during a block will get run for the current data and future data in the block.

e.g extrinsic 0 event 1 creates a dynamic datasource, then it can be run from extrinsic 0 event 1 and any extrinsics/events after


packages/node/src/indexer/indexer.manager.ts
```diff
@@ -456,84 +445,171 @@ export class IndexerManager {
...
-          );
-          for (const e of filteredEvents) {
-            await vm.securedExec(handler.handler, [e]);
-          }
-          break;
-        }
-        default:
+    await this.indexBlockContent(block, dataSources, getVM);
+
+    for (const extrinsic of extrinsics) {
```
ianhe8x, line 455: Initialise phase events should be handled before any extrinsics and ApplyExtrinsic phase events.
Finalise phase events and other remaining events are handled last.

ianhe8x (approved):

Issue #415, opened 2021-08-04
Support query by block number (historical state)


jiqiang90:
**Problem:** 

Currently, subquery only supports storing the final state of index data. For example, we indexed a network called `ABC` and synchronized it to its latest height of 5000,  then the data in the database is reflected the result at this block height. 
But we don't know the state of the data when the altitude of 4000 is queried. Therefore, we need a way to record state changes with block heights. 

**Approach**

- Add an extra field to the table to record the range of blocks, eg [35, 38) means from block 35 to 37 it remain in this status. And when this record need to be "update", we create a new record with block range [38, ) 
- Insert a new record rather than update means we need to resolve the conflict of id, so rather let user created ID to primary key, we will generate another field to be PK. But we will face challenge for update entity relationships.  
- Ensure that the solution is scalable to current large projects (e.g. SubVis)

ianhe8x:
## Scope 
### subql-node
1. new table schema for entity
  1. new an internal primary key (real primary key), auto-generated
  2. make id a normal column (indexed)
  3. add a extra column to store available range of blocks for each records
  4. Remove all foreign keys
2. handle primary key generation
3. maintain the block range for each set operation
4. adapt `store`'s set() and get()

### subql-query
1. Refactor/extend/fork postgraphile plugins
    1. Hide special columns: _id, block_range
    2. adapt to using the virtual primary id
        1. map id to ID in graphql schema instead of the real primary key (_id)
        2. Single entity query accept virtual id
    3. Rework relation implementation
        1. Abandon the way to detect relation from foreign key, there will be no foreign keys any more.
        2. Read relation from user’s graphql schema
        3. Refactor sql to use virtual id in the join statements
    4. Query schema supports additional parameters: blockheight/blocknumber 
    5. Wrap all query sql statements with blockrange condition

jamesbayly:
https://github.com/subquery/grants/issues/7

Zedonboy:
Hello.

I am a bit confused, I have a few questions.

1. What State<aspect of network> do you want to version.

2. Are we to create a Subquery project or contribute @subquery/node to have this feature.

Thank you.

PR #858, opened 2022-03-22
Feature/query historical state

Lpaydat:
Create plugins to handle the generated primary key and block height columns.
Modified a plugin to exclude generated primary key from where clause.
Update README

### DIFF
--- a/packages/query/README.md
+++ b/packages/query/README.md
@@ -15,3 +15,35 @@ then run the follow command
 ```
 $ NODE_OPTIONS="-r dotenv/config" yarn start -- --name <subuqery_name> --playground
 ```
+
+## Query Historical State
+
+We added two more special columns to the data to make it versionable: `pid` and `_$block height`
+(these two columns will be hidden from the user and will not appear in the generated graphql schema).
+
+`pid` is the generated primary key, we add this column to resolve issues around ID conflicts and will use the `id` column as virtual primary key instead, allowing user can query a single entity by using its `id`.
+
+```gql
+user(id: "001") {
+  address
+  balance
+}
+```
+
+`_@block_hegiht` is the important column that allowed user to query what the state of the data was at a certain blockheight.
+
+For example, on block 1000 your balance is $2000, then at the block 2200, someone sends you 500 coin, so when you query with `blockHeight: 2100` your result should be `2000`
+
+```gql
+account(id: "0x0004", blockHeight: 2100) {
+  balance  # result is 2000
+}
+```
+
+but if you use `blockHeight: 2200`, your result should be `2500` instead
+
+```gql
+account(id: "0x0004", blockHeight: 2200) {
+  balance  # result is 2500
+}
+```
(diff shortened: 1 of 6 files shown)


jiqiang90:
Thanks for contribution @Lpaydat , we will look at it as soon as we can

ianhe8x:
I haven't ran it locally, though I think it would be good to confirm a few things first.
1. I don't think you can still use foreignkey for relations. Image 
```
Table1： pid, id,  table2_id, _$block_height
A00, 1, B00, [1000,)

Table2: pid,id, data, _$block_height
B00, 1, value_old, [1000,1005)
B01, 1, value_new, [1005,)
```
At block 1000, A00 and B00 are inserted into db.
At block 1005, Table2(id:1) was updated to a new value, thus B01 was inserted and B00's blockheight was chaged from `[1000,)` to `[1000,1005)`

The following query should return different table2 records on different blockheight input
```
table1(blockheight: 1000) {
    nodes{
        table2{
          data
        }
}
}
// return value_old
table1(blockheight: 1005) {
    nodes{
        table2{
          data
        }
}
}
// return value_new
```
You can see foreign key is no use here.

2. if foreign key is not used, (which is also good for insert performance), then to understand the relation between entities, subql-query will need to read the graphql schema defined by user in their query project.

jiqiang90:
@Lpaydat , can you provide an example schema (maybe include some data) for your work? 

I try to write one, and when i apply the blockHeight , I got this error.


```
 "errors": [
    {
      "message": "operator does not exist: bigint @> bigint",
      "locations": [

```

Lpaydat:
@jiqiang90 Sorry for the delayed response; I haven't checked GitHub in a while.

I just updated the repo code after noticing that my earlier code did not solve the issue mentioned by Ian.
My approach may differ somewhat from Ian's in https://github.com/subquery/subql/issues/415 since I continue to rely on foreign keys but redirect the argument value from the `pid` column to the `id` column instead.

The database schema that I tested is as follows:

```
create table hellos
(
  pid                 text    not null    primary key,  -- random generated, but first item needs to use provided `id` as its `pid`
  id                  text    not null,
  value               text    not null,
  _$block_height      int8range  not null
);
```

and

```
create table demos
(
    pid              text    not null   primary key, -- again, first item needs to use the provided `id` is its `pid`
    id               text    not null, -- original primary key
    field            text    not null,
    _$block_height   int8range not null, -- new added versioned field
    hello_id         text    REFERENCES hellos(pid)    -- foreign key to Hello entity
);
```

```
INSERT INTO
    hellos (pid, id, value, _$block_height)
VALUES
    ('B01', 'B01', '0-100', '[0,100)'),
    ('aaa', 'B01', '100-250', '[100,250)'),
    ('bbb', 'B01', '250++', '[250,)');
```

```
INSERT INTO
    demos (pid, id, field, _$block_height, hello_id)
VALUES
    ('A01', 'A01', '0-150', '[0,150)', 'B01'),
    ('xxx', 'A01', '150++', '[150,)', 'B01');
```

The root `blockHeight` value will be the default for that root query and will be passed down to its children, but it will be reset at each root query level.
At the child level, the user can specify a new `blockHeight` which will override the default 'blockHeight' for that field only.

PR #859, opened 2022-03-22
Add support for querying historical state

### LINKED ISSUE #415: Support query by block number (historical state)

spitko:
Resolve #415

- Created a plugin which adds blockHeight as an input argument when underlying table contains column with name "_$block_range".
  - Default value for blockHeight is max bigint
  - blockHeight is then used to filter and return only entities with (_$block_range @> blockHeight)
  - For nested queries, if joining a table that also has "_$block_range" column, same filter is also applied to join
- Hid _$id and _$block_range columns from schema
- Copied and modified PgRowByUniqueConstraint plugin to handle tables with hidden _$id primary key (replace with normal id column)

Relationships between two history tables can be defined using [@foreignKey](https://www.graphile.org/postgraphile/smart-tags/#foreignkey) smart comments to create fake foreign keys (even to non-unique columns), for example: `@foreignKey (table2_id) references table2 (id)`
The join clause is then later modified by PgBlockHeightPlugin to join the two tables using both the foreign key and the blockHeight provided in input

### DIFF
--- a/packages/query/src/graphql/plugins/PgBlockHeightPlugin.ts
+++ b/packages/query/src/graphql/plugins/PgBlockHeightPlugin.ts
@@ -1,0 +1,83 @@
+// Copyright 2022 OnFinality Limited authors & contributors
+// SPDX-License-Identifier: Apache-2.0
+
+import {Plugin} from 'graphile-build';
+import {QueryBuilder} from 'graphile-build-pg';
+import {GraphQLString} from 'graphql';
+
+export const PgBlockHeightPlugin: Plugin = (builder) => {
+  // Adds blockHeight condition to join clause when joining a table that has _$block_height column
+  builder.hook(
+    'GraphQLObjectType:fields:field',
+    (
+      field,
+      {pgSql: sql},
+      {
+        addArgDataGenerator,
+        scope: {
+          isPgBackwardRelationField,
+          isPgBackwardSingleRelationField,
+          isPgForwardRelationField,
+          pgFieldIntrospection,
+        },
+      }
+    ) => {
+      if (!isPgBackwardRelationField && !isPgForwardRelationField && !isPgBackwardSingleRelationField) {
+        return field;
+      }
+      if (
+        !pgFieldIntrospection?.attributes?.some(({name}) => name === '_$block_range') &&
+        !pgFieldIntrospection?.class?.attributes?.some(({name}) => name === '_$block_range')
+      ) {
+        return field;
+      }
+
+      addArgDataGenerator(() => ({
+        pgQuery: (queryBuilder: QueryBuilder) => {
+          queryBuilder.where(
+            sql.fragment`${queryBuilder.getTableAlias()}._$block_range @> ${queryBuilder.context.args.blockHeight}`
+          );
+        },
+      }));
+      return field;
+    }
+  );
+  // Adds blockHeight argument to single entity and connection queries for tables with _$block_height column
+  builder.hook(
+    'GraphQLObjectType:fields:field:args',
+    (
+      args,
+      {extend, pgSql: sql},
+      {addArgDataGenerator, scope: {isPgFieldConnection, isPgRowByUniqueConstraintField, pgFieldIntrospection}}
+    ) => {
+      if (!isPgRowByUniqueConstraintField && !isPgFieldConnection) {
+        return args;
+      }
+      if (
+        !pgFieldIntrospection?.attributes?.some(({name}) => name === '_$block_range') &&
+        !pgFieldIntrospection?.class?.attributes?.some(({name}) => name === '_$block_range')
+      ) {
+        return args;
+      }
+
+      addArgDataGenerator(({blockHeight}) => ({
+        pgQuery: (queryBuilder: QueryBuilder) => {
+          // Save blockHeight to context, so it gets passed down to children
+          if (!queryBuilder.context.args?.blockHeight) {
+            queryBuilder.context.args = {blockHeight: sql.fragment`${sql.value(blockHeight)}::bigint`};
+          }
+          queryBuilder.where(
+            sql.fragment`${queryBuilder.getTableAlias()}._$block_range @> ${queryBuilder.context.args.blockHeight}`
+          );
+        },
+      }));
+      return extend(args, {
+        blockHeight: {
+          description: 'Block height',
+          defaultValue: '9223372036854775807',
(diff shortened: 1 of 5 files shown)


jiqiang90:
Thanks for contribution @spitko , we will look at it as soon as we can

ianhe8x:
I haven't ran it locally, though I think it would be good to confirm a few things first.
1. I don't think you can still use foreignkey for relations. Image 
```
Table1： pid, id,  table2_id, _$block_height
A00, 1, B00, [1000,)

Table2: pid,id, data, _$block_height
B00, 1, value_old, [1000,1005)
B01, 1, value_new, [1005,)
```
At block 1000, A00 and B00 are inserted into db.
At block 1005, Table2(id:1) was updated to a new value, thus B01 was inserted and B00's blockheight was chaged from `[1000,)` to `[1000,1005)`

The following query should return different table2 records on different blockheight input
```
table1(blockheight: 1000) {
    nodes{
        table2{
          data
        }
}
}
// return value_old
table1(blockheight: 1005) {
    nodes{
        table2{
          data
        }
}
}
// return value_new
```
You can see foreign key is no use here.

2. if foreign key is not used, (which is also good for insert performance), then to understand the relation between entities, subql-query will need to read the graphql schema defined by user in their query project.

spitko:
Here's a more complete example:
<details> 
  <summary>Schema</summary>

```
DROP TABLE IF EXISTS app.table1;
CREATE TABLE app.table1 (
	"_$id" bigint GENERATED ALWAYS AS IDENTITY PRIMARY KEY,
	"_$block_range" int8range NOT NULL,
	"id" TEXT NOT NULL,
	"table2_id" text NOT NULL
);

DROP TABLE IF EXISTS app.table2;
CREATE TABLE app.table2 (
	"_$id" 	 bigint GENERATED ALWAYS AS IDENTITY PRIMARY KEY,
	"_$block_range" int8range NOT NULL,
	"id" TEXT NOT NULL,
	"data" TEXT NOT NULL
);

COMMENT ON TABLE app.table1 IS E'@foreignKey (table2_id) REFERENCES table2 (id)';

INSERT INTO app.table2 ("_$block_range", "id", "data") VALUES (int8range(1000, 1005, '[)'), 1, 'old');
INSERT INTO app.table2 ("_$block_range", "id", "data") VALUES (int8range(1005, null, '[)'), 1, 'current');
INSERT INTO app.table1 ("_$block_range", "id", "table2_id") VALUES (int8range(1000, null, '[)'), 1, 1);
```
</details>

There is no SQL foreign key constraint, only a comment on the table, which is read by subql-query and used to create the relationship between the two tables.

Without the custom plugin, the SQL generated for your query is:
```
[...]
from "app"."table2" as __local_2__
where (__local_1__."table2_id" = __local_2__."id") and (TRUE) and (TRUE)
)))) as "@nodes" from (select __local_1__.*
from "app"."table1" as __local_1__
where (__local_1__._$block_range @> $1::bigint) and (TRUE) and (TRUE)
[...]
```
However, after enabling the plugin, the join condition is modified to include the blockHeight:
```
[...]
from "app"."table2" as __local_2__
where (__local_1__."table2_id" = __local_2__."id") and (__local_2__._$block_range @> $1::bigint) and (TRUE) and (TRUE)
)))) as "@nodes" from (select __local_1__.*
from "app"."table1" as __local_1__
where (__local_1__._$block_range @> $1::bigint) and (TRUE) and (TRUE)
[...]
```

jiqiang90:
Hi @spitko , I have a few questions.

I tested your feature with my sql script, (https://gist.github.com/jiqiang90/812bb4e932a3bb36c6c14dbabed8ebf2), this is  a many-to-many relationship with reverse lookup we had in the documents (https://doc.subquery.network/create/graphql/#reverse-lookups).

I found a few things:

1. the `from` field (Account) is missing when query entity Transfer , and here it has been comment on table (https://gist.github.com/jiqiang90/812bb4e932a3bb36c6c14dbabed8ebf2#file-testhistorical-sql-L19).
Screen Shot 2022-03-25 at 4 15 09 PM

2. At the moment our user project schema allow user to redefine the reverse lookup field name, for example, `sentTransfers` and it is enabled by 
```
COMMENT ON CONSTRAINT transfers_from_id_fkey ON "subql-starter"."transfers" IS E'@foreignFieldName sentTransfers';

```
Do you have any idea how this can be work together with the historical state?

Thank you

spitko:
> 1. the `from` field (Account) is missing when query entity Transfer , and here it has been comment on table (https://gist.github.com/jiqiang90/812bb4e932a3bb36c6c14dbabed8ebf2#file-testhistorical-sql-L19).

There can only be one comment on an object, so in this case the second comment overwrites the first. Instead you should use a single comment and add `\n` between the tags:
`COMMENT ON TABLE "subql-starter".transfers IS E'@foreignKey (from_id) REFERENCES accounts (id)\n@foreignKey (to_id) REFERENCES accounts (id)';`



> 2\. At the moment our user project schema allow user to redefine the reverse lookup field name, for example, `sentTransfers` and it is enabled by

Add the `@foreignFieldName` tag at the end of the comment, [seperated by a pipe](https://www.graphile.org/postgraphile/smart-comments/#adding-smart-tags-to-fake-constraints):
`COMMENT ON TABLE "subql-starter".transfers IS E'@foreignKey (from_id) REFERENCES accounts (id)|@foreignFieldName sentTransfers';`

jiqiang90:
Hi @spitko , one small thing, can we only enable this feature with the `experimental-historical` flag? Thanks

spitko:
> Hi @spitko , one small thing, can we only enable this feature with the `experimental-historical` flag? Thanks

Done, moved the two new plugins behind this flag.

stwiname (approved):

PR #869, opened 2022-04-04
Update vm2

jiqiang90:


stwiname (approved):

ianhe8x:
Do we need to remove the custom resolve fn we passed to vm2?

stwiname:
A quick test on my end seems that we don't need the custom resolve function

PR #882, opened 2022-04-05
remove trigger if subscription disabled

### LINKED ISSUE #857: Subscription check notification trigger
https://github.com/subquery/subql/blob/1a719fb26cc4cd04900122e3918a27c062b2afb6/packages/node/src/utils/sync-helper.ts#L42

Notification trigger should never been replaced.

- [x] remove trigger if unsafe disabled

### LINKED ISSUE #872: Remove notification trigger if subscription is disabled

jiqiang90:
Closes #857

### DIFF
--- a/packages/node/src/utils/sync-helper.ts
+++ b/packages/node/src/utils/sync-helper.ts
@@ -72,9 +72,12 @@ export function dropNotifyTrigger(schema: string, table: string): string {
     ON "${schema}"."${table}";`;
 }
 
+export function getNotifyTriggers(schema: string, table: string): string {
+  return `select trigger_name as "triggerName", event_manipulation as "eventManipulation" from information_schema.triggers
+          WHERE trigger_name = '${schema}_${table}_notify_trigger'`;
+}
 export function createNotifyTrigger(schema: string, table: string): string {
   return `
-${dropNotifyTrigger(schema, table)}
 CREATE TRIGGER "${schema}_${table}_notify_trigger"
     AFTER INSERT OR UPDATE OR DELETE
     ON "${schema}"."${table}"
--- a/packages/node/src/utils/sync-helper.ts
+++ b/packages/node/src/utils/sync-helper.ts
@@ -72,9 +72,9 @@ export function dropNotifyTrigger(schema: string, table: string): string {
     ON "${schema}"."${table}";`;
 }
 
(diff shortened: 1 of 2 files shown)


jiqiang90:
https://github.com/subquery/subql/blob/1a719fb26cc4cd04900122e3918a27c062b2afb6/packages/node/src/utils/sync-helper.ts#L42

In https://github.com/subquery/subql/issues/857 we mentioned the function should never been replaced. 

Not sure how to make it `CREATE IF NOT EXISTS FUNCTION send_notification()` here.

Also check the function again, seems it is harmless, any thought? @stwiname @ianhe8x

stwiname (approved):
Replace function seems fine to me. I'm not an expert at this though

packages/node/src/utils/sync-helper.ts
```diff
@@ -67,10 +67,14 @@ BEGIN
...
+export function dropNotifyTrigger(schema: string, table: string): string {
+  return `DROP TRIGGER IF EXISTS "${schema}_${table}_notify_trigger"
+    ON "${schema}"."${table}";`;
+}
+
 export function createNotifyTrigger(schema: string, table: string): string {
   return `
-DROP TRIGGER IF EXISTS "${schema}_${table}_notify_trigger"
-    ON "${schema}"."${table}";
+${dropNotifyTrigger(schema, table)}
```
ianhe8x, line 77: Do you plan to improve here? Every restart will drop and recreate trigger doesn't seems a good way.

jiqiang90: Seems there is no `CREATE IF NOT EXISTS` on trigger.

stwiname (approved):

packages/node/src/utils/sync-helper.ts
```diff
@@ -72,9 +72,12 @@ export function dropNotifyTrigger(schema: string, table: string): string {
     ON "${schema}"."${table}";`;
 }
 
+export function getNotifyTriggers(schema: string, table: string): string {
+  return `select trigger_name as "triggerName", event_manipulation as "eventManipulation" from information_schema.triggers
```
ianhe8x, line 76: Use query parameter to avoid sql injection

ianhe8x (approved):

PR #885, opened 2022-04-06
add mantlemint support

DeveloperInProgress:


packages/node-terra/src/indexer/apiterra.service.ts
```diff
@@ -76,10 +89,54 @@ export class TerraClient {
...
+  async blockInfoMantlemint(height?: number): Promise<BlockInfo> {
+    const config = {
+      method: 'get',
+      url: `${this.mantlemintURL}/index/blocks/${height}`,
+    };
+    return axios(config)
+      .then((d) => d.data)
+      .catch((e) => {
+        throw e;
+      });
```
stwiname, line 102: We can use async await for these. (can also remove the eslint disable)

```suggestion
    const { data } = await axios(config);
    return data;
```

packages/node-terra/src/utils/terra-helper.ts
```diff
@@ -102,10 +107,18 @@ export function filterEvents(
...
 
 async function getBlockByHeight(api: TerraClient, height: number) {
-  return api.blockInfo(height).catch((e) => {
+  let blockInfo: BlockInfo;
+  try {
+    if (api.mantlemintHealthOK) {
+      blockInfo = await api.blockInfoMantlemint(height);
+    } else {
+      blockInfo = await api.blockInfo(height);
+    }
```
stwiname, line 116: I think we can move this case to the TerraApi. This file shouldn't be concerned about the implementation that is used to fetch the content. We can do the same for TxInfo
 
```ts
// In TerraApi
async blockInfo(height) {
  if (this.mantlemintHealthOK) {
     return this.blockInfoMantlemint(height);
  }
  
  // return current implementation
}

```

packages/node-terra/src/utils/terra-helper.ts
```diff
@@ -212,7 +225,16 @@ export async function fetchTerraBlocksBatches(
...
       }
-      const txInfos = await getTxInfobyHashes(api, txHashes);
+
+      let txInfos: TxInfo[];
+      if (api.mantlemintHealthOK) {
+        txInfos = await api.txsByHeightMantlemint(
+          blockInfo.block.header.height,
+        );
+      } else {
+        txInfos = await getTxInfobyHashes(api, txHashes);
```
stwiname, line 235: `getTxInfobyHashes` can be moved to the TerraApi

packages/node-terra/src/indexer/apiterra.service.ts
```diff
@@ -76,10 +89,54 @@ export class TerraClient {
...
+    const health = axios(config)
+      .then((d) => d.data)
+      .catch((e) => {
+        throw e;
+      });
+    return health === 'OK';
+  }
+
+  // eslint-disable-next-line @typescript-eslint/require-await
+  async txsByHeightMantlemint(height: string): Promise<TxInfo[]> {
```
stwiname, line 128: How sure are we on this return type? Could there be differences in encoding? Or anything that needs to be parsed?

DeveloperInProgress: The returned object had the same set of keys as in TxInfo

stwiname (approved):

stwiname:
@ianhe8x what are your thoughts on providing the mantlemint endpoint? Its currently just a CLI flag and is additional to the RPC endpoint

packages/node-terra/src/indexer/apiterra.service.ts
```diff
@@ -77,19 +78,36 @@ export class TerraClient {
...
   async blockInfo(height?: number): Promise<BlockInfo> {
     if (this.mantlemintHealthOK && height) {
       return this.blockInfoMantlemint(height);
     }
-    return this.baseApi.tendermint.blockInfo(height, this.params);
+    const config = {
+      method: 'get',
+      url: height
+        ? `${this.tendermintURL}/cosmos/base/tendermint/v1beta1/blocks/${height}`
+        : `${this.tendermintURL}/cosmos/base/tendermint/v1beta1/blocks/latest`,
```
stwiname, line 105:
```suggestion
url: `${this.tendermintURL}/cosmos/base/tendermint/v1beta1/blocks/${height ?? 'latest'}`

```

DeveloperInProgress: looks good

PR #892, opened 2022-04-13
Rework handling spec versions

stwiname:
I'm not super happy with this approach, its quite hard to work with data with the blockbuffer queue. It also might make more sense to separate the spec version dictionary query into another request separate of block numbers

- Introduce SpecVersion service to keep track of spec version data
- Update dictinary query to filter spec versions
- When dictionary isn't used, be smarter about fetching spec versions

### DIFF
--- a/packages/node/src/indexer/SpecVersions.service.ts
+++ b/packages/node/src/indexer/SpecVersions.service.ts
@@ -1,0 +1,69 @@
+// Copyright 2020-2022 OnFinality Limited authors & contributors
+// SPDX-License-Identifier: Apache-2.0
+
+import { Injectable } from '@nestjs/common';
+import { ApiPromise } from '@polkadot/api';
+import { BlockHash, RuntimeVersion } from '@polkadot/types/interfaces';
+
+export type SpecVersionMap = { [blockNumber: number]: number }; // blockNumber => specVersion
+
+export function getSpecVersionBlockNumbers(
+  specVersions: SpecVersionMap,
+): number[] {
+  return Object.keys(specVersions).map((key) => parseInt(key, 10));
+}
+
+@Injectable()
+export class SpecVersionService {
+  private _specVersions: SpecVersionMap;
+
+  private _runtimeVersions: { [specVersion: number]: RuntimeVersion };
+  private _lastQueriedHeight = 0; // Keep track of the latest know block spec version
+
+  constructor() {
+    this._specVersions = {};
+    this._runtimeVersions = {};
+  }
+
+  addSpecVersions(
+    specVersions: SpecVersionMap,
+    lastQueriedHeight?: number,
+  ): void {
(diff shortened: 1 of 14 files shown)


packages/node/src/indexer/SpecVersions.service.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  async getRuntimeVersion(
+    api: ApiPromise,
+    blockNumber: number,
+    blockHash: string | BlockHash,
+  ): Promise<RuntimeVersion | undefined> {
+    const specVersion = this.getSpecVersion(blockNumber);
+    if (!this._runtimeVersions[specVersion]) {
+      this._runtimeVersions[specVersion] =
+        await api.rpc.state.getRuntimeVersion(blockHash);
```
ianhe8x, line 54: You need parentHash to know specVersion of current block

packages/node/src/indexer/fetch.service.ts
```diff
@@ -364,16 +368,18 @@ export class FetchService implements OnApplicationShutdown {
...
             this.dictionaryQueryEntries,
           );
-          //TODO
-          // const specVersionMap = dictionary.specVersions;
           if (
             dictionary &&
             this.dictionaryValidation(dictionary, startBlockHeight)
           ) {
-            const { batchBlocks } = dictionary;
+            const { batchBlocks, specVersions } = dictionary;
```
ianhe8x, line 379: You can get all chain upgrades in one query, since there won't be many. Then you don't need to query it in every batch

packages/node/src/indexer/SpecVersions.service.ts
```diff
@@ -0,0 +1,69 @@
...
+
+export function getSpecVersionBlockNumbers(
+  specVersions: SpecVersionMap,
+): number[] {
+  return Object.keys(specVersions).map((key) => parseInt(key, 10));
+}
+
+@Injectable()
+export class SpecVersionService {
+  private _specVersions: SpecVersionMap;
```
ianhe8x, line 18: When dictionary is used, this is fine since its length is very limited. But when dictionary is not used, this could be too many to be kept in memory, and unnecessary.

packages/node/src/indexer/SpecVersions.service.ts
```diff
@@ -0,0 +1,69 @@
...
+export type SpecVersionMap = { [blockNumber: number]: number }; // blockNumber => specVersion
+
+export function getSpecVersionBlockNumbers(
+  specVersions: SpecVersionMap,
+): number[] {
+  return Object.keys(specVersions).map((key) => parseInt(key, 10));
+}
+
+@Injectable()
+export class SpecVersionService {
```
ianhe8x, line 17: Basically this is a local cache of spec versions. It can be merged back into api.service.

Issue #855, opened 2022-03-21
Update mmr poi logic with store operation


jiqiang90:
Set offset value until first store operation, currently offset is calculate from user project start block height. 
Also, start write .mmr until first store operation, so its 1st leaf height match with the poi 1st record.

PR #894, opened 2022-04-14
Update mmr

jiqiang90:
-  https://github.com/subquery/merkle-mountain-range/pull/2 merge and release, to resolve inconsistent mmr value issue
-  changed offset value to the 1st indexing opertaion block height -1
-  remove unnecessary steps in appendMmr, as this is steps is duplicate in ensureStartHeight
-  testing docker file https://gist.github.com/jiqiang90/dfc4210c6d0c68f42bdbc232c7a05ddb


jiqiang90:
Closes #855

packages/node/src/indexer/indexer.manager.ts
```diff
@@ -170,6 +170,14 @@ export class IndexerManager {
     this.fetchService.latestProcessed(block.block.header.number.toNumber());
     this.prevSpecVersion = block.specVersion;
     if (this.nodeConfig.proofOfIndex) {
+      if (this.blockOffset === undefined) {
```
ianhe8x, line 173: I can see a reason you move the blockOffset initialization here is to skip no operation blocks. If so, the better place is to move it to line 162 and attach it to tx.

packages/node/src/indexer/mmr.service.ts
```diff
@@ -52,54 +52,71 @@ export class MmrService implements OnApplicationShutdown {
...
 
-  async fetchBlockOffsetFromDb(): Promise<number | null> {
-    const blockOffset = await this.metadataRepo.findOne({
-      where: { key: 'blockOffset' },
-    });
-    if (blockOffset === null) {
-      throw new Error(`Poi service failed to fetch block offset from metadata`);
+  // await index 1st block and set the blockOffset
+  async awaitFetchBlockOffsetFromDb(): Promise<number> {
+    while (!this.isShutdown && this.blockOffset === null) {
```
ianhe8x, line 111: don't get why use `while` here

ianhe8x (changes requested):
there're redundant codes can be optimised

packages/node/src/indexer/indexer.manager.ts
```diff
@@ -214,13 +226,17 @@ export class IndexerManager {
...
     this.filteredDataSources = this.filterDataSources(startHeight);
     this.fetchService.register((block) => this.indexBlock(block));
+  }
 
-    if (this.nodeConfig.proofOfIndex) {
-      void this.mmrService.syncFileBaseFromPoi().catch((err) => {
+  private setBlockOffset(offset: number): void {
+    this.blockOffset = offset;
+    logger.info(`set blockoffset to ${offset}`);
+    void this.mmrService
```
stwiname, line 234: What's the reason for not awaiting this?

jiqiang90: I assume you are asking  the `mmr.syncFileBaseFromPoi` here, while sync the between poi and filebased mmr,  this should be independent and we try to not block indexing service.

stwiname (approved):

ianhe8x (approved):

PR #899, opened 2022-04-20
add dynamic datasources support to terra

DeveloperInProgress:


stwiname (commented):
Can you please check that codegen works for terra templates.

stwiname (approved):

PR #900, opened 2022-04-21
drop support for specVersion 0.0.1

jiqiang90:


packages/cli/src/commands/init.ts
```diff
@@ -84,15 +84,6 @@ export default class Init extends Command {
 
   async run(): Promise<void> {
     const {args, flags} = await this.parse(Init);
-
-    if (flags.specVersion === '0.0.1') {
```
stwiname, line 88: Did we want this message for `< 1.0.0`?

jiqiang90: Good idea, updated this message

stwiname (approved):

PR #901, opened 2022-04-21
allow cli run tsc check before build

### LINKED ISSUE #861: Have output of tsc when running `subql build`
There should be a way to run typescript check, so we get the output of build errors when running `subql build`

jiqiang90:
Closes #861

### DIFF
--- a/packages/cli/src/commands/build.ts
+++ b/packages/cli/src/commands/build.ts
@@ -4,8 +4,7 @@
 import {lstatSync, readFileSync} from 'fs';
 import path from 'path';
 import {Command, Flags} from '@oclif/core';
-import cli from 'cli-ux';
-import {runWebpack, typescriptBuildCheck} from '../controller/build-controller';
+import {runWebpack} from '../controller/build-controller';
 
 export default class Build extends Command {
   static description = 'Build this SubQuery project code';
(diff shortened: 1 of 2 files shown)


packages/cli/src/controller/build-controller.ts
```diff
@@ -45,6 +46,20 @@ const getBaseConfig = (
   },
 });
 
+export async function typescriptBuildCheck(projectDir: string): Promise<void> {
+  await new Promise((resolve) => {
+    childProcess.exec('tsc --noEmit', {cwd: projectDir}, (error, stdout, stderr) => {
```
stwiname, line 51: I think this assumes that `tsc` command is available.

packages/cli/src/commands/build.ts
```diff
@@ -53,6 +53,8 @@ export default class Build extends Command {
         }
       }
 
+      //Checking need to be done before cli.action,otherwise error logs won't display
+      await typescriptBuildCheck(directory);
       cli.action.start('Building and packing code');
```
stwiname, line 58: I think this suppresses any other logging, we should be able to get webpack to output TS errors but we might need to disable this line.

stwiname (approved):

PR #923, opened 2022-04-27
Allow set query-complexity for query

jiqiang90:
```
query{
  starterEntities(first:5){
    nodes{
      field1
    }
  }
}

```
This will count complexity to 3

default value is 10,  to change this by set `query-complexity=20`

### DIFF
--- a/packages/query/package.json
+++ b/packages/query/package.json
@@ -44,6 +44,7 @@
     "graphile-build-pg": "^4.12.2",
     "graphile-utils": "^4.12.2",
     "graphql": "^15.8.0",
+    "graphql-query-complexity": "^0.11.0",
     "lodash": "^4.17.21",
     "pg": "^8.7.1",
     "postgraphile": "^4.12.9",
--- a/packages/query/src/graphql/graphql.module.ts
+++ b/packages/query/src/graphql/graphql.module.ts
@@ -21,6 +21,7 @@ import {PinoConfig} from '../utils/logger';
 import {getYargsOption} from '../yargs';
 import {plugins} from './plugins';
 import {PgSubscriptionPlugin} from './plugins/PgSubscriptionPlugin';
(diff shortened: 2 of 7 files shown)


stwiname (approved):
Is there more to complexity than just depth? Its hard to know if 10 is an appropriate default

ianhe8x:
I forget where I saw it, but anyway there's a cool idea that we can return the complexity in response header

jiqiang90:
> I forget where I saw it, but anyway there's a cool idea that we can return the complexity in response header

we just need to add this .

```
 willSendResponse({ response }) {
          response.extensions = response.extensions || {};
          response.extensions.complexity = complexity;
          response.extensions.maxComplexity = maxComplexity;
        },
```

ianhe8x:
> > I forget where I saw it, but anyway there's a cool idea that we can return the complexity in response header
> 
> we just need to add this .
> 
> ```
>  willSendResponse({ response }) {
>           response.extensions = response.extensions || {};
>           response.extensions.complexity = complexity;
>           response.extensions.maxComplexity = maxComplexity;
>         },
> ```

@jiqiang90 Can you add this and make sure 10 is enough for our usecase.

ianhe8x:
@jiqiang90 Some test failures about graphql

jamesbayly:
I like returning the query complexity in the header
I do not think we should set a default in SubQL and it will only piss people off when running it locally
We should only set a default in the managed service, we need to carefully consider the number

jamesbayly:
I'm very against setting a default value in this package actually

jiqiang90:
> I like returning the query complexity in the header I do not think we should set a default in SubQL and it will only piss people off when running it locally We should only set a default in the managed service, we need to carefully consider the number

Sounds reasonable to me, I will remove the restriction,  only implement plugin when `query-complexity` is set,

stwiname (approved):

PR #968, opened 2022-05-04
Mmr query api 

### LINKED ISSUE #913: Api to query latest mmr in filebase db
In order to fetch the latest processed block MMR, using the value from the Postgres DB POI table is not accurate enough, as it only records blocks that have store operation. Therefore, this value should be read from the project Filebase database. This will require: 

- We need to provide an API to query the project filebase database
- After the project is completely synced,  we can insert MMR value to filebased DB as per block indexed. We don't need to wait for the next poi block height to query the POI table. 
- Consider removing reset logic, only verify the last poi with its MMR value against the filebased DB.

jiqiang90:
Closes #913


jiqiang90:
Todo:

- [ ] still unsure how to validate the blockoffset

packages/node/src/meta/mmrQuery.controller.ts
```diff
@@ -0,0 +1,54 @@
...
+          error: e.message,
+        },
+        HttpStatus.INTERNAL_SERVER_ERROR,
+      );
+    }
+  }
+
+  @Get()
+  async getLatestMmr(@Query() query) {
+    if (query.latest) {
```
stwiname, line 39: What happens if it's not latest?

packages/node/src/meta/mmrQuery.controller.ts
```diff
@@ -0,0 +1,54 @@
...
+  Query,
+} from '@nestjs/common';
+import { getLogger } from '../utils/logger';
+import { MmrQueryService } from './mmrQuery.service';
+
+const logger = getLogger('mmrs');
+
+@Controller('mmrs')
+export class MmrQueryController {
+  constructor(private mmrService: MmrQueryService) {}
```
stwiname, line 19: Whats the reason for not using `MmrService` instead?

packages/node/src/indexer/mmr.service.ts
```diff
@@ -53,42 +48,28 @@ export class MmrService implements OnApplicationShutdown {
...
+    this.blockOffset = blockOffset;
 
     // The file based database current leaf length
     const fileBasedMmrLeafLength = await this.fileBasedMmr.getLeafLength();
     // However, when initialization we pick the previous block for file db and poi mmr validation
     // if mmr leaf length 0 ensure the next block height to be processed min is 1.
     this.nextMmrBlockHeight = fileBasedMmrLeafLength + blockOffset + 1;
     // The first Poi in database with null mmr value.
-    const poiWithoutMmr = await this.getFirstPoiWithoutMmr();
+    // const poiWithoutMmr = await this.getFirstPoiWithoutMmr();
```
stwiname, line 60: Can be removed

stwiname (approved):

packages/node/src/meta/mmrQuery.controller.ts
jiqiang90: Thanks, I have make it throw if query not supported

jiqiang90:
This require more test before include in release, as we have added logic for handle empty blocks, better merge after 1.0.0

packages/node/src/meta/mmrQuery.controller.ts
```diff
@@ -0,0 +1,54 @@
...
+
+@Controller('mmrs')
+export class MmrQueryController {
+  constructor(private mmrService: MmrService) {}
+
+  @Get(':blockHeight')
+  async getMmr(@Param() params) {
+    try {
+      // add await, otherwise error been skipped
+      return await this.mmrService.getMmr(params.blockHeight);
```
ianhe8x, line 25: I'm not sure what is the default way nestjs serialize a Uint8Array. Can you make the process more explicit?

packages/node/src/meta/mmrQuery.controller.ts
ianhe8x: Can you make `/mmrs` return latest mmr  and `/mmrs/xxx` return mmr for xxx height?

packages/node/src/indexer/indexer.manager.ts
```diff
@@ -151,19 +166,13 @@ export class IndexerManager {
...
-                value: blockHeight - 1,
-              },
-              { transaction: tx },
-            );
-            this.setBlockOffset(blockHeight - 1);
-          }
           poiBlockHash = poiBlock.hash;
           await this.storeService.setPoi(poiBlock, { transaction: tx });
           this.poiService.setLatestPoiBlockHash(poiBlockHash);
+          await this.storeService.setMetadataBatch(
```
ianhe8x, line 172: `lastPoiHeight` isn't the height that we have a poi record, but the height that we have mmrRoot calculated.

ianhe8x: lastPoiHeight is ok to be not realtime, so you won't do too many db writes when you are catching up mmr root calculation to poi table.

packages/node/src/indexer/events.ts
```diff
@@ -36,3 +36,9 @@ export interface NetworkMetadataPayload {
   specName: string;
   genesisHash: string;
 }
+
+export interface MmrPayload {
+  leafIndex: number;
```
ianhe8x, line 41: Why we need it?

ianhe8x:
I still have concerns that increase wordsize isn't the right way to go, and it increases storage size.

ianhe8x (approved):

PR #972, opened 2022-05-05
make metadata chain name consistent

jiqiang90:


stwiname (approved):
Does this break anything? Mainly around the terra dictionary

PR #981, opened 2022-05-06
Update dependencies

jiqiang90:


stwiname:
I think this needs to be updated too https://github.com/subquery/subql/blob/main/packages/utils/package.json#L15

stwiname (approved):

PR #992, opened 2022-05-08
Support query by block number

### LINKED ISSUE #415: Support query by block number (historical state)

spitko:
Resolve #415 

- [x] Add `_id` and `_block_range` columns
- [x] Add exclude constraint on (`id`, `_block_range`)
- [x] Remove foreign keys and instead track relations in smart comments
- [x] Add sequelize hooks to keep `_block_range` up to date
- [x] Handle unique/other indexes
- [x] Fix subscription notification trigger

### DIFF
--- a/packages/node/src/indexer/store.service.ts
+++ b/packages/node/src/indexer/store.service.ts
@@ -99,7 +99,7 @@ export class StoreService {
   ): Promise<void> {
     this.schema = schema;
     this.modelsRelations = modelsRelations;
-    this.historical = !!argv['experimental-historical'];
+    this.historical = await this.getHistoricalStateEnabled();
     try {
       await this.syncSchema(this.schema);
     } catch (e) {
@@ -298,7 +298,6 @@ export class StoreService {
       this.poiRepo = PoiFactory(this.sequelize, schema);
     }
     this.metaDataRepo = MetadataFactory(this.sequelize, schema);
-    await this.checkHistoricalState();
 
     await this.sequelize.sync();
     await this.setMetadata('historicalStateEnabled', this.historical);
@@ -307,20 +306,24 @@ export class StoreService {
     }
   }
 
-  async checkHistoricalState() {
-    let historicalStateEnabled: MetadataModel | null = null;
+  async getHistoricalStateEnabled(): Promise<boolean> {
+    let enabled = true;
     try {
-      historicalStateEnabled = await this.metaDataRepo.findByPk(
-        'historicalStateEnabled',
+      // Throws if _metadata doesn't exist (first startup)
+      const result = await this.sequelize.query(
+        `SELECT value FROM "${this.schema}"."_metadata" WHERE key = 'historicalStateEnabled'`,
+        { type: QueryTypes.SELECT },
       );
+      if (result.length > 0) {
+        enabled = result[0].value;
+      } else {
+        enabled = false;
+      }
     } catch (e: any) {
-      // ignored
+      enabled = !argv['disable-historical'];
     }
-    assert(
-      !historicalStateEnabled ||
-        historicalStateEnabled.value === this.historical,
-      "Option 'experimental-historical' cannot be changed after starting indexing",
-    );
+    logger.info(`Historical state is ${enabled ? 'enabled' : 'disabled'}`);
+    return enabled;
   }
 
   addBlockRangeColumnToIndexes(indexes: IndexesOptions[]) {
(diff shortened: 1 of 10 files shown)


packages/node/src/indexer/store.service.ts
```diff
@@ -80,6 +95,7 @@ export class StoreService {
   ): Promise<void> {
     this.schema = schema;
     this.modelsRelations = modelsRelations;
+    this.historical = argv['experimental-historical'];
```
stwiname, line 98: We need further checks that the project was started with this enabled, it should not be possible to add this in after a project has started indexing.

There also needs to be handling for if the project is started with this flag then it is no longer enabled. I think we can throw an error in this case.

packages/node/src/indexer/store.service.ts
```diff
@@ -379,6 +505,28 @@ group by
     return rows.map((result) => camelCaseObjectKey(result)) as IndexField[];
   }
 
+  private async markAsDeleted(model: ModelStatic<any>, id: string) {
```
stwiname, line 508: This should use the transaction.

packages/node/src/indexer/store.service.ts
```diff
@@ -249,6 +279,13 @@ export class StoreService {
           throw new Error('Relation type is not supported');
       }
     }
+    foreignKeyMap.forEach((keys, tableName) => {
```
stwiname, line 282: Should this be under a `this.historical` check?

stwiname (commented):
Nice work @spitko, you've made good progress. 

I have left some comments, do you think it would be feasible to add some tests to the store service?

@jiqiang90 can you please have a look over this.

packages/node/src/indexer/store.service.ts
spitko: done

packages/node/src/indexer/store.service.ts
spitko: it does :+1:

packages/node/src/indexer/store.service.ts
spitko: not really, it's not a no-op if historical=false

packages/node/src/indexer/store.service.ts
```diff
@@ -249,17 +287,140 @@ export class StoreService {
...
+        .map((tags) => smartTags(tags, '|'))
+        .join('\n');
+      const query = commentTableQuery(`"${schema}"."${tableName}"`, comment);
+      extraQueries.push(query);
+    });
     if (this.config.proofOfIndex) {
       this.poiRepo = PoiFactory(this.sequelize, schema);
     }
     this.metaDataRepo = MetadataFactory(this.sequelize, schema);
+    this.checkHistoricalState();
```
jiqiang90, line 301: missing `await`

packages/node/src/indexer/store.service.ts
```diff
@@ -249,17 +287,140 @@ export class StoreService {
...
 
     await this.sequelize.sync();
+    this.setMetadata('historicalStateEnabled', this.historical);
```
jiqiang90, line 304: missing `await`

stwiname:
Hey @spitko 

Thank you for your work. We've tested this and we're happy with the functionality.

I'd like to request a couple of changes relating to the cli flags:

* On `query` can we please remove the `experimental-historical` flag and decide based on the db schemas whether this should be enabled and log whether it is enabled or not on startup
* On the `node`
    * Can we please default historical to be on
    * Remove `experimental-historical` and replace it with `disable-historical`, which would disable historical if provided
    * If the project was initially started without historical feature it should always be disabled
    * Log whether historical is enabled or not on startup

spitko:
Made it so that the value of --disable-historical is only checked on first startup and the value from _metadata table (or false for pre-existing projects) is used afterwards.

stwiname (approved):

PR #995, opened 2022-05-09
Add TerraSafeApi

DeveloperInProgress:


packages/node-terra/src/indexer/apiterra.service.ts
```diff
@@ -233,3 +249,98 @@ export class TerraClient {
...
+    if (this.mantlemintURL) {
+      this._mantlemintConnection = axios.create({
+        baseURL: this.mantlemintURL,
+        headers: {
+          Accept: 'application/json',
+          'User-Agent': `SubQuery-Node ${packageVersion}`,
+        },
+      });
+    }
+  }
```
stwiname, line 287: We want to reuse these connections, they should be shared between any instances of safeapi. This is to keep the connections limited to the endpoint

packages/types-terra/src/interfaces.ts
```diff
@@ -50,3 +50,9 @@ export interface TerraEvent<M extends Msg = Msg> {
 export type TerraContractMessage = TerraMessage<WasmMsg>;
 
 export type DynamicDatasourceCreator = (name: string, args: Record<string, unknown>) => Promise<void>;
+
+export interface ITerraSafeApi {
```
stwiname, line 54: We want this api to match LCD client, but partially implemented. Currently these methods are a layer too high. We want it to be `api.wasm.contractInfo` but it is currently `api.contractInfo`. 

Follow the terra repo for more https://github.com/terra-money/terra.js/blob/main/src/client/lcd/LCDClient.ts#L100

This should look something like:
```ts
export interface SafeLCD {

    wasm: typeof WasmAPI;
}

```

packages/node-terra/src/indexer/apiterra.service.ts
```diff
@@ -233,3 +249,98 @@ export class TerraClient {
...
+
+  async codeInfo(codeID: number): Promise<CodeInfo> {
+    const preferredConnection = this.mantlemintHealthOK
+      ? this._mantlemintConnection
+      : this._lcdConnection;
```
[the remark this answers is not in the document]
DeveloperInProgress: we do mantlemint health check after the creation of TerraSafeApi object: https://github.com/DeveloperInProgress/subql/blob/3dc6478d592e1ab496458cc5989a1e8d84855f1d/packages/node-terra/src/indexer/apiterra.service.ts#L90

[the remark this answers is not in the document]
DeveloperInProgress: changed it to pass mantlemint health data into the constructor so that we can accomodate preferredConnection assignment inside constructor

packages/node-terra/src/indexer/apiterra.service.ts
```diff
@@ -233,3 +250,65 @@ export class TerraClient {
...
+      `/terra/wasm/v1beta1/codes/${codeID}`,
+    );
+    return <CodeInfo>{
+      code_id: Number.parseInt(data.code_info.code_id),
+      code_hash: data.code_info.code_hash,
+      creator: data.code_info.creator,
+    };
+  }
+
+  async contractInfo(contractAddress: string): Promise<ContractInfo> {
```
stwiname, line 286: Do these 2 methods support specifying the height parameter?

packages/types-terra/src/interfaces.ts
```diff
@@ -50,3 +61,13 @@ export interface TerraEvent<M extends Msg = Msg> {
 export type TerraContractMessage = TerraMessage<WasmMsg>;
 
 export type DynamicDatasourceCreator = (name: string, args: Record<string, unknown>) => Promise<void>;
+
+export interface ISafeWasmApi {
```
stwiname, line 65: Can this be `type ISafeWasmApi = typeof WasmAPI`? If we can avoid defining our own types that would be ideal

stwiname (commented):
Can you please test out with this project https://github.com/Talis-Art/notification-index. Currently there are a few cases of `(global as any).unsafeApi` that will need to be updated to `api`

packages/node-terra/src/indexer/apiterra.service.ts
DeveloperInProgress: No, looks like there is no height parameter on these methods.

stwiname:
A couple of the tests are failing, once they are resolved I am happy

DeveloperInProgress:
made an update, tests must be passing now

DeveloperInProgress:
looks like terra.js accepts string as a type for execute_msg in MsgExecuteContract now: https://github.com/terra-money/terra.js/commit/3608415f9cff77fe24a0905051b15d433d326b29

packages/node-terra/src/utils/terra-helper.ts
```diff
@@ -40,7 +41,11 @@ export function filterMessageData(
   if (
     filter.type === '/terra.wasm.v1beta1.MsgExecuteContract' &&
     filter.contractCall &&
-    !(filter.contractCall in (data.msg as MsgExecuteContract).execute_msg)
+    typeof (data.msg as MsgExecuteContract).execute_msg === 'object' &&
```
stwiname, line 44: I think if its type string then we should do `execute_msg.contains(filter.ContractCall)`

stwiname (approved):

PR #1001, opened 2022-05-10
remove use manifest in codegen

jiqiang90:


packages/common-substrate/src/project/utils.ts
```diff
@@ -38,3 +39,13 @@ export function isCustomDs<F extends SubstrateNetworkFilter>(
...
   return ds.kind === SubstrateDatasourceKind.Runtime;
 }
+
+export function isSubstrateTemplates(
+  templatesData: any,
+  specVersion: string
+): templatesData is (RuntimeDatasourceTemplate | CustomDatasourceTemplate)[] {
+  return (
+    (isRuntimeDs(templatesData[0]) || isCustomDs(templatesData[0])) &&
+    (specVersion === '0.2.1' || specVersion === '1.0.0')
```
stwiname, line 49: Would we be better off using semver here with >= 0.2.1? And same for in terra

stwiname (approved):

PR #1010, opened 2022-05-11
fix undefined file under chainTypes

jiqiang90:


stwiname (approved):
Interesting that typescript doesn't pick up on this. Maybe we need to set stricter typescript config

PR #1011, opened 2022-05-11
[release] 1.0.0

jiqiang90:


stwiname (commented):
Can we hold off making Terra and contract processors v1? I think Terra is still not as stable and contract processors would imply a breaking change.

jiqiang90:
> Can we hold off making Terra and contract processors v1? I think Terra is still not as stable and contract processors would imply a breaking change.

stwiname (approved):

PR #1046, opened 2022-05-20
use dictionary SpecVersion map

jiqiang90:
TODO:

- [x] since we moved specVersion query out from dictionary query, we need to update specVersion map frequently when reach last runtime.
- [x] use prefetched runtime for block
- [ ] Need to consider if dictionary specVersion are behind

### DIFF
--- a/packages/node/src/indexer/dictionary.service.ts
+++ b/packages/node/src/indexer/dictionary.service.ts
@@ -17,6 +17,12 @@ import { getLogger } from '../utils/logger';
 import { profiler } from '../utils/profiler';
 import { getYargsOption } from '../yargs';
 
+export type SpecVersion = {
+  id: string;
+  start: number; //start with this block
+  end: number;
+};
+
 export type Dictionary = {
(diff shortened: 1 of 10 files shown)


packages/node/src/indexer/dictionary.service.ts
```diff
@@ -232,4 +238,61 @@ export class DictionaryService implements OnApplicationShutdown {
...
+        }
+      }
+      return Array.from(specVersionBlockHeightSet);
+    } catch (err) {
+      logger.warn(err, `failed to fetch specVersion result`);
+      return undefined;
+    }
+  }
+
+  private specVersionQuery(): GqlQuery {
```
stwiname, line 277: Do we need to worry about the fact that we limit queries to 100 results?

jiqiang90: For now I think we are safe, might remove this query limit just for specVersion in query service.

stwiname (approved):

PR #1058, opened 2022-05-23
Fix cli init duplicate chainId and genesisHash

jiqiang90:
Due to missing determine project version


packages/cli/src/commands/init.ts
```diff
@@ -224,6 +223,9 @@ export default class Init extends Command {
     //TODO, only substrate family project should fetch its genesis hash, keep it this way for now
     if (this.networkFamily === NETWORK_FAMILIES.Substrate && flags.specVersion !== '0.0.1') {
       cli.action.start('Fetching network genesis hash');
+      if (gte(flags.specVersion, '1.0.0')) {
+        (this.project as ProjectSpecV1_0_0).chainId = await getGenesisHash(this.project.endpoint);
+      }
```
stwiname, line 228: Should below be in `else` block?

stwiname (approved):

PR #1066, opened 2022-05-25
Add performance logger and use to keep track of processing a block

stwiname:
Adds a new option `perf` which is a number to set the amount of performance logging. Default is nothing

Example output:
```
Block 501: 6979 ms
   getBlockHeight(): 1012 ms
     getBlockHash(): 200 ms
     getBlockByHash(): 812 ms
   events/specVersion: 1356 ms
   wrapBlock: 1 ms
   in queue: 3696 ms
   indexBlock(): 445 ms
     getPatchedApi: 6 ms
     indexBlockData(): 425 ms
       block: 0 ms
       initialization events: 0 ms
       extrinsics and events: 424 ms
       finalization events: 0 ms
     poi/mmr: 0 ms
     commitTx: 6 ms
```

### DIFF
--- a/packages/node/src/indexer/indexer.manager.ts
+++ b/packages/node/src/indexer/indexer.manager.ts
@@ -29,7 +29,7 @@ import { QueryTypes, Sequelize } from 'sequelize';
 import { NodeConfig } from '../configure/NodeConfig';
 import { SubqlProjectDs, SubqueryProject } from '../configure/SubqueryProject';
 import { SubqueryRepo } from '../entities';
-import { getLogger } from '../utils/logger';
+import { getLogger, perfLogger, PerfScope } from '../utils/logger';
 import { profiler } from '../utils/profiler';
 import * as SubstrateUtil from '../utils/substrate';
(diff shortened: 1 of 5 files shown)


packages/node/src/indexer/indexer.manager.ts
```diff
@@ -493,15 +505,23 @@ export class IndexerManager {
     { block, events, extrinsics }: BlockContent,
     dataSources: SubqlProjectDs[],
     getVM: (d: SubqlProjectDs) => IndexerSandbox,
+    perfScope: PerfScope,
```
ianhe8x, line 508: this is a little invasive

ianhe8x:
I always wanted to find a lib in node that can generate Histogram.

stwiname:
> I always wanted to find a lib in node that can generate Histogram.

I've been using https://clinicjs.org/documentation/flame/03-first-analysis/ if thats what you're after

stwiname:
Closing because we just need this for demo purposes and its not considered production ready

PR #1075, opened 2022-05-29
Fix the sequelize version

jiqiang90:
Temp fix the sequelize replacement issue


stwiname (approved):
Can we refer to a sequelize issue or change to keep track of this if we want to upgrade in the future

jiqiang90:
wait https://github.com/sequelize/sequelize/issues/14568 to be resolved

PR #1079, opened 2022-05-30
Disable historical queries by default

stwiname:


jiqiang90 (approved):

jiqiang90 (commented):
Can we use `default: true`

jiqiang90 (approved):

PR #1083, opened 2022-05-31
draft PR Fix benchmark logging

bz888:


packages/node/src/indexer/benchmark.service.spec.ts
```diff
@@ -0,0 +1,61 @@
...
+import { ApiService } from './api.service';
+import { BenchmarkService } from './benchmark.service';
+
+jest.setTimeout(90000);
+describe('Benchmark service', () => {
+  let app: INestApplication;
+
+  afterEach(async () => {
+    return app?.close();
+  });
```
stwiname, line 20: Seems this is never defined.

packages/node/src/indexer/benchmark.service.spec.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  const newbenchmarkService = new BenchmarkService();
+
+  (newbenchmarkService as any).currentProcessingHeight = 1208163;
+  (newbenchmarkService as any).targetHeight = 1208163;
+  (newbenchmarkService as any).lastRegisteredHeight = 1208163;
+  (newbenchmarkService as any).lastRegisteredTimestamp = 10000;
+  (newbenchmarkService as any).currentProcessingTimestamp = 10000;
```
stwiname, line 28: This should be in a before each.

Also generally when we write tests they are setup so they are independent of one another and can run the single test.

jiqiang90:
Looks good to me.

stwiname (approved):

PR #1091, opened 2022-06-02
handle multi ds with different start block height

### LINKED ISSUE #1089: Multiple datasources can be filtered out and never run
If there is more than 1 data source and it has a start block that is greater than the lowest start height it will be filtered out and not run unless the indexer is restarted after the start height of that datasource.

Example project: https://github.com/stwiname/demo-filter-ds-bug

Solution:

Rather than persisting the filtered data sources [here](https://github.com/subquery/subql/blob/main/packages/node/src/indexer/indexer.manager.ts#L233) we should filter for each block. I think this might also impact dictionary queries.

jiqiang90:
Closes #1089

### DIFF
--- a/packages/node/src/indexer/indexer.manager.ts
+++ b/packages/node/src/indexer/indexer.manager.ts
@@ -106,6 +106,10 @@ export class IndexerManager {
         isUpgraded ? block.block.header.parentHash : undefined,
       );
 
+      this.filteredDataSources = this.filterDataSources(
+        block.block.header.number.toNumber(),
+      );
+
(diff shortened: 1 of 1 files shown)


packages/node/src/indexer/indexer.manager.ts
```diff
@@ -481,6 +477,7 @@ export class IndexerManager {
     }
   }
 
+  //Deprecate method, filter only apply in manifest spec 0.0.1
```
stwiname, line 480: Can we just remove this function if it's no longer used?

stwiname (approved):

PR #1101, opened 2022-06-09
 both packages working

roduquen:


stwiname (commented):
This is looking good. I have left a couple of comments.

My other questions are:
- Whats going on with the minor formatting changes? Mostly just whitespace for renamed/moved files
- The types package should be for each chain, we don't want avalanche to have to install polkadot dependencies and vice versa.

stwiname:
Replaced by https://github.com/subquery/subql/pull/1167

Issue #713, opened 2021-12-20
Use nodejs worker for block fetching


ianhe8x:
to solve the process hang issue when processing a large ws payload

wqsz7xn:
I've been working on this a little bit today. It's fairly simple and just means that you have to offload [`fetchBlockBatches`](https://github.com/wqsz7xn/subql/blob/e6d57ae1212e08a764b645ae60c1c8c43b4a2cd5/packages/node/src/utils/substrate.ts#L348) to a nodejs worker. The only problem that i'm experiencing is that you have to pass the `api`, `block numbers` and `overallspecVersion` to the worker and doing that is difficult without some amount of serialisation (`worker.postMessage` only accepts some primitives/is limited). This would probably have a hit on performance.

PR #1103, opened 2022-06-10
Indexer workers

stwiname:
Usage:

`--worker=<number>` flag to specify number of workers. If this flag is not provided then workers will not be used and everything will run in the same thread.

Fixes https://github.com/subquery/subql/issues/713

TODO:

- [x] Get tests working
- [ ] ~~Scale batch size based on free memory~~ Not used and non-trivial to implement
- [x] Fix file based MMR not working
- [x] Apply https://github.com/subquery/subql/pull/1140 to block dispatcher implementations
- [x] Clean up
- [ ] Fix MMR failure with worker threads enabled


ianhe8x:
`ERROR Unable to get dynamic datasources SequelizeDatabaseError: relation "_metadata" does not exist`

packages/node/src/indexer/worker/block-dispatcher.service.ts
```diff
@@ -0,0 +1,383 @@
...
+
+    assert(worker, `Worker ${workerIdx} not found`);
+
+    // Used to compare before and after as a way to check if queue was flushed
+    const bufferedHeight = this._latestBufferedHeight;
+    const pendingBlock = worker.fetchBlock(height);
+
+    const processBlock = async () => {
+      const start = new Date();
+      const result = await pendingBlock;
```
ianhe8x, line 307: If i'm not wrong, `processBlock` live in main thread, so in workthread `worker.fetchBlock()` then transfer the result to main thread. Seems something we should avoid.

packages/node/src/indexer/worker/block-dispatcher.service.ts
```diff
@@ -0,0 +1,383 @@
...
+
+  await indexerWorker.initWorker();
+
+  return indexerWorker;
+}
+
+type GetRuntimeVersion = (block: SubstrateBlock) => Promise<RuntimeVersion>;
+
+function getMaxWorkers(numWorkers?: number): number {
+  const maxCPUs = os.cpus().length - 1;
```
ianhe8x, line 66:
```suggestion
  const maxCPUs = os.cpus().length;
```
Main thread doesn't do a lot things. It doesn't make sense if I need 3 cpu so I can have two workers.

ianhe8x:
`Error: Queue exceeds max size` sometimes

stwiname:
I can't find a reasonable way to scale the batch size based on memory, we don't use the feature anyway.

packages/node/src/indexer/worker/block-dispatcher.service.ts
```diff
@@ -0,0 +1,389 @@
...
+  private _latestBufferedHeight: number;
+
+  constructor(
+    private apiService: ApiService,
+    private nodeConfig: NodeConfig,
+    private indexerManager: IndexerManager,
+    private eventEmitter: EventEmitter2,
+  ) {
+    this.fetchQueue = new Queue(nodeConfig.batchSize * 3);
+    this.processQueue = new AutoQueue(nodeConfig.batchSize * 3);
```
ianhe8x, line 113: cucurrency is 1 then, I suppose
```suggestion
    this.processQueue = new AutoQueue(nodeConfig.batchSize * 3, nodeConfig.batchSize);
```

ianhe8x: if batchSize is overall batchSize, then concurrency will become `nodeConfig.batchSize / workers`

stwiname:
MMR is not currently working with threads: failed to sync poi to mmr Error: Poi block height 1463, Poi mmr 0x3ac9ff27f72360faffb6c63fd64c96d82233bb6811dde95428f91fc5f350fafc not same as filebased mmr: 0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470

jiqiang90 (approved):

PR #1104, opened 2022-06-10
WIP, index worker dynamic ds dictionary

jiqiang90:


jiqiang90:
Assume user expect the mapping will hit on block height 60, 70, 80, 100, 115, 140, 150... but this is unknown until dictionary is updated


- before update dictionary entry 

fetched block queue:  [50, **60** ,**70**, 75, **80** ]
block number queue:  [90, **100**, 110, **115**, 200]

- After we reset the block number queue and update the dictionary, dictionary will query start from block number 90 to ensure cover dropped blocks 

we should have 

fetched block queue:  [50, **60** ,**70**, 75, **80** ]
block number queue:  [**100, 115, 140, 150**]

________________________________________________________
Some doubts here, it is possible to drop the fetched block queue too, and use dictionary query from block 50,  but it will need to fetch these blocks has already been cached. 

- [ ] We can keep it original cached blocks in memory and use new block number queue to find their intersection, so they don't need to be fetched again.

PR #1119, opened 2022-06-22
handle when templates undefined

jiqiang90:


packages/node/src/configure/SubqueryProject.ts
```diff
@@ -175,9 +175,9 @@ async function loadProjectFromManifest0_2_1(
     path,
     networkOverrides,
   );
-
-  project.templates = await loadProjectTemplates(projectManifest, reader);
-
+  if (projectManifest.templates && projectManifest.templates.length !== 0) {
```
stwiname, line 178: Can we move this check into `loadProjectTemplates`

stwiname (approved):

PR #1133, opened 2022-06-27
fix set block offset

jiqiang90:


packages/node/src/indexer/project.service.ts
```diff
@@ -313,9 +318,10 @@ export class ProjectService {
     return startHeight;
   }
 
-  // FIXME Dedupe with indexermanager
+  // FIXME Dedupe with indexer manager
```
stwiname, line 321: I had event left a comment!

stwiname (approved):

packages/node/src/indexer/project.service.ts
jiqiang90: : D

PR #1152, opened 2022-07-01
Validate dictionary

bz888:


packages/node/src/indexer/fetch.service.ts
```diff
@@ -325,8 +321,13 @@ export class FetchService implements OnApplicationShutdown {
     await this.getFinalizedBlockHead();
     await this.getBestBlockHead();
 
-    if (this.useDictionary) {
-      const specVersionResponse = await this.dictionaryService.getSpecVersion();
+    const specVersionResult = await this.dictionaryService.getSpecVersion();
```
stwiname, line 324: We should check `this.useDictionary`

packages/node/src/indexer/fetch.service.ts
```diff
@@ -643,26 +644,34 @@ export class FetchService implements OnApplicationShutdown {
   }
 
   private dictionaryValidation(
-    { _metadata: metaData }: Dictionary,
-    startBlockHeight: number,
+    dictionary: Dictionary | SpecVersionDictionary,
```
stwiname, line 647: We can use the minimal data type here, means its easier to determine what data is used and easier to write tests
```suggestion
    { _metadata}: { _metadata: MetaData },
```

packages/node/src/indexer/dictionary.service.ts
```diff
@@ -239,34 +245,48 @@ export class DictionaryService implements OnApplicationShutdown {
     return buildQuery(vars, nodes);
   }
 
-  async getSpecVersion(): Promise<SpecVersion[]> {
+  async getSpecVersionMap(
```
stwiname, line 248: I find this function rather confusing. The naming doesn't make sense when comparing to `getSpecVersion` and functions should do only one thing where possible.

How about the interface below, then its easier to infer what the functions do

```ts

getSpecVersions(): SpecVersion[];
getSpecVersionsRaw(): SpecVersonDictionary);
parseSpecVersions(raw: SpecVersionDictionary): SpecVersion[];
```

stwiname (approved):

PR #1153, opened 2022-07-01
Try to catch error if create server failed, also give retry build schema

jiqiang90:


packages/query/src/graphql/graphql.module.ts
```diff
@@ -43,13 +45,36 @@ export class GraphqlModule implements OnModuleInit, OnModuleDestroy {
...
+      try {
+        const builder = await getPostGraphileBuilder(this.pgPool, [dbSchema], options);
+        const graphqlSchema = builder.buildSchema();
+        return graphqlSchema;
+      } catch (e) {
+        await delay(SCHEMA_RETRY_INTERVAL);
+        return this.buildSchema(dbSchema, options, --retries);
+      }
+    } else {
+      console.error(`Failed to build schema ${dbSchema} ${SCHEMA_RETRY_NUMBER} times`);
```
stwiname, line 74: Should be throw here?

stwiname (approved):

PR #1164, opened 2022-07-05
skip worker who is currently indexing

ianhe8x:


ianhe8x:
Maybe this doesn't bring a lot improvements. The block fetching tasks are all dispatched to worker long before worker starts indexing any blocks.

stwiname:
This seems to make indexing much slower for me

PR #1165, opened 2022-07-05
Add cosmos family to cli init

stwiname:


jiqiang90 (commented):
We should tidy this up before, import it from https://github.com/subquery/subql/blob/61987ade6232a86f7fbaf4f765cc23c9e5e8cd2f/packages/common/src/constants.ts#L20

jiqiang90 (approved):

PR #1177, opened 2022-07-09
add dictionary query timeout option

jiqiang90:
# Description

While query service it self has `query-timeout`, (on host service this is set to 60 seconds) but sometime long waiting for dictionary query is not the idea. We should provide an option allow indexer to decide how they want to dictionary act. This can provide a shorter timeout for dictionary query, like 20 seconds. So if query has timeout, it will back to use original batch (range)

Fixes # (issue)

## Type of change

## Checklist

- [x] I have tested locally

### DIFF
--- a/packages/node/src/indexer/dictionary.service.ts
+++ b/packages/node/src/indexer/dictionary.service.ts
@@ -15,6 +15,7 @@ import fetch from 'node-fetch';
 import { SubqueryProject } from '../configure/SubqueryProject';
 import { getLogger } from '../utils/logger';
 import { profiler } from '../utils/profiler';
+import { timeout } from '../utils/promise';
 import { getYargsOption } from '../yargs';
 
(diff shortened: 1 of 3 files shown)


jiqiang90:
image

packages/node/src/indexer/dictionary.service.ts
```diff
@@ -155,10 +156,13 @@ export class DictionaryService implements OnApplicationShutdown {
...
-      const resp = await this.client.query({
-        query: gql(query),
-        variables,
-      });
+      const resp = await timeout(
+        this.client.query({
+          query: gql(query),
+          variables,
+        }),
+        argv['dictionary-timeout'],
```
stwiname, line 164: Whats the reasoning here between using `argv` and `NodeConfig`?

stwiname (approved):

PR #1193, opened 2022-07-17
Validate blockHeight argument

### LINKED ISSUE #1124: Pass future block height in historical state query
Precondition
* historical enabled
* use blockHeight in the query

When specify a future blockHeight (comparing with lastProcessedHeight)
Currently it returns results same as `blockHeight` is not specified, this might lead to some confusion. Maybe a better way is throw error in this case.

spitko:
Fixes #1124 

## Type of change

## Checklist

- [x] I have tested locally

### DIFF
--- a/packages/query/src/graphql/graphql.module.ts
+++ b/packages/query/src/graphql/graphql.module.ts
@@ -21,6 +21,7 @@ import {Config} from '../configure';
 import {PinoConfig} from '../utils/logger';
 import {getYargsOption} from '../yargs';
 import {plugins} from './plugins';
+import {BlockHeightArgValidationPlugin} from './plugins/BlockHeightArgValidationPlugin';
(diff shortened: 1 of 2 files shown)


jiqiang90:
Hi @spitko I wonder if there is a better approach for this, current approach will need run additional query for to get `lastProcessedHeight` in order to make comparison, this means we will have extra consumption for each query. 

Is that possible to merge the comparison query and original query into one? Thanks

marilene38:
> Hi @spitko I wonder if there is a better approach for this, current approach will need run additional query for to get `lastProcessedHeight` in order to make comparison, this means we will have extra consumption for each query. 
> 
> 
> 
> Is that possible to merge the comparison query and original query into one? Thanks
> 
> 
> 
>

PR #1196, opened 2022-07-19
add block modulo filter

DeveloperInProgress:


packages/node/src/indexer/fetch.service.ts
```diff
@@ -429,6 +448,13 @@ export class FetchService implements OnApplicationShutdown {
       }
       if (this.useDictionary) {
         const queryEndBlock = startBlockHeight + DICTIONARY_MAX_QUERY_SIZE;
+        const modulos = this.getModulos();
+        const moduloBlocks: number[] = [];
+        for (let i = startBlockHeight; i < queryEndBlock; i++) {
+          if (modulos.find((m) => i % m === 0)) {
+            moduloBlocks.push(i);
+          }
+        }
```
stwiname, line 457: Can we please put this into a function

jiqiang90 (approved):

PR #1199, opened 2022-07-20
Fix wrong default endpoint for cli deploy command

fewensa:
# Description
The ipfs validate api returned data not have a correct endpoint, there you set chainId to the endpoint. this is wrong.

I think if not have endpoint we don't need to set it. because the indexer will be read from project.yaml.

whatever the chainId is not endpoint.

Fixes # (issue)

## Type of change

## Checklist

- [ ] I have tested locally

### DIFF
--- a/packages/cli/src/commands/deployment/deploy.ts
+++ b/packages/cli/src/commands/deployment/deploy.ts
@@ -60,7 +60,7 @@ export default class Deploy extends Command {
       if (!flags.useDefaults && !validateEndpoint) {
         endpoint = await promptWithDefaultValues(cli, 'Enter endpoint', validateEndpoint, null, true);
       } else {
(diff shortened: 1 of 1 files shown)


bz888:
Endpoint is one of the required fields for deploying to hosted services, and if the endpoint provided in the project.yaml is not apart of the list of endpoints we have. then `validateEndpoint` would return `undefined` hence failing the deployment, and in the case of `undefined ` it should then use the endpoint provided by the project.yaml

fewensa:
> 

Okay, so how to read project.yaml there, I can not found a example this file. (Initially I also wanted to be able to read in this file

bz888:
> > 
> 
> Okay, so how to read project.yaml there, I can not found a example this file. (Initially I also wanted to be able to read in this file

Hey, I have made some changes in-accordance with your suggestion, however, i have made a slight change just to better align with our pipeline.

PR #1226, opened 2022-08-02
schema migration

jiqiang90:
# Description

Current implementation we only allow user to add new entity in schema while project already in syncing, and modification to existing entity will be ignored. This change will allow user to : 

- Add new field to entity
- Drop current filed on entity  (this will also delete corresponding column in table and data)
- Change current field attributes (This will be limited from future)

## Type of change

## Checklist

- [x] I have tested locally

### DIFF
--- a/packages/node/src/indexer/store.service.ts
+++ b/packages/node/src/indexer/store.service.ts
@@ -303,7 +303,9 @@ export class StoreService {
     }
     this.metaDataRepo = MetadataFactory(this.sequelize, schema);
 
-    await this.sequelize.sync();
+    // this will allow alter current entity, including fields
+    // TODO, add rules for changes, eg only allow add nullable field
+    await this.sequelize.sync({ alter: { drop: true } });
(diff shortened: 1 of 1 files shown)


jamesbayly:
Jesus, what other one line feature requests can we make?

jamesbayly:
Jesus, what other one line feature requests can we make?

packages/node/src/indexer/store.service.ts
```diff
@@ -303,7 +303,9 @@ export class StoreService {
     }
     this.metaDataRepo = MetadataFactory(this.sequelize, schema);
 
-    await this.sequelize.sync();
+    // this will allow alter current entity, including fields
+    // TODO, add rules for changes, eg only allow add nullable field
```
stwiname, line 307: Is this going to happen in another PR?

stwiname (commented):
How does a user use this feature? They're going to want it to happen at a certain block right?

stwiname (approved):

Issue #1233, opened 2022-08-04
Improve performance logging


stwiname:
Things to track:

- [x] Add handlers to profiler to measure how long handlers take to run
- [ ] Keep track of processed BPS, this is to give a more accurate reading when using a dictionary
- [ ] Longer term BPS measurement. Currently we only measure BPS in 15 second windows

PR #1244, opened 2022-08-11
Improve Performance logging

bz888:
# Description

Fixes # (#1233 )

## Type of change

## Checklist

- [x] I have tested locally

### DIFF
--- a/packages/node/src/indexer/store.service.ts
+++ b/packages/node/src/indexer/store.service.ts
@@ -110,6 +110,7 @@ export class StoreService {
       logger.error(e, `Having a problem when get indexed fields`);
       process.exit(1);
     }
+    await this.setMetadata('processedBlockHeight', 0);
   }
 
   // eslint-disable-next-line complexity
@@ -616,6 +617,7 @@ group by
         );
       }
     }
+
     await this.setMetadata('lastProcessedHeight', targetBlockHeight, {
       transaction,
     });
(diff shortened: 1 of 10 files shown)


packages/node/src/indexer/store.service.ts
```diff
@@ -110,6 +110,12 @@ export class StoreService {
     }
   }
 
+  async incrementBlockCount(key: string): Promise<void> {
```
stwiname, line 113: This should us a db transaction and I think the key can be hard coded

packages/node/src/indexer/worker/block-dispatcher.service.ts
```diff
@@ -212,17 +212,20 @@ export class BlockDispatcherService
         blockNums,
       );
 
+      const processedBlockCount =
```
stwiname, line 215: Getting this every time doesn't make much sense. I think we can keep track of it in memory.

jiqiang90: Agree, we should only get this once when project initialized

stwiname:
I think we should log the processed bps along with he current bps

packages/node/src/indexer/store.service.ts
```diff
@@ -110,6 +110,12 @@ export class StoreService {
     }
   }
 
+  async incrementBlockCount(): Promise<void> {
```
stwiname, line 113: This should use the db transaction

stwiname (approved):

jiqiang90 (approved):

PR #1246, opened 2022-08-12
support bulkGet and add bulkUpdate 

### LINKED ISSUE #1237: Add bulkUpdate and bulkGet to injected sandbox store
## Description

In order for users to further optimise their projects, if they perform multiple updates within a handler they need to get multiple entities before updating them.

We can provide the ability to do this in bulk. Much like with bulkCreate

jiqiang90:
# Description

Fixes #1237 

**bulkGet:** 

We currently already support bulkGet method , you can use  `getByField` and pass 3rd parameter as an array.
```
const starterEntites = await store.getByField('StarterEntity','field1',[50,100,150])
```
This will returns entities `field1` equal to 50, 100 and 150.

**bulkCreate:** 

Add bulkUpdate method, allow user to update multiple entities in single call. 
```
starterEntites[0]['field2']= 'Apple'
starterEntites[1]['field2']= 'Lemon'
starterEntites[2]['field2']= 'Orange
await store.bulkUpdate('StarterEntity',starterEntites)
```
This will update all starterEntities and its fields with updated field2.

You can also specify which filed you want to update, by :

```
// If field5 has changed
starterEntites[0]['field5']= false

// This won't update field5, only will update field2. 
await store.bulkUpdate('StarterEntity',starterEntites,['field2'])
```

Follow up: **BulkUpdate with Historical is supported** 

## Type of change

## Checklist

- [x] I have tested locally

### DIFF
--- a/packages/node/src/indexer/store.service.ts
+++ b/packages/node/src/indexer/store.service.ts
@@ -634,32 +634,6 @@ group by
     }
   }
 
-  private async updateHistorical(
-    model: ModelCtor<Model>,
-    data: Entity,
-    attributes: CreationAttributes<Model>,
-  ) {
-    const [updatedRows] = await model.update(attributes, {
-      hooks: false,
-      transaction: this.tx,
-      where: this.sequelize.and(
-        { id: data.id },
-        this.sequelize.where(
-          this.sequelize.fn('lower', this.sequelize.col('_block_range')),
-          this.blockHeight,
-        ),
-      ),
-    });
-    if (updatedRows < 1) {
-      // TODO, can be improved here if called from BulkUpdate
-      // We can collect entities meet this conditions, use promise.all and bulkCreate here
-      await this.markAsDeleted(model, data.id);
-      await model.create(attributes, {
-        transaction: this.tx,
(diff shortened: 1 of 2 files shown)


packages/node/src/indexer/store.service.ts
```diff
@@ -767,6 +768,37 @@ group by
...
       },
+
+      bulkUpdate: async (
+        entity: string,
+        data: Entity[],
+        fields?: string[],
+      ): Promise<void> => {
+        try {
+          const model = this.sequelize.model(entity);
+          const modelFields =
```
stwiname, line 779: I think this needs to be below the `assert`

stwiname (commented):
Did we want to throw an error if `bulkUpdate` is called with historical for now?

packages/node/src/indexer/store.service.ts
jiqiang90: > Did we want to throw an error if `bulkUpdate` is called with historical for now?

Sure, will have a follow up pr for historical

stwiname (approved):

Issue #1198, opened 2022-07-20
[Feature] Add unique with aggregation to query service


jiqiang90:
### Description

We want to support query able to get  aggregated and uniqued result. 

For example we want to query a dictionary and return the first 100 blocks that contain the same events,  

**Current behaviour:** 

Only allow us find the first 100 matched events, and return all the blocks height, due to some block has the multiple matched events, this will return same block heights in multiple records, the result is not aggregated and uniqued.
![Screen Shot 2022-07-20 at 1 13 02 PM](https://user-images.githubusercontent.com/10431657/179876119-a42ef2ef-5416-43f3-bed3-757011eaf376.png)

**Expect behaviour:** 

we want to return 100 records with unique block heights.

Issue #770, opened 2022-01-25
Improve syntax of groupby in subql/query


lamcc21:
At the moment we are using group-by to simulate distinct in subql query. 
Could improve the graphql so that it is more readable, by creating a seperate plugin.

[docs of plugin we are using for group-by](https://github.com/graphile/pg-aggregates)

PR #1274, opened 2022-08-29
Add distinct query plugin

stwiname:
# Description

Depends on https://github.com/graphile/graphile-engine/pull/805

Fixes https://github.com/subquery/subql/issues/1198, Fixes https://github.com/subquery/subql/issues/770

## Type of change

## Checklist

- [x] I have tested locally

### DIFF
--- a/packages/node-core/src/indexer/dictionary.service.ts
+++ b/packages/node-core/src/indexer/dictionary.service.ts
@@ -27,6 +27,8 @@ export type SpecVersionDictionary = {
 
 const logger = getLogger('dictionary');
 
+const distinctErrorEscaped = `Unknown argument \\"distinct\\"`;
+
 function extractVar(name: string, cond: DictionaryQueryCondition): GqlVar {
   let gqlType: string;
   switch (typeof cond.value) {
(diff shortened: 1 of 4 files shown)


jiqiang90 (approved):

packages/node-core/src/indexer/dictionary.service.ts
```diff
@@ -120,6 +123,11 @@ function buildDictQueryFragment(
       first: batchSize.toString(),
     },
   };
+
+  if (useDistinct) {
+    node.args.distinct = ['BLOCK_HEIGHT'];
```
jiqiang90, line 128: I think if we useDistinct, we no longer need to have 
```
lessThan: `"${queryEndBlock}"`,
```

if the distinct blocks in between difference is large than queryEndBlock threshold 10000, this will become inefficient. 

For example expected distinct return 1999, 8999, 28999... (total number of batch size blocks)
but due to query limit 

        blockHeight: {
          greaterThanOrEqualTo: `1999`,
          lessThan: `11999`,
        },

we will only have 2 blocks in return

stwiname: I think we need to retain `lessThan` for the work that @bz888 has done. But thinking about that, the `queryEndBlock` could be set to the height when a next dataSource is introduced. 

Maybe we should save this for another task?

jiqiang90: Ok, let's do it in another task

stwiname: https://github.com/subquery/subql/issues/1381

PR #1280, opened 2022-09-03
update polkadot and types, to support with wasm

jiqiang90:
# Description

Update types in order to work with wasm processor.

- starter project: https://github.com/subquery/substrate-wasm-starter
- wasm processor:  https://github.com/subquery/datasource-processors/pull/21

## Type of change

## Checklist

- [ ] I have tested locally

### DIFF
--- a/packages/node/src/indexer/indexer.manager.ts
+++ b/packages/node/src/indexer/indexer.manager.ts
@@ -72,7 +72,6 @@ export class IndexerManager {
   ) {
     logger.info('indexer manager start');
     this.api = this.apiService.getApi();
-    this.dsAssets = {};
   }
 
   @profiler(argv.profiler)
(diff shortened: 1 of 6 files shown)


packages/node/src/indexer/indexer.manager.ts
```diff
@@ -271,6 +272,16 @@ export class IndexerManager {
     }
   }
 
+  //Avoid repeat read same asset
```
stwiname, line 275: Could this be moved to the ds processor service?

packages/node/src/indexer/indexer.manager.ts
```diff
@@ -352,12 +364,12 @@ export class IndexerManager {
         const processor = asSecondLayerHandlerProcessor_1_0_0(
           plugin.handlerProcessors[handler.kind],
         );
-
         try {
           return processor.filterProcessor({
             filter: handler.filter,
             input: data,
             ds,
+            assets,
```
stwiname, line 372: I'm guessing we need to decode the input in order to filter. Is that going to slow things down as it will be decoded twice now (filter and transform)? 

Ideally we don't need to decode with the filter bit if we do can we somehow share the decoded data? Maybe keeping it in memory in the processor

stwiname (approved):

PR #1296, opened 2022-09-15
Set useragent headers on polkadot api

stwiname:
Sets the useragent header to identify requests coming from subery


jiqiang90 (approved):

jamesbayly:
Note many services will look at origin rather than user agent
I suggest setting origin to the project name

PR #1297, opened 2022-09-15
Fix tests

bz888:
# Description

Fixes # (issue)

## Type of change

Fixed CLI publish tests

## Checklist

- [x] I have tested locally


packages/cli/src/controller/publish-controller.spec.ts
```diff
@@ -139,7 +140,7 @@ describe('Cli publish', () => {
     await expect(uploadToIpfs('', ipfsEndpoint, projectDir)).rejects.toBeDefined();
   });
 
-  it('throw error when v0.0.1 try to deploy', async () => {
+  it.skip('throw error when v0.0.1 try to deploy', async () => {
```
stwiname, line 143: Why is this being skipped now? Would be good to have a comment if we can enable later on or should we just remove the test?

stwiname (approved):

PR #1300, opened 2022-09-19
Fix exporting logger

stwiname:


packages/node-core/src/indexer/benchmark.service.ts
```diff
@@ -3,9 +3,9 @@
 
 import {OnEvent} from '@nestjs/event-emitter';
 import {Interval} from '@nestjs/schedule';
-import {NodeConfig} from '@subql/node-core/configure';
```
stwiname, line 6: Node core cannot import from node core. This will fail when published. Can you please check for other imports in the PR

Issue #1144, opened 2022-06-29
Index unfinalized blocks


stwiname:
Now that historical data is supported we can have the ability to index blocks that are unfinalized, and blocks that don't finalize can be rolled back so that we keep correct data.

A use case of this is that it reduces the delay between interactions with the chain and data from subquery

Todo:
- [ ] use flag `--unfinalized` to enable the feature
- [ ] once a project is upto the latest finalized block, start indexing unfinalized blocks.
- [ ] Roll back any indexing of unfinalized blocks that are dropped and unfinalized
- [ ] Add the ability to filter by finalized data, this would effectively result in the current behaviour.

PR #1308, opened 2022-09-20
Best Block

jiqiang90:
Enable best block feature by `--best-block`

Fixes https://github.com/subquery/subql/issues/1144

Todo:

- [x] Need consider if program restart, memory best block lost and unfinalized block will be missed checking, need to also update this to metadata table and reuse it when restart
- [x] instead update best block by periodically, use point fetching to best all the time, and if this block been processed write it memory. 
- [x] enable with a flag
- [ ] multi workers


packages/node/src/indexer/project.service.ts
```diff
@@ -388,7 +392,7 @@ export class ProjectService {
     );
   }
 
-  private async reindex(targetBlockHeight: number): Promise<void> {
+  async reindex(targetBlockHeight: number): Promise<void> {
```
stwiname, line 395: The subcommands PR has moved this to another location. Will require rebasing.

stwiname (commented):
This looks really good

stwiname:
I think we should have the ability to disable this with a flag

packages/node/src/indexer/project.service.ts
jiqiang90: Sure, once reindex pr is merged, I will have this updated

packages/node/src/indexer/fetch.service.ts
```diff
@@ -291,13 +297,16 @@ export class FetchService implements OnApplicationShutdown {
...
       const finalizedBlock = await this.api.rpc.chain.getBlock(finalizedHead);
+      this.bestBlockService.registerFinalizedBlock(finalizedBlock);
       const currentFinalizedHeight =
         finalizedBlock.block.header.number.toNumber();
       if (this.latestFinalizedHeight !== currentFinalizedHeight) {
         this.latestFinalizedHeight = currentFinalizedHeight;
-        this.eventEmitter.emit(IndexerEvent.BlockTarget, {
-          height: this.latestFinalizedHeight,
-        });
+        if (!argv['best-block']) {
```
stwiname, line 305: Can we please get access to this via NodeConfig instead of args

packages/node/src/indexer/project.service.ts
```diff
@@ -105,6 +122,44 @@ export class ProjectService {
...
+        } else {
+          // if finalized block haven't been process yet, then remove best blocks from both metadata and memory
+          const transaction = await this.sequelize.transaction();
+          await this.storeService.resetBestBlocks(transaction);
+          await this.storeService.resetLastFinalizedVerifiedHeight(transaction);
+          await transaction.commit();
+          this.bestBlockService.resetBestBlocks();
+        }
+      }
+    }
```
stwiname, line 162: It would be nice if we could move as much as possible into the best block service.

packages/node/src/indexer/project.service.ts
```diff
@@ -318,4 +402,38 @@ export class ProjectService {
           this.apiService.getApi().runtimeVersion.specName.toString(),
     );
   }
+
+  async reindex(targetBlockHeight: number): Promise<void> {
```
stwiname, line 406: Can we use the function in reindex service instead? This mostly duplicates code

stwiname (commented):
* Historical indexing is required for this feature to work. It would be good to have an error message if the user tries to enable this without historical enabled on the db
* I find the name `bestBlock` can be ambiguious, maybe something like `unfinalizedBlocks` is a clearer name

packages/node/src/indexer/bestBlock.service.ts
```diff
@@ -0,0 +1,186 @@
...
+            `Block folk found, enqueued un-finalized block at ${lastVerifiableBlock.blockHeight} with hash ${lastVerifiableBlock.hash}, actual hash is ${actualHash} `,
+          );
+          return false;
+        }
+      }
+    }
+    return true;
+  }
+
+  async getLastCorrectBestBlock(): Promise<number | undefined> {
```
ianhe8x, line 168: Better to use the received finalized block's parent hash and iterate backwards. `getBlockHash` might not be reliable (the node is not catching finalized block or hit cache)

packages/node/src/indexer/unfinalizedBlocks.service.spec.ts
```diff
@@ -0,0 +1,194 @@
...
+  });
+
+  it('can validate best block if finalized block can NOT be found in its record', async () => {
+    unfinalizedBlocksService.storeUnfinalizedBlock(50, '0x50abcd');
+    unfinalizedBlocksService.storeUnfinalizedBlock(80, '0x80abcd');
+    unfinalizedBlocksService.storeUnfinalizedBlock(100, '0x100abc');
+    unfinalizedBlocksService.storeUnfinalizedBlock(120, '0x123456');
+    unfinalizedBlocksService.registerFinalizedBlock({
+      block: { header: { number: { toNumber: () => 150 } } },
+      hash: { toHex: () => '0x100abc' },
```
ianhe8x, line 159: finalized block is not likely having same hash as any unfinalized block.

packages/node/src/indexer/unfinalizedBlocks.service.ts
```diff
@@ -0,0 +1,219 @@
...
+      if (record) {
+        const [bestBlockHeight, hash] = record;
+        return { blockHeight: Number(bestBlockHeight), hash: hash };
+      }
+    }
+    return undefined;
+  }
+
+  // verify best blocks with finalized block
+  async validateUnfinalizedBlocks(): Promise<boolean> {
```
ianhe8x, line 146: Not sure if we need a seperate `validateUnfinalizedBlocks()`, can we do everything in one function instead? (find  fork, if yes, return fork height and purge unfinalized block array, if no, purge unfinalized block array to latest finalised height.

packages/node/src/indexer/unfinalizedBlocks.service.spec.ts
```diff
@@ -0,0 +1,194 @@
...
+    unfinalizedBlocksService.storeUnfinalizedBlock(120, '0x120abc');
+    unfinalizedBlocksService.registerFinalizedBlock({
+      block: { header: { number: { toNumber: () => 150 } } },
+      hash: { toHex: () => '0x100abc' },
+    } as unknown as SignedBlock);
+    // blockHash mocked value is `0x123456`
+    await expect(
+      unfinalizedBlocksService.validateUnfinalizedBlocks(),
+    ).resolves.toBeFalsy();
+    expect(unfinalizedBlocksService.unfinalizedBlock(120)).toBeUndefined();
```
ianhe8x, line 177: Which function removed block 120? I don't think it is `registerFinalizedBlock` or `validateUnfinalizedBlocks`

packages/node/src/indexer/unfinalizedBlocks.service.ts
```diff
@@ -0,0 +1,219 @@
...
+      if (hash === this.finalizedBlock.block.header.hash.toHex()) {
+        return Number(block);
+      }
+      // Should match for index 1....n
+      if (hash === checkingHeader.parentHash.toHex()) {
+        return Number(block);
+      }
+
+      // Get the new parent
+      checkingHeader = await this.api.rpc.chain.getHeader(
```
ianhe8x, line 194: How many times you need to search backwards isn't decided by length of bestVerifiableBlocks, but the distance of finalised block height and unfinalised block height.

packages/node/src/indexer/unfinalizedBlocks.service.spec.ts
```diff
@@ -0,0 +1,194 @@
...
+  });
+
+  it('get last correct best block', async () => {
```
ianhe8x, line 180: need another testcase that the received finalized block height doesn't have a matched unfinalised block in the stored array.

packages/node/src/indexer/unfinalizedBlocks.service.ts
```diff
@@ -0,0 +1,222 @@
...
+      return false;
+    }
+
+    // Unfinalized blocks beyond finalized block
+    if (lastVerifiableBlock.blockHeight === this.finalizedBlockNumber) {
+      return lastVerifiableBlock.hash !== this.finalizedBlock.hash.toHex();
+    } else {
+      // Unfinalized blocks below finalized block
+      const actualHash = (
+        await this.api.rpc.chain.getBlockHash(lastVerifiableBlock.blockHeight)
```
ianhe8x, line 150: Can't use `getBlockHash`, it is not reliable. Use finalisedBlock's parant hash and iterate backwards.

packages/node/src/indexer/unfinalizedBlocks.service.ts
```diff
@@ -0,0 +1,222 @@
...
+    );
+    if (record) {
+      const [bestBlockHeight, hash] = record;
+      return { blockHeight: Number(bestBlockHeight), hash };
+    }
+    return undefined;
+  }
+
+  // check unfinalized blocks for a fork
+  private async hasForked(): Promise<boolean> {
```
ianhe8x, line 134: I feel `hasForked()` and `getLastCorrectFinalizedBlock()` can be merged. Do both in once iteration

ianhe8x:
```suggestion
  private async hasForked(): Promise<BlockNumber> {
```

stwiname: I'm not sure we gain anything from this. Its easy to detect a fork because block hashes don't match, it takes more work to find at what block the fork occurred

packages/node/src/indexer/project.service.ts
```diff
@@ -111,6 +124,50 @@ export class ProjectService {
...
+      await this.getLastFinalizedVerifiedHeight();
+    if (this.nodeConfig.unfinalizedBlocks) {
+      this.unfinalizedBlockService.init(
+        this.metadataRepo,
+        startUnfinalizedBlocks ?? {},
+        lastFinalizedVerifiedHeight,
+      );
+    } else {
+      // Has previous indexed with unfinalized blocks, but it is no longer enabled
+      if (
```
ianhe8x, line 144: in case `Has previous indexed with unfinalized blocks, but it is no longer enabled`, I can think of two possibilities
We check the finality of these blocks, 
1. if they are no forks. then we remove them from unfinalised block array in metadata and continue index
2. if there are forks, we rewind to the height before the earlest block in them.

In this method, the `lastFinalizedVerifiedHeight` is read from metadata, not from latest network, which could be improved.

packages/node/src/indexer/unfinalizedBlocks.service.ts
```diff
@@ -0,0 +1,239 @@
...
+  private async getLastCorrectFinalizedBlock(): Promise<number | undefined> {
+    const bestVerifiableBlocks = this.getSortedUnfinalizedBlocks().filter(
+      ([bestBlockHeight, hash]) =>
+        Number(bestBlockHeight) <= this.finalizedBlockNumber,
+    );
+
+    let checkingHeader = this.finalizedHeader;
+
+    // Work backwards through the blocks until we find a matching hash
+    for (const [block, hash] of bestVerifiableBlocks.reverse()) {
```
ianhe8x, line 187: this is still somehow duplicated with `hasForked()`.
Also, as I pointed out, `bestVerifiableBlocks` is not continuous, so you can't call `checkingHeader` only once in a loop.

packages/node/src/utils/reindex.ts
```diff
@@ -0,0 +1,64 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+import { getLogger, MmrService, StoreService } from '@subql/node-core';
+import { Sequelize } from 'sequelize';
+import { UnfinalizedBlocksService } from '../indexer/unfinalizedBlocks.service';
+import { ForceCleanService } from '../subcommands/forceClean.service';
+
+const logger = getLogger('Reindex');
+
+export async function reindex(
```
ianhe8x, line 11: this is not a good util function, not a pure function but depends on other services.

packages/node/src/indexer/unfinalizedBlocks.service.ts
```diff
@@ -0,0 +1,239 @@
...
+
+  private get api(): ApiPromise {
+    return this.apiService.getApi();
+  }
+
+  private get finalizedBlockNumber(): number {
+    return this.finalizedHeader.number.toNumber();
+  }
+
+  private getSortedUnfinalizedBlocks(): [string, HexString][] {
```
ianhe8x, line 48: why don't use a sorted array to store unfinalised blocks?

ianhe8x (changes requested):

packages/node/src/indexer/unfinalizedBlocks.service.ts
ianhe8x: > I'm not sure we gain anything from this. Its easy to detect a fork because block hashes don't match, it takes more work to find at what block the fork occurred

But you still need to find the earlist forked block in the unfinalised block array, which is currently in `getLastCorrectFinalizedBlock` that I think can be merged to `hasForked()`

ianhe8x:
I think it's important that when we index unfinalised blocks, we don't skip blocks (no dictionary), and even we do nothing in a block, we added that block to metadata as unfalised block.

packages/node/src/indexer/unfinalizedBlocks.service.ts
stwiname: `bestVerifiableBlocks` should be continuous as we do it for every block we index and the dictionaries should not be indexing unfinalized block so it will be disabled

ianhe8x:
Sometime in the log `bps` line, target is higher than the one in metadata table.

ianhe8x (approved):

packages/node/src/indexer/unfinalizedBlocks.service.ts
```diff
@@ -0,0 +1,306 @@
...
+    for (const [block, hash] of bestVerifiableBlocks.reverse()) {
+      if (
+        hash === checkingHeader.hash.toHex() ||
+        hash === checkingHeader.parentHash.toHex()
+      ) {
+        return Number(block);
+      }
+
+      // Get the new parent
+      checkingHeader = await this.api.rpc.chain.getHeader(
```
ianhe8x, line 244: This can be moved to head of the loop, so if `bestVerifiableBlocks` has only 1 element, it can be skipped.

PR #1354, opened 2022-10-18
support version without v

### LINKED ISSUE #1344: Support `subql publish` more versions
### Description

The cli publish command requires providing versions for node and query. Currently we only support the semver (e.g. `1.0.0`) we should also support the version prefixed with a `v` like `v1.0.0`. We can remove the `v` in the cli.

bighammerw:
# Description
support more version for deployment command 

Fixes #1344 

## Type of change

## Checklist

- [ ] I have tested locally

### DIFF
--- a/packages/cli/src/utils.ts
+++ b/packages/cli/src/utils.ts
@@ -22,9 +22,8 @@ export async function valueOrPrompt<T>(value: T, msg: string, error: string): Pr
 }
 
 export function addV(str: string | undefined) {
-  if (!str?.includes('v')) {
-    const res = 'v';
-    return res.concat(str);
+  if (str && !str.includes('v')) {
+    return `v${str}`;
   }
   return str;
 }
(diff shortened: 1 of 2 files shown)


packages/cli/src/utils.ts
```diff
@@ -21,6 +21,14 @@ export async function valueOrPrompt<T>(value: T, msg: string, error: string): Pr
   }
 }
 
+export function addV(str: string | undefined) {
+  if (!str?.includes('v')) {
+    const res = 'v';
+    return res.concat(str);
```
stwiname, line 27: String interpolation can be used here
```suggestion
    return `v${str}`
```

packages/cli/src/utils.ts
```diff
@@ -21,6 +21,14 @@ export async function valueOrPrompt<T>(value: T, msg: string, error: string): Pr
   }
 }
 
+export function addV(str: string | undefined) {
+  if (!str?.includes('v')) {
```
stwiname, line 25: I think if `str` is undefined then this function will return "v"

stwiname (approved):

PR #1358, opened 2022-10-20
Optimise modulo filter

### LINKED ISSUE #1200: Optimise `modulo` block filter
If we only have block handlers and they have the `modulo` filter then we can calculate what blocks to fetch rather than fetching every block

bz888:
# Description

Fixes #1200 

## Checklist

- [x] I have tested locally

### DIFF
--- a/packages/node/src/indexer/fetch.service.ts
+++ b/packages/node/src/indexer/fetch.service.ts
@@ -363,19 +363,6 @@ export class FetchService implements OnApplicationShutdown {
     );
   }
 
-  checkHandlers(handlerType: SubstrateHandlerKind): number[] {
-    const checker = this.project.dataSources.map((ds) => {
-      const arr = [];
-      for (const handler of ds.mapping.handlers) {
-        if (handler.kind !== handlerType) {
-          arr.push(handler);
-        }
-      }
-      return arr;
-    });
-    return [].concat(...checker);
-  }
-
   async fillNextBlockBuffer(initBlockHeight: number): Promise<void> {
     await this.prefetchMeta(initBlockHeight);
 
@@ -459,11 +446,7 @@ export class FetchService implements OnApplicationShutdown {
         scaledBatchSize,
       );
 
-      if (
-        this.getModulos().length > 0
-        // &&
-        // this.checkHandlers(SubstrateHandlerKind.Block).length === 0
-      ) {
+      if (this.getModulos().length > 0) {
         this.runModuloBlocks(startBlockHeight, endHeight);
       } else {
         this.blockDispatcher.enqueueBlocks(
--- a/packages/node/src/indexer/fetch.service.ts
+++ b/packages/node/src/indexer/fetch.service.ts
@@ -455,7 +455,8 @@ export class FetchService implements OnApplicationShutdown {
         }
       }
 
-      const ds = this.project.dataSources.map((ds) => {
+      // TODO: make sure this check is working,
+      const handlerCheck = this.project.dataSources.map((ds) => {
         const arr = [];
         for (const handler of ds.mapping.handlers) {
           if (handler.kind !== SubstrateHandlerKind.Block) {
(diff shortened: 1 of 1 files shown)


packages/node/src/indexer/fetch.service.ts
```diff
@@ -435,14 +441,18 @@ export class FetchService implements OnApplicationShutdown {
...
-      // the original method: fill next batch size of blocks
       const endHeight = this.nextEndBlockHeight(
         startBlockHeight,
         scaledBatchSize,
       );
-      this.blockDispatcher.enqueueBlocks(
-        range(startBlockHeight, endHeight + 1),
-      );
+
+      if (this.getModulos().length > 0) {
```
stwiname, line 449: I think this length needs to be equal to the number of datasources. If theres a block handler without modulo then we would only fetch modulo blocks.

packages/node/src/indexer/fetch.service.ts
```diff
@@ -435,14 +441,18 @@ export class FetchService implements OnApplicationShutdown {
...
+        this.runModuloBlocks(startBlockHeight, endHeight);
```
stwiname, line 450: End height here wouldn't be right. The modulo is only going to be a subset of this and we might not fill the buffer. 

e.g. if modulo is 100 and batch size is 10 we would have 0 blocks added

packages/node/src/indexer/fetch.service.ts
```diff
@@ -435,14 +469,21 @@ export class FetchService implements OnApplicationShutdown {
...
         startBlockHeight,
         scaledBatchSize,
       );
-      this.blockDispatcher.enqueueBlocks(
-        range(startBlockHeight, endHeight + 1),
-      );
+
+      if (
+        this.getModulos().length ===
+        this.getDatasources(SubstrateHandlerKind.Block).length
```
stwiname, line 479: This should be for all handler kinds, not just Block handler kinds.

packages/node/src/indexer/fetch.service.ts
```diff
@@ -435,14 +469,21 @@ export class FetchService implements OnApplicationShutdown {
...
+      ) {
+        this.runModuloBlocks(startBlockHeight, endHeight);
```
stwiname, line 481: End height shouldn't be used at all with the modulo case

packages/node/src/indexer/fetch.service.ts
```diff
@@ -357,11 +357,45 @@ export class FetchService implements OnApplicationShutdown {
...
 
+  runModuloBlocks(startBlockHeight: number, endHeight: number): void {
+    let scalingEndHeight = endHeight + 1;
+    let modulosBlocks = this.getModuloBlocks(
+      startBlockHeight,
+      scalingEndHeight,
+    );
+
+    const largestModulo = Math.max(...this.getModulos());
+    if (largestModulo > this.nodeConfig.batchSize) {
```
stwiname, line 368: This will always be true, I think

packages/node/src/indexer/fetch.service.ts
```diff
@@ -357,11 +357,45 @@ export class FetchService implements OnApplicationShutdown {
...
+      scalingEndHeight =
+        Math.round(this.nodeConfig.batchSize * largestModulo) +
```
stwiname, line 370: Why round? These would always be integers

packages/node/src/indexer/fetch.service.ts
```diff
@@ -357,11 +357,45 @@ export class FetchService implements OnApplicationShutdown {
...
+        startBlockHeight;
+      modulosBlocks = this.getModuloBlocks(
+        startBlockHeight,
+        scalingEndHeight,
+      ).slice(0, this.nodeConfig.batchSize);
+    }
+    this.blockDispatcher.enqueueBlocks(modulosBlocks);
```
stwiname, line 377: Return `moduloBlocks` here

stwiname (approved):

PR #1360, opened 2022-10-21
hot fix tests

bz888:
# Description
update tests:
- api.service.spec
- init-controller
- publish-controller

### DIFF
--- a/packages/cli/src/controller/init-controller.test.ts
+++ b/packages/cli/src/controller/init-controller.test.ts
@@ -67,7 +67,7 @@ describe('Cli can create project', () => {
   it('prepare correctly applies project details', async () => {
     const tempPath = await makeTempDir();
     const templates = await fetchTemplates();
-    const template = templates.find(({name}) => name === 'Polkadot-starter');
+    const template = templates.find(({name}) => name.toLowerCase() === 'polkadot-starter');
     const projectPath = await cloneProjectTemplate(tempPath, projectSpec.name, template);
     await prepare(projectPath, projectSpec);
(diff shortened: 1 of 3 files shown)


packages/cli/src/controller/publish-controller.spec.ts
```diff
@@ -107,32 +105,45 @@ describe('Cli publish', () => {
...
     // await expect(Validate.run(['-l', cid, '--ipfs', ipfsEndpoint])).resolves.toBe(undefined);
   });
 
   it('upload project from a manifest', async () => {
-    projectDir = await createTestProject(projectSpecV0_2_0);
-    const manifestPath = path.resolve(projectDir, 'project.yaml');
-    const testManifestPath = path.resolve(projectDir, 'test.yaml');
-    fs.renameSync(manifestPath, testManifestPath);
-    await Publish.run(['-f', testManifestPath]);
+    try {
```
stwiname, line 138: Tests shouldn't have a try/catch, the testing framework should handle that.

packages/cli/src/controller/init-controller.test.ts
```diff
@@ -67,7 +67,7 @@ describe('Cli can create project', () => {
   it('prepare correctly applies project details', async () => {
     const tempPath = await makeTempDir();
     const templates = await fetchTemplates();
-    const template = templates.find(({name}) => name === 'polkadot-starter');
+    const template = templates.find(({name}) => name === 'Polkadot-starter');
```
stwiname, line 70: Why does this keep changing? It was changed not that long ago. Maybe we should just do a case insensitive check

stwiname (approved):

PR #1363, opened 2022-10-24
Fix template issues

hariu-starfish:
# Description

1. When multiple templates are created in the same block, the template information is not immediately saved to _metadata.dynamicDatasources, so only the address of the last template is saved in _metadata.dynamicDatasources. When the service is restarted, multiple identical templates are created that contain only the last address.
https://github.com/subquery/subql/blob/main/packages/node/src/indexer/dynamic-ds.service.ts#L98
```diff

async saveDynamicDatasourceParams(dsParams, tx) {
    ...
    // Only insert the last dynamic datasource in the current block event
-   await this.metaDataRepo.upsert({ key: METADATA_KEY, value: JSON.stringify([...existing, dsParams]) }, { transaction: tx });
+   await this.metaDataRepo.upsert({ key: METADATA_KEY, value: JSON.stringify([...existing, dsParams]) });
}

```
2. When creating multiple templates in the same block, due to object merging, in this._datasources, the last dsObj.processor.options will overwrite all previous dsObj.processor.options, resulting in an error in template event monitoring
https://github.com/subquery/subql/blob/main/packages/node/src/indexer/dynamic-ds.service.ts#L105
```diff
+ import _ from 'lodash';

async getDatasource(params) {
    // Dynamic datasources are the same in the current block event
-   const template = this.project.templates.find((t) => t.name === params.templateName);
+   const template = _.cloneDeep(this.project.templates.find((t) => t.name === params.templateName));
    ...
}
```
Fixes # (issue)

## Type of change

## Checklist

- [x] I have tested locally

### DIFF
--- a/packages/node/src/indexer/dynamic-ds.service.ts
+++ b/packages/node/src/indexer/dynamic-ds.service.ts
@@ -8,11 +8,12 @@ import { getLogger, MetadataRepo } from '@subql/node-core';
 import { Transaction } from 'sequelize/types';
 import { SubqlProjectDs, SubqueryProject } from '../configure/SubqueryProject';
 import { DsProcessorService } from './ds-processor.service';
-import _ from 'lodash';
+import { cloneDeep } from 'lodash';
 
 const logger = getLogger('dynamic-ds');
 
 const METADATA_KEY = 'dynamicDatasources';
+const TEMP_DS_PREFIX = 'ds_';
 
 interface DatasourceParams {
   templateName: string;
@@ -23,6 +24,7 @@ interface DatasourceParams {
 @Injectable()
 export class DynamicDsService {
   private metaDataRepo: MetadataRepo;
+  private tempDsRecords: Record<string, string>;
 
   constructor(
     private readonly dsProcessorService: DsProcessorService,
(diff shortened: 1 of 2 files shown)


packages/node/src/indexer/dynamic-ds.service.ts
```diff
@@ -94,17 +95,16 @@ export class DynamicDsService {
 
     assert(this.metaDataRepo, `Model _metadata does not exist`);
     await this.metaDataRepo.upsert(
-      { key: METADATA_KEY, value: JSON.stringify([...existing, dsParams]) },
-      { transaction: tx },
+      { key: METADATA_KEY, value: JSON.stringify([...existing, dsParams]) }
```
stwiname, line 98: Removing the tx here will lead to other problems. If the block fails to index for any reason then without a tx here the template will still be created and if indexing is resumed it will result in the template being created multiple times. 

I think a better option is to maintain `existing` in memory so we don't overwrite any other templates created in the same block

stwiname (changes requested):
Hi @hariu-starfish, thanks for your PR. Good find on both these issues.

packages/node/src/indexer/dynamic-ds.service.ts
europa-starfish: @stwiname  
Thanks for the comment.
In our testing, as long as `{ transaction: tx }` is there, templates will still override each other.

Could you please suggest how we could prevent such overriding ?

europa-starfish:
We come up with the approach of handling the existing record in a temporary map so that new creation won't override the existing ones.

stwiname (approved):

PR #1371, opened 2022-10-28
Imporve dictionary query

### LINKED ISSUE #1306: Improve dictionary with multiple datasources block height
**Issue:**

When we have project has multiple datasources block height, it will start index from the minimum block height among the datasources, and the dictionary query been build with filter conditions and with **SAME** start and end block height.

And the dictionary query will return all the block height matches with filter conditions but not within its own datasource start height. Then returned block heights will be fetched by api and enqueued, even we skipped these blocks within handlers, but some works has been done by subql/node but meaningless. 

image

image

**Proposed solution**

Before we build each query, we can compare ds start height with query's **END** height , if it is below then join the condition to the query.

bz888:
# Description

Fixes #1306 

## Type of change

## Checklist

- [x] I have tested locally


packages/node-core/src/utils/dictionaryEntry.ts
```diff
@@ -0,0 +1,66 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+import {Text} from '@polkadot/types-codec';
+import {DictionaryQueryEntry} from '@subql/types';
+import {Dictionary} from '../indexer/dictionary.service';
+// import { SubqlProjectDs } from '../../../node/src/configure/SubqueryProject';
+
+export function buildDictionaryEntryMap(
+  // unsure how to make this more generic
+  dataSources: any[],
```
stwiname, line 11: Just leave these as `SubqlProjectDs[]` for now.

packages/node/src/indexer/fetch.service.ts
```diff
@@ -215,10 +250,15 @@ export class FetchService implements OnApplicationShutdown {
   }
 
   updateDictionary(): void {
-    this.dictionaryQueryEntries = this.getDictionaryQueryEntries();
+    this.buildDictionaryEntriesMap();
+    let checkDictionaryQueryEntries: boolean;
+    this.mappedDictionaryQueryEntries.forEach((value) => {
+      if (value?.length) {
```
stwiname, line 256: Im not sure this works. We could have dictionary disabled at a later time if a block filter exists

packages/node-core/src/utils/dictionaryEntry.ts
```diff
@@ -0,0 +1,66 @@
+// Copyright 2020-2022 OnFinality Limited authors & contributors
+// SPDX-License-Identifier: Apache-2.0
+
+import {Text} from '@polkadot/types-codec';
+import {DictionaryQueryEntry} from '@subql/types';
+import {Dictionary} from '../indexer/dictionary.service';
+// import { SubqlProjectDs } from '../../../node/src/configure/SubqueryProject';
+
+export function buildDictionaryEntryMap(
```
stwiname, line 9: I'd like to see these functions as a class, and we can maintain the `mappedDictionaryQueryEntries` in the class. It would be good to clean up the args to these functions too

packages/node-core/src/indexer/dictionary.service.ts
```diff
@@ -228,4 +229,47 @@ export class DictionaryService implements OnApplicationShutdown {
...
+    buildDictionaryQueryEntries: (startBlock: number) => DictionaryQueryEntry[]
+  ): void {
+    const mappedDictionaryQueryEntries = new Map();
+
+    for (const ds of dataSources) {
+      mappedDictionaryQueryEntries.set(ds.startBlock, buildDictionaryQueryEntries(ds.startBlock));
+    }
+    // sort map from lowest ds.startBlock to highest
+    this.mappedDictionaryQueryEntries = new Map(
+      [...mappedDictionaryQueryEntries.entries()].sort((a, b) => a[0] - b[0])
```
stwiname, line 245: Might be easier to sort the datasources

packages/node/src/indexer/fetch.service.test.ts
```diff
@@ -744,4 +738,20 @@ describe('FetchService', () => {
     // Should be called 91151,9170,9180
     expect(getPrefechMetaSpy).toBeCalledTimes(3);
   }, 500000);
+
```
jiqiang90, line 741: do we still need this?

packages/node/src/indexer/fetch.service.test.ts
```diff
@@ -15,7 +15,7 @@ import {
 import { NodeConfig } from '@subql/node-core';
 import { GraphQLSchema } from 'graphql';
 import { Sequelize } from 'sequelize';
-import { SubqueryProject } from '../configure/SubqueryProject';
+import { SubqlProjectDs, SubqueryProject } from '../configure/SubqueryProject';
```
jiqiang90, line 18: and this?

packages/node-core/src/indexer/dictionary.service.test.ts
```diff
@@ -143,4 +205,60 @@ describe('DictionaryService', () => {
...
+  });
+
+  it('able to getDicitonaryQueryEntries', () => {
+    const dictionaryService = new DictionaryService(DICTIONARY_ENDPOINT, nodeConfig);
+    const dictionaryQueryMap = new Map();
+
+    for (let i = 0; i < mockDS.length; i++) {
+      dictionaryQueryMap.set(
+        [mockDS[i].startBlock],
+        HAPPY_PATH_CONDITIONS.filter((dictionaryQuery, index) => i >= index)
```
jiqiang90, line 226: I know you are trying to mock `HAPPY_PATH_CONDITIONS` here, but it is not align with mockDs filters, can we make them match? Or just leave a comment here, otherwise people will confused

jiqiang90 (approved):

jiqiang90:
Note: 

To decide whether include ds filter condition in the query, we are using queryEndBlock height here to compare with ds start height. 

Use nextBlock (endblock + batch) is **not** correct, as the queryEndblock is larger, the last block in previous query result might skip the nextBlock, so between 2nd ds start height to the last block in previous query result is not been checked by dictionary.

packages/node-core/src/indexer/dictionary.service.ts
```diff
@@ -228,4 +229,43 @@ export class DictionaryService implements OnApplicationShutdown {
     }
     return buildQuery(vars, nodes);
   }
+  buildDictionaryEntryMap(
+    // TODO: generic type
+    dataSources: any[],
```
stwiname, line 234: We should type this.

PR #1374, opened 2022-10-30
Fix bugs with unfinalized blocks

stwiname:


packages/node/src/indexer/unfinalizedBlocks.service.ts
```diff
@@ -88,7 +88,7 @@ export class UnfinalizedBlocksService {
     if (block) {
       await this.registerUnfinalizedBlock(
         block.block.header.number.toNumber(),
-        block.hash.toHex(),
```
ianhe8x, line 91: ` block.hash` should be equal to `block.block.header.hash`, in packages/types/src/generic/Block.ts

stwiname: From my testing they are not the same. This was the reason that forks would happen when there wasn't actually a fork

ianhe8x (approved):

Issue #1290, opened 2022-09-12
Multi-chain indexing


stwiname:
## Outcome

- A user of SubQuery can have a single dataset that combines data indexed from different chains.
- All data would be indexed in the exact same schema. 
- The user can access this data via one unified endpoint where they can get ordered and combined data from all networks.

For example, a dataset that includes all transactions from all Polkadot parachains that is accessed through a single unified endpoint. 

## Initial idea

Allow running multiple nodes against the same database/schema:

### Requirements:

* The nodes must all have the same graphql schema
* It should be possible to add more chains later on
* We need an attribute to every table to include an array of chain ids that have modified the record
* All project db operations should be an upsert, this is to avoid any race condition

### Internal changes

* The `_metadata` table will go from being a key/value store to a table with the current keys as columns
* The first project in the `_metadata` table will be the master, it will be responsible for migrations.
* Historical support is not going to be possible so it will have to be disabled.

## Other notes

- This could lead to a feature where we index portions of the chain with different nodes. e.g. Node 1: Blocks 1 - 1000000, Node 2: Blocks 1000001 - 2000000 and so on.

PR #1375, opened 2022-10-31
Feat/multi chain indexing

lamcc21:
Feat #1290 

## Checklist

- [x] I have manually tested locally

## Todo

- [x] Only the first network can run database migrations (first entry of metadata table)
- [x] If there is a metadata entry exists and historical is enabled throw error and exit
- [x] Update dictionary queries
- [x] Rebase

Usage: subql-node --multi-chain --db-schema=name


packages/query/src/graphql/plugins/GetMetadataPlugin.ts
```diff
@@ -12,7 +13,7 @@ const {version: packageVersion} = require('../../../package.json');
 
 const METADATA_TYPES = {
   lastProcessedHeight: 'number',
-  lastProcessedTimestamp: 'number',
+  lastProcessedTimestamp: 'string',
```
stwiname, line 16: reason for this being a string now?

packages/node-core/src/db/db.module.ts
```diff
@@ -39,17 +40,43 @@ async function establishConnection(sequelize: Sequelize, numRetries: number): Pr
...
+
+    const {chainId} = project.network;
+    const {dbSchema, migrate} = nodeConfig;
+
+    let chainIdMatch = false;
+
+    if (migrate) {
+      try {
+        const result = await sequelize.query<{chain_id: string}>(`SELECT chain_id FROM "${dbSchema}"._metadata`, {
+          type: QueryTypes.SELECT,
```
lamcc21, line 57: mistake here didn't order by

packages/node-core/src/indexer/store.service.ts
```diff
@@ -392,13 +407,28 @@ export class StoreService {
...
 
-  async setMetadata(key: string, value: string | number | boolean, options?: UpsertOptions<Metadata>): Promise<void> {
+  async setMetadata(key: keyof Metadata, value: string | number | boolean, tx?: Transaction): Promise<void> {
+    const {chainId} = this.subqueryProject.network;
     assert(this.metaDataRepo, `Model _metadata does not exist`);
-    await this.metaDataRepo.upsert({key, value}, options);
+
+    //TODO: fix typing
+    const entry: any = await this.metaDataRepo.findByPk(chainId);
+
```
lamcc21, line 423: Couldn't figure out how to type this better.

packages/node-core/src/db/db.module.ts
```diff
@@ -39,17 +40,44 @@ async function establishConnection(sequelize: Sequelize, numRetries: number): Pr
...
-  return sequelize;
-};
+const sequelizeFactory =
+  (option: SequelizeOption, project: ISubqueryProject<IProjectNetworkConfig>, nodeConfig: NodeConfig) => async () => {
+    const sequelize = new Sequelize(option);
+    const numRetries = 5;
+    await establishConnection(sequelize, numRetries);
+
+    const {chainId} = project.network;
+    const {dbSchema, migrate} = nodeConfig;
```
jiqiang90, line 50: we need to deprecate migrate in nodeConfig

packages/node-core/src/db/db.module.ts
```diff
@@ -39,17 +40,44 @@ async function establishConnection(sequelize: Sequelize, numRetries: number): Pr
...
+
+    let chainIdMatch = false;
+
+    if (migrate) {
+      try {
+        const result = await sequelize.query<{chain_id: string}>(
+          `SELECT chain_id FROM "${dbSchema}"._metadata ORDER BY "createdAt"`,
+          {type: QueryTypes.SELECT}
+        );
+        if (result[0].chain_id === chainId) chainIdMatch = true;
```
jiqiang90, line 60: rather than compare with chainId from project.yaml (as it is editable), we should compare it with existing genesishHash/chainId in metadata

packages/node-core/src/indexer/dictionary.service.ts
```diff
@@ -165,9 +165,16 @@ export class DictionaryService implements OnApplicationShutdown {
     startBlock: number,
     queryEndBlock: number,
     batchSize: number,
+    metadataTableName: string,
```
stwiname, line 168: Can we make this the last arg and default like so
```suggestion
    metadataTableName = '_metadata',
```

packages/node/src/indexer/project.service.ts
```diff
@@ -287,11 +300,11 @@ export class ProjectService {
   }
 
   async getMetadataBlockOffset(): Promise<number | undefined> {
-    return getMetaDataInfo(this.metadataRepo, 'blockOffset');
+    return getMetaDataInfo(this.metadataRepo, 'block_offset');
   }
 
   async getLastProcessedHeight(): Promise<number | undefined> {
-    return getMetaDataInfo(this.metadataRepo, 'lastProcessedHeight');
+    return getMetaDataInfo(this.metadataRepo, 'last_processed_height');
```
stwiname, line 307: Should these be reverted?

packages/node-core/src/configure/NodeConfig.ts
```diff
@@ -56,10 +55,8 @@ const DEFAULT_CONFIG = {
   proofOfIndex: false,
   dictionaryTimeout: 30,
   profiler: false,
-  migrate: false,
   subscription: false,
-  disableHistorical: true,
-  unfinalizedBlocks: false,
```
stwiname, line 62: Can we bring this back please

packages/query/src/graphql/plugins/GetMetadataPlugin.ts
```diff
@@ -55,11 +55,12 @@ async function fetchFromApi(): Promise<void> {
   }
 }
 
-async function fetchFromTable(pgClient: any, schemaName: string): Promise<MetaData> {
+async function fetchFromTable(pgClient: any, schemaName: string, chainId: string | undefined): Promise<MetaData> {
   const metadata = {} as MetaData;
   const keys = Object.keys(METADATA_TYPES);
 
-  const {rows} = await pgClient.query(`select key, value from "${schemaName}"._metadata WHERE key = ANY ($1)`, [keys]);
+  const tableName = `_metadata${chainId ? `_${chainId.substring(0, 10)}` : ''}`;
```
stwiname, line 62: I don't think we can do `.substring(0, 10)`. Theres no constraint on the length of the chain it and some networks have chain ids < 10 (e.g. `juno-1`)

stwiname (commented):
The metadata query changes (query service and dictionary queries) are a little confusing. It looks like to query the metadata you need to provide chainId as a parameter but the dictionary is using a different entity name.

We also need to maintain backwards compatibility with dictionary queries.
1. We can't break existing versions of the SDK if we update a dictionary to use this
2. These node changes could be working with a dictionary using the current SDK

I think we could do a couple of things for compatibility:
1. Do something like [this](https://github.com/subquery/subql/pull/1274/files#diff-3f4f058897ecda3c4e15f6c34439075d1bdb00f51023c5462d58f576de2081c6R139) and try the new way, otherwise fall back to the old way
2. The query service should support querying `_metadata` and `_metadata(chainId: String)`. If chainId is not specified we could use the first `metadata` table we find. Or add an argument to the service to specify a specific one.

packages/node/src/indexer/fetch.service.ts
```diff
@@ -440,6 +442,7 @@ export class FetchService implements OnApplicationShutdown {
               startBlockHeight,
               queryEndBlock,
               scaledBatchSize,
+              this.projectService.metadataName,
```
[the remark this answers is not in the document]
lamcc21: The dictionary service that defines this method is the node-core one. So doesn't have access to project service.

packages/node-core/src/indexer/store.service.ts
```diff
@@ -267,23 +269,51 @@ export class StoreService {
   }
 
   async getHistoricalStateEnabled(): Promise<boolean> {
-    let enabled = true;
+    const {disableHistorical} = this.config;
+    const multiChainHistoricalError =
+      'Cannot multi-chain index when historical state is enabled, try using --disable-historical';
```
stwiname, line 274: I think we want to change this behaviour. 

If historical is enabled when the project is started theres no way to disable it. I think we should keep that. So if a user tries to change to multi chain and theres any historical data we can throw an error and tell them to index from scratch

packages/query/src/graphql/plugins/smartTagsPlugin.ts
```diff
@@ -8,7 +8,7 @@ export const smartTagsPlugin = makePgSmartTagsPlugin([
   {
     //Rule 1, omit `_metadata` from node
     kind: PgEntityKind.CLASS,
-    match: ({name}: PgEntity) => /^_metadata$/.test(name),
+    match: ({name}: PgEntity) => /^_metadata$/.test(name) || /^_metadata_[a-zA-Z0-9-]+$/.test(name),
```
stwiname, line 11: Can we define this regex in utils package? Its used in a few places, it would be good to have a single source of truth

packages/node-core/src/indexer/store.service.ts
```diff
@@ -282,23 +290,57 @@ export class StoreService {
...
 
   async getHistoricalStateEnabled(): Promise<boolean> {
-    let enabled = true;
+    const {multiChain} = this.config;
+    let {disableHistorical} = this.config;
+
+    if (multiChain && !disableHistorical) {
+      logger.info('Historical state is not compatible with multi chain indexing, disabling historical..');
+      disableHistorical = true;
+    }
```
stwiname, line 299: The disableHistorical flag only applies with clean projects. e.g if you had already indexed historical and then disable through the flag it will still be enabled because the database is setup for it.

We should retain this behaviour, if the project already exists then historical is enabled based on the db schema.

With this current behaviour if multiChain is enabled and there is historical data its going to cause issues with the DB.

packages/query/src/graphql/plugins/GetMetadataPlugin.ts
```diff
@@ -55,11 +56,13 @@ async function fetchFromApi(): Promise<void> {
...
 }
 
-async function fetchFromTable(pgClient: any, schemaName: string): Promise<MetaData> {
+async function fetchFromTable(pgClient, schemaName: string, chainId: string | undefined): Promise<MetaData> {
   const metadata = {} as MetaData;
   const keys = Object.keys(METADATA_TYPES);
 
-  const {rows} = await pgClient.query(`select key, value from "${schemaName}"._metadata WHERE key = ANY ($1)`, [keys]);
+  // 63 chars is the max postgres database table name length
+  const tableName = `_metadata${chainId ? `_${chainId.substring(0, 63)}` : ''}`;
```
stwiname, line 64: I think this can be in utils package and use by query and node

stwiname (approved):

packages/node-core/src/indexer/entities/Metadata.entity.ts
```diff
@@ -14,9 +37,39 @@ export type MetadataRepo = typeof Model & {
...
+  if (multichain) {
+    const oldMetadataName = await checkSchemaMetadata(sequelize, schema, chainId);
+
+    if (oldMetadataName) {
+      throw new Error(
+        'Found metadata entry with matching chain but wrong table name, must match _metadata_[chainId] syntax for multi-chain indexing.'
+      );
+    }
+
+    tableName = `${tableName}_${chainId}`;
```
jiqiang90, line 68: have we count how long the table name is ? According to rule https://til.hashrocket.com/posts/8f87c65a0a-postgresqls-max-identifier-length-is-63-bytes#:~:text=In%20PostgreSQL%2C%20identifiers%20%E2%80%93%20table%20names,the%20allowed%20length%20of%2063., we can not access 63 bytes.

jiqiang90: image

packages/node-core/src/indexer/store.service.ts
```diff
@@ -104,7 +112,7 @@ export class StoreService {
 
   async incrementJsonbCount(key: string, tx?: Transaction): Promise<void> {
     await this.sequelize.query(
-      `UPDATE "${this.schema}"._metadata SET value = (COALESCE(value->0):: int + 1)::text::jsonb WHERE key ='${key}'`,
+      `UPDATE "${this.schema}".${this.metaDataRepo.tableName} SET value = (COALESCE(value->0):: int + 1)::text::jsonb WHERE key ='${key}'`,
```
jiqiang90, line 115: the over size table name will make problem here

packages/node-core/src/indexer/entities/Metadata.entity.ts
```diff
@@ -14,9 +37,39 @@ export type MetadataRepo = typeof Model & {
...
+  multichain: boolean,
+  chainId: string
+): Promise<MetadataRepo> {
+  let tableName = '_metadata';
+
+  if (multichain) {
+    const oldMetadataName = await checkSchemaMetadata(sequelize, schema, chainId);
+
+    if (oldMetadataName) {
+      throw new Error(
```
jiqiang90, line 63: I think the message here be can be improved. 

It is like if we had a oldMetadata table here, we can not process any further. We need to tell them multi-chain indexing not compatible with old project, user should have `multi-chain` flag on when first time indexing . 

Either we are able to covert the old metadata table in future pr.

jiqiang90 (changes requested):

packages/node-core/src/indexer/entities/Metadata.entity.ts
```diff
@@ -1,11 +1,34 @@
...
 // SPDX-License-Identifier: Apache-2.0
 
-import {BuildOptions, DataTypes, Model, Sequelize} from 'sequelize';
+import {BuildOptions, DataTypes, Model, QueryTypes, Sequelize} from 'sequelize';
+
+export interface MetadataKeys {
+  chain: string;
+  genesisHash: string;
+  historicalStateEnabled: boolean;
+  indexerNodeVersion: string;
```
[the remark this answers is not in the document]
lamcc21: Screen Shot 2022-11-18 at 1 27 59 PM

packages/utils/src/query/metadata.ts
```diff
@@ -1,11 +1,17 @@
...
 export const METADATA_REGEX = /^_metadata$/;
 
 export const MULTI_METADATA_REGEX = /^_metadata_[a-zA-Z0-9-]+$/;
 
-export function getMetadataTableName(chainId: string): string {
-  // 63 chars is the max postgres database table name length
-  return `_metadata${chainId ? `_${chainId.substring(0, 63)}` : ''}`;
+export function getMetadataTableName(chainId: string | undefined): string {
+  if (chainId) {
+    const hash = blake2AsHex(chainId);
```
jiqiang90, line 12: change to use  64 bytes  `const hash = blake2AsHex(chainId,64)`; the output hash length will be 18.

jiqiang90 (approved):

PR #1401, opened 2022-11-14
Hot schema trigger

### LINKED ISSUE #799: hot reload schema in the query service on schema change
when an indexer updated/add a table, we have to restart the query service to get the latest schema generated from the db tables. this is not convenient and can be auto-updated.

graphile provides a `watchSchema` method may help. https://github.com/graphile/graphile-engine/blob/1619878d96/packages/graphile-build/src/SchemaBuilder.js#L526

bz888:
# Description
Updated fix for `hot-schema-reload`
Fixes #799 

## Checklist

- [x] I have tested locally

### DIFF
--- a/packages/query/src/graphql/graphql.module.ts
+++ b/packages/query/src/graphql/graphql.module.ts
@@ -125,8 +125,10 @@ export class GraphqlModule implements OnModuleInit, OnModuleDestroy {
       const pgClient = await this.pgPool.connect();
       await pgClient.query(`LISTEN "${dbSchema}._metadata.hot_schema"`);
 
-      pgClient.on('notification', () => {
-        void this.schemaListener(dbSchema, options);
+      pgClient.on('notification', (msg) => {
+        if (msg.payload === 'schema_updated') {
+          void this.schemaListener(dbSchema, options);
+        }
       });
     }
     const schema = await this.buildSchema(dbSchema, options);
--- a/packages/query/src/graphql/graphql.module.ts
+++ b/packages/query/src/graphql/graphql.module.ts
@@ -57,6 +57,7 @@ export class GraphqlModule implements OnModuleInit, OnModuleDestroy {
 
   async schemaListener(schema: GraphQLSchema): Promise<void> {
     // In order to apply hotSchema Reload without using apollo Gateway, must access the private method, hence the need to use set()
+
     try {
       // @ts-ignore
       if (schema && !!this.apolloServer?.generateSchemaDerivedData) {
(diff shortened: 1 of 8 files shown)


packages/query/src/graphql/graphql.module.ts
```diff
@@ -123,7 +120,16 @@ export class GraphqlModule implements OnModuleInit, OnModuleDestroy {
         options.replaceAllPlugins.push(options.appendPlugins.pop());
       }
     }
-    const schema = await this.buildSchema(dbSchema, options, SCHEMA_RETRY_NUMBER);
+
+    if (!argv['disable-hot-schema']) {
+      const pgClient = await this.pgPool.connect();
+      await pgClient.query(`LISTEN "${dbSchema}._metadata.hot_schema"`);
+
+      pgClient.on('notification', () => {
```
stwiname, line 128: I think we need to filter here for a specific notification type. Like check if the notification relates to `hot_schema`

packages/query/src/graphql/graphql.module.ts
```diff
@@ -55,8 +55,10 @@ export class GraphqlModule implements OnModuleInit, OnModuleDestroy {
     }
   }
 
-  async schemaListener(schema: GraphQLSchema): Promise<void> {
+  async schemaListener(dbSchema: string, options: PostGraphileCoreOptions): Promise<void> {
     // In order to apply hotSchema Reload without using apollo Gateway, must access the private method, hence the need to use set()
+    const schema = await this.buildSchema(dbSchema, options);
```
stwiname, line 60: Can we move this to the try/catch please

packages/query/src/graphql/graphql.module.ts
```diff
@@ -123,7 +120,18 @@ export class GraphqlModule implements OnModuleInit, OnModuleDestroy {
...
     }
-    const schema = await this.buildSchema(dbSchema, options, SCHEMA_RETRY_NUMBER);
+
+    if (!argv['disable-hot-schema']) {
+      const pgClient = await this.pgPool.connect();
+      await pgClient.query(`LISTEN "${dbSchema}._metadata.hot_schema"`);
+
+      pgClient.on('notification', (msg) => {
+        if (msg.payload === 'schema_updated') {
+          void this.schemaListener(dbSchema, options);
```
stwiname, line 130: I think we should exit if this fails. In the catch for schemaListener, can we log the error and process.exit

stwiname (approved):

PR #1404, opened 2022-11-15
fix hot schema

bz888:


packages/node-core/src/indexer/store.service.ts
```diff
@@ -191,10 +191,19 @@ export class StoreService {
 
       /* These SQL queries are to allow hot-schema reload on query service */
       extraQueries.push(createSchemaTriggerFunction(schema));
-      extraQueries.push(createSchemaTrigger(schema));
+      const schemaTriggerName = makeTriggerName(schema, '_metadata', 'schema');
+
+      const schemaNotifyTrigger = await this.sequelize.query(getNotifyTriggers(), {
```
jiqiang90, line 196: maybe rename `getNotifyTriggers()` to `getTriggers()`, since both `notify` and `schema` trigger are using it

stwiname (approved):

PR #1411, opened 2022-11-16
dictionary auth link integration

lamcc21:
# Description
Adds functionary use use sponsored endpoint for dictionary.

## Type of change

## Checklist

- [x] I have tested locally

## TODO

- [x] update code to use flag
- [x] add retry logic
- [x] replace everything that's hard coded


packages/node-core/src/indexer/dictionary.service.ts
```diff
@@ -140,12 +142,38 @@ export class DictionaryService implements OnApplicationShutdown {
...
-  ) {
+  ) {}
+
+  async init(): Promise<void> {
+    const headers = {'Content-Type': 'application/json'};
+
+    //TODO: need to use flag here
+    //TODO: add retry logic
+
+    const res = await axios.get(`${this.dictionaryEndpoint}/metadata/${this.chainId}`, {headers});
```
HZHAIX, line 156: Use `AuthHttpLink` instead of `AuthLink`.
AuthHttpLink include both `HttpLink` and `AuthLink`, so that you don't need to handle the `metadata` request

lamcc21: Was just doing it this way before I could restructure to use async method.

HZHAIX:
Need to fix the failed test?

packages/node/src/indexer/indexer.module.ts
```diff
@@ -31,7 +36,15 @@ import { WorkerService } from './worker/worker.service';
       },
       inject: [SubqueryProject, EventEmitter2],
     },
-    DictionaryService,
+    {
```
stwiname, line 39: Does indexer.module need the dictionary service? I feel like its unnecessary

packages/node-core/src/indexer/dictionary.service.ts
```diff
@@ -140,12 +142,28 @@ export class DictionaryService implements OnApplicationShutdown {
...
     protected readonly metadataKeys = ['lastProcessedHeight', 'genesisHash'] // Cosmos uses chain instead of genesisHash
-  ) {
+  ) {}
+
+  async init(): Promise<void> {
+    let link: ApolloLink;
+
+    if (this.nodeConfig.authDictionary) {
+      assert(this.dictionaryEndpoint, 'dictionary endpoint must be included in manifest to use --authDictionary');
+      link = await authHttpLink({
```
stwiname, line 155: Do we need to handle errors this could throw?

packages/node/src/yargs.ts
```diff
@@ -152,6 +152,12 @@ export const yargsOptions = yargs(hideBin(process.argv))
       describe: 'Specify the dictionary api for this network',
       type: 'string',
     },
+    'auth-dictionary': {
+      demandOption: false,
+      describe: 'Use auth center as endpoint for dictionary',
```
stwiname, line 157: It would be good to have a better name and description.

packages/node-core/src/indexer/dictionary.service.ts
```diff
@@ -140,12 +142,28 @@ export class DictionaryService implements OnApplicationShutdown {
...
+        authUrl: this.dictionaryEndpoint,
```
stwiname, line 156: Do you have an example with this url?

packages/node/src/indexer/indexer.module.ts
lamcc21: Didn't add it, I've removed it though.

packages/node/src/yargs.ts
lamcc21: updated

packages/node-core/src/indexer/dictionary.service.ts
lamcc21: updated

stwiname (commented):
Im happy with these changes but it seems the changes to graphql deps has broken a few tests. `packages/utils/src/graphql/graphql.spec.ts` there is a compiler error here now

lamcc21:
> Im happy with these changes but it seems the changes to graphql deps has broken a few tests. `packages/utils/src/graphql/graphql.spec.ts` there is a compiler error here now

This compile error has been fixed

stwiname (approved):

PR #1424, opened 2022-11-21
Manifest header

DeveloperInProgress:
Add header option to manifest network that is to be added to api connection


DeveloperInProgress:
wrong repo

PR #1435, opened 2022-11-25
Bypass blocks

### LINKED ISSUE #1413: Add bypass blocks in manifest, allow skip blocks
3 chains (Karura, KILT, Bifrost) have known issues where past blocks can no longer be decoded from PolkadotJS. These blocks that can’t be decoded stall a SubQuery project as we prevent SubQuery from indexing data in a non-deterministic way by default.

Some customers would prefer that we bypass these blocks that can't be decoded, since imperfect data is better than no data. The tricky thing here is that when it fails to fetch a block we have to detect whether network issue or decode issue, it also difficult for keep update with different network family.

- [ ] add field in project.yaml that changes the behaviour of a SubQuery project to bypass blocks on fail
- [ ] create a field in _metadata (or a new table) that contains an array of all bypassed blocks and the error log for each block
- [ ] On Automated historical indexing, remove blocks from the bypassed blocks array on reindexing
- [ ] The solution should also work with some cosmos chains that have skipped blocks in their past (e.g. Juno)

bz888:
# Description
Fixes #1413
Still working on Range aspect of the bypassBlocks

## Feature

## Checklist

- [x] I have tested locally


packages/node/src/indexer/fetch.service.ts
```diff
@@ -481,23 +493,45 @@ export class FetchService implements OnApplicationShutdown {
...
     }
   }
+  private filteredBlockBatch(batchBlocks: number[]): number[] {
+    const minBypass = Math.min(...this.bypassBlocks);
+    const maxBatchBlock = Math.max(...batchBlocks);
+    const bypassingBlocks = batchBlocks.filter((blk) =>
+      this.bypassBlocks.includes(blk),
+    );
+
+    if (!this.bypassBlocks?.length && maxBatchBlock < minBypass) {
```
jiqiang90, line 521: is this suppose be or condition? 
if this.bypassBlocks is empty we can return, or if maxBatchBlock < minBypass we also return? 
Indeed if this.bypassBlocks is empty, minBypass will return Infinity, so check 
this.bypassBlocks?.length is not necessary here.

packages/node/src/indexer/fetch.service.ts
```diff
@@ -481,23 +493,45 @@ export class FetchService implements OnApplicationShutdown {
...
-          range(startBlockHeight, endHeight + 1),
+          this.filteredBlockBatch(range(startBlockHeight, endHeight + 1)),
         );
       }
     }
   }
+  private filteredBlockBatch(batchBlocks: number[]): number[] {
+    const minBypass = Math.min(...this.bypassBlocks);
+    const maxBatchBlock = Math.max(...batchBlocks);
+    const bypassingBlocks = batchBlocks.filter((blk) =>
```
jiqiang90, line 517: this can move under after done `maxBatchBlock < minBypass ` checking

packages/node/src/indexer/project.service.ts
```diff
@@ -253,6 +254,12 @@ export class ProjectService {
         'Specified project manifest chain id / genesis hash does not match database stored genesis hash, consider cleaning project schema using --force-clean',
       );
     }
+    if (this.project.network.bypassBlocks) {
```
jiqiang90, line 257: I think we should only upsert the blocks that has been bypassed, what if user changed this in manifest..

Previous bypassed can be changed

jiqiang90 (changes requested):

packages/node/src/indexer/fetch.service.ts
```diff
@@ -481,23 +501,54 @@ export class FetchService implements OnApplicationShutdown {
...
+        this.blockDispatcher.enqueueBlocks(cleanedBatchBlocks);
+        this.setLatestBufferHeight(cleanedBatchBlocks, enqueuingBlocks);
       } else {
-        this.blockDispatcher.enqueueBlocks(
-          range(startBlockHeight, endHeight + 1),
-        );
+        const enqueuingBlocks = range(startBlockHeight, endHeight + 1);
+        const cleanedBatchBlocks = this.filteredBlockBatch(enqueuingBlocks);
+        this.blockDispatcher.enqueueBlocks(cleanedBatchBlocks);
+        this.setLatestBufferHeight(cleanedBatchBlocks, enqueuingBlocks);
```
jiqiang90, line 518: maybe add comment here why we need to set last buffer height manually

jiqiang90 (approved):

packages/node/src/indexer/fetch.service.ts
```diff
@@ -481,23 +502,56 @@ export class FetchService implements OnApplicationShutdown {
...
```
stwiname, line 519: Maybe `enqueueBlocks` should take a `lastBufferedHeight` parameter as we always call the 2 together

jiqiang90: Maybe we can set `enqueueBlocks(cleanedBatchBlocks: number[],  lastBufferedHeight?: number )`. 
If `lastBufferedHeight` is defined, set it to blockDispatcher, otherwise get last one from `cleanedBatchBlocks `

packages/node/src/indexer/blockDispatcher/base-block-dispatcher.ts
```diff
@@ -22,7 +22,7 @@ export interface IBlockDispatcher {
     runtimeService?: RuntimeService,
   ): Promise<void>;
 
-  enqueueBlocks(heights: number[]): void;
+  enqueueBlocks(heights: number[], rawBlocks?: number[]): void;
```
stwiname, line 25: I assume the worker block dispatcher needs updating too? Ideally we can reuse the new logic so we dont need to duplicate it in each implmentation

jiqiang90 (approved):

PR #1438, opened 2022-11-28
hash sql queries

bz888:
# Description
hashing SQL queries
- Fix: SQL method name too long

## Checklist

- [x] I have tested locally

### DIFF
--- a/packages/utils/src/query/metadata.ts
+++ b/packages/utils/src/query/metadata.ts
@@ -15,5 +15,5 @@ export function getMetadataTableName(chainId: string): string {
 // Hash names of SQL functions, triggers, channels to ensure it does not exceed the char limit
 export function hashName(schema: string, type: string, tableName: string): string {
   // Postgres identifier limit is 63 bytes (chars)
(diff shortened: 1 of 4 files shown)


packages/utils/src/query/metadata.ts
```diff
@@ -11,3 +11,7 @@ export function getMetadataTableName(chainId: string): string {
   const hash = blake2AsHex(chainId, 64);
   return `_metadata_${hash}`.substring(0, 63); // 63 chars is the max postgres database table name length
 }
+
+export function hashName(schema: string, type: string, tableName: string): string {
+  return blake2AsHex(`${schema}_${tableName}_${type}`).substr(2, 10);
```
stwiname, line 16: Whats the reason for 10? Would be good to have a comment.

stwiname (approved):

packages/utils/src/query/metadata.ts
jiqiang90: I think just change it to `blake2AsHex(chainId, 64)`

PR #1491, opened 2023-01-17
Worker use runtime service

### LINKED ISSUE #1403: Fix optimise block getRuntimeVersion with dictionary
### Prerequisites

* [ ] Are you running the latest version(s)?
* [ ] Did you check the [debugging guide](https://doc.subquery.network/academy/tutorials_examples/debug-projects.html#how-to-debug-a-subquery-project)?
* [ ] Did you check the [FAQs](https://doc.subquery.network/faqs/faqs.html) and [Discussions](https://github.com/subquery/subql/discussions)?
* [ ] Are you reporting to the correct repository?
* [ ] Did you search for an existing issue or pull request?

### Description

Since we use block dispatcher and worker, this part of logic is missing or not been applied. It causing each batch of blocks to call api get runtime.

### Details

**Network Details:**
  * [Network]

### Steps to Reproduce

1. [First Step]
2. [Second Step]
3. [and so on...]

### Any other information

jiqiang90:
# Description

Main runtime service fetch specVersion maps once, and each workers has own runtime service, copied specVersions from map.

After that, each worker runtime service maintain own runtimes and specVersion.

TODO/Improvement:  

- [x] if able to know which workers has the latest specVersion map, could sync in between main and each workers. 
- [x] wait issue https://github.com/subquery/subql/issues/1490 to be fixed
- [x] for block in same block specVersion, may distribute in each workers, and runtime been fetched multiple time, if we can handle this in main and distribute to each will be nice. 

Fixes #1403

## Type of change

## Checklist

- [ ] I have tested locally


packages/node/src/indexer/blockDispatcher/worker-block-dispatcher.service.ts
```diff
@@ -112,7 +110,15 @@ export class WorkerBlockDispatcherService
 
     const blockAmount = await this.projectService.getProcessedBlockCount();
     this.setProcessedBlockCount(blockAmount ?? 0);
-    this.runtimeService = runtimeService;
+
+    // copy Runtime service to each worker
+    this.workers.map((w) =>
+      w.syncRuntimeService(
```
stwiname, line 116: If this is only called once, why not make it part of the `init` worker function so that its clearer. The way I read this function is that it should be called each time the state of the main thread runtime service changes.

stwiname (commented):
Maybe another approach would be to have a worker version of the runtime service thats got the same interface, but rather than it fetching data from the api/dictionary it gets fed data from the main thread. 

Im not sure what the advantages of that approach would be, but it would remove the need for dictionary service dependency and maybe reduce some network requests

jiqiang90:
> Maybe another approach would be to have a worker version of the runtime service thats got the same interface, but rather than it fetching data from the api/dictionary it gets fed data from the main thread.

It kind of have similar logic here, current each worker has it own runtime service. However they only copyied a few thing from main thread runtime service at init. 

I'm still improving this, trying to achieve this flow here. This will reduce use dictionary for each worker runtime service. However, api will still needed for detect specVersion change when block out of dictionary specVersions.
  
image

> 
> Im not sure what the advantages of that approach would be, but it would remove the need for dictionary service dependency and maybe reduce some network requests


The most important change is made to improve this logic here.

Before: worker fetch each block will require get its runtime version.

After:  worker use its own runtime service to check whether block height specVersion changed (by specVersion map provide from dictionary, which synced from main runtime service) compare to previous processed block. And if not change, the runtime stored in local memory can be re-used. This will save one api call for each block. 

https://github.com/subquery/subql/blob/c4fd01623b3a0efcf5d9de2b559586858298b5b2/packages/node/src/indexer/worker/worker.service.ts#L54
https://github.com/subquery/subql/blob/c4fd01623b3a0efcf5d9de2b559586858298b5b2/packages/node/src/utils/substrate.ts#L326

stwiname (approved):

PR #1495, opened 2023-01-21
Arranges commands in alphabetical order

seandotau:


stwiname (approved):
@seandotau is this something that you want done to the query service as well? 

https://github.com/subquery/subql/blob/main/packages/query/src/yargs.ts

PR #1504, opened 2023-01-31
Check for target block sync on endpoint health checks

DeveloperInProgress:


packages/node-core/src/meta/health.service.ts
```diff
@@ -60,7 +60,14 @@ export class HealthService {
 
   getHealth(): void {
     if (this.recordBlockTimestamp && Date.now() - this.recordBlockTimestamp > this.blockTime * 10) {
-      throw new Error('Endpoint is not healthy');
+      //check if current processing height reached target height
+      if (
+        this.recordBlockHeight &&
+        this.currentProcessingHeight &&
+        this.recordBlockHeight !== this.currentProcessingHeight
```
stwiname, line 67:
```suggestion
        this.recordBlockHeight <= this.currentProcessingHeight
```

DeveloperInProgress:
> @DeveloperInProgress have you tested this to see that health checks are improved?

Yes tested them, I don't get any health check errors after target height has been reached

stwiname:
I think this was the actual problem https://github.com/subquery/subql-ethereum/pull/39. This will be back ported

PR #1506, opened 2023-02-01
add SmartBatchService and BlockSizeBuffer

DeveloperInProgress:


packages/node/src/indexer/fetch.service.ts
```diff
@@ -80,6 +81,84 @@ function callFilterToQueryEntry(
...
+  average() {
+    if (this.buffer.length === 0) {
+      throw new Error('No block sizes to average');
+    }
+
+    let sum = 0;
+    for (let i = 0; i < this.buffer.length; i++) {
+      sum += this.buffer[i];
+    }
+    return ~~(sum / this.capacity);
```
stwiname, line 126: As cool as `~~` is, I think its better to use `Math.floor` for readability.

stwiname (commented):
I tested this out and it quickly settled on a batch size of 2 then never changed.

It would be good to see some testing outputs and how they change with different memory restrictions (e.g. docker VM settings).

I think this is a good approach to keep working with, but it would be good to base it on actual memory rather than a hard coded value. I do think a better approach is to go off free/available memory rather than the size of the blocks as there may be other memory usage that is proportional to the batch size

stwiname (commented):
It would be great to extend this to worker-block-dispatcher. We should be able to get the heap sizes for each worker and even assign different blocks to workers with more available memory

packages/node/src/indexer/worker/worker.ts
```diff
@@ -117,6 +130,8 @@ registerWorker({
   getStatus,
   syncRuntimeService,
   getSpecFromMap,
+  getMemoryLeft,
```
stwiname, line 133: We should be able to get this directly from the worker https://nodejs.org/api/worker_threads.html#workergetheapsnapshotoptions

packages/node/src/indexer/blockDispatcher/worker-block-dispatcher.service.ts
```diff
@@ -158,6 +167,20 @@ export class WorkerBlockDispatcherService
...
+
+      if(batchSize === 0) {
+        return this.enqueueBlocks(cleanedBlocks);
+      }
+
+      if(batchSize < cleanedBlocks.length) {
+        cleanedBlocks.slice(0, batchSize).map((height) => this.enqueueBlock(height, workerIdx));
+        this.latestBufferedHeight = cleanedBlocks[batchSize-1];
+        return this.enqueueBlocks(cleanedBlocks.slice(batchSize));
+      }
```
stwiname, line 182: Here we're discarding valuable data and will end up duplicating dictionary queries.

Instead I think we could update getNextWorkerIndex (or some other function) to get a worker with available memory

packages/node/src/indexer/fetch.service.ts
```diff
@@ -80,6 +85,97 @@ function callFilterToQueryEntry(
...
 
+class BlockSizeBuffer extends Queue<number> {
+  
+  constructor(capacity: number) {
+    super(capacity);
+  }
+
+  putMany(items: number[]): void {
+    if (this.capacity && items.length > this.freeSpace) {
+      this.takeMany(items.length);
```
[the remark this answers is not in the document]
DeveloperInProgress: there is a capacity to this queue and we only need the sizes of the latest blocks to compute the moving average, so we remove the old ones while adding new ones

packages/node/src/indexer/worker/worker.ts
DeveloperInProgress: snapshots are huge, they could affect the performance.

packages/node/src/indexer/blockDispatcher/worker-block-dispatcher.service.ts
```diff
@@ -194,8 +219,15 @@ export class WorkerBlockDispatcherService
         if (syncedDictionary) {
           this.syncWorkerRuntimes();
         }
+
+        if ((await worker.getMemoryLeft()) < 256) {
```
stwiname, line 223: This check shouldn't be necessary. `worker.waitForWorkerHeap(256)` should resolve instantly if memory is available

packages/node/src/utils/blockSizeBuffer.ts
```diff
@@ -0,0 +1,34 @@
...
+    super(capacity);
+  }
+
+  putMany(items: number[]): void {
+    if (items.length > this.capacity) {
+      throw new Error(`Queue out of capacity`);
+    }
+
+    if (this.capacity && items.length > this.freeSpace) {
+      this.takeMany(items.length);
```
stwiname, line 17: Why does putting items remove items from the queue? This is very implicit and items would seem to disappear

stwiname (changes requested):
We should move any new things that are blockchain agnostic to node core. (smartBatchService and blockSizeBuffer are good examples)

packages/node/src/utils/blockSizeBuffer.ts
DeveloperInProgress: we only need the latest block sizes to compute moving average, so we need to make space for the new items by removing the old ones

packages/node-core/src/utils/blockSizeBuffer.ts
```diff
@@ -0,0 +1,34 @@
...
+    super(capacity);
+  }
+
+  putMany(items: number[]): void {
+    if (items.length > this.capacity) {
+      throw new Error(`Queue out of capacity`);
+    }
+
+    if (this.capacity && items.length > this.freeSpace) {
+      this.takeMany(items.length);
```
stwiname, line 17: Following on from the other comments. This is really implicit, it should be moved to where putMany is called

packages/node/src/indexer/blockDispatcher/worker-block-dispatcher.service.ts
```diff
@@ -155,19 +169,19 @@ export class WorkerBlockDispatcherService
...
+        const workerIdx = this.getNextWorkerIndex();
+        const batchSize = Math.min(
+          cleanedBlocks.length - startIndex,
+          await this.maxBatchSize(workerIdx),
+        );
+        cleanedBlocks
+          .slice(startIndex, startIndex + batchSize)
+          .forEach((height) => this.enqueueBlock(height, workerIdx));
+        startIndex += batchSize;
+      }
```
stwiname, line 183: Im not sure this makes sense. If we add 5 blocks to worker A, then the 2nd block is really large this will cause issues. 

We should assign blocks to the worker with the most free space I think

packages/node/src/indexer/blockDispatcher/worker-block-dispatcher.service.ts
```diff
@@ -245,6 +264,12 @@ export class WorkerBlockDispatcherService
     void this.queue.put(processBlock);
   }
 
+  private async maxBatchSize(workerIdx: number): Promise<number> {
+    const memLeft = await this.workers[workerIdx].getMemoryLeft();
```
stwiname, line 268: Maybe we could use https://nodejs.org/api/worker_threads.html#workerresourcelimits instead here. It would be great to remove the async changes to enqueuBlock

packages/node/src/indexer/fetch.service.ts
```diff
@@ -499,8 +504,8 @@ export class FetchService implements OnApplicationShutdown {
               const enqueuingBlocks = batchBlocks.slice(0, maxBlockSize);
               const cleanedBatchBlocks =
                 this.filteredBlockBatch(enqueuingBlocks);
-
-              this.blockDispatcher.enqueueBlocks(
+              // eslint-disable-next-line @typescript-eslint/await-thenable
```
stwiname, line 507: Why is this now needed?

packages/node/src/indexer/blockDispatcher/worker-block-dispatcher.service.ts
DeveloperInProgress: we can modify `getNextWorkerIndex` to return the worker with most free space

packages/node/src/indexer/blockDispatcher/worker-block-dispatcher.service.ts
DeveloperInProgress: this wouldn't give us the amount of heap being used

packages/node/src/indexer/fetch.service.ts
DeveloperInProgress: enqueueBlocks of `WorkerBlocksDispacher` is Thenable while that of `BlockDispatcher` is not. So, the linter will catch that.

packages/node/src/indexer/fetch.module.ts
```diff
@@ -25,13 +25,21 @@ import { IndexerManager } from './indexer.manager';
...
     StoreService,
     ApiService,
     IndexerManager,
+    {
+      provide: SmartBatchService,
+      useFactory: (nodeConfig: NodeConfig) => {
+        return new SmartBatchService(nodeConfig.batchSize);
+      },
+      inject: [NodeConfig],
+    },
```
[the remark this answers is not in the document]
DeveloperInProgress: this doesn't work, the indexer doesn't start, it gets stuck.

stwiname (approved):

PR #1508, opened 2023-02-07
Get actual queue size after flush and add test

jiqiang90:
# Description
Current autoQueue after flush, and when we get queue size we are expecting to get 0 but return 1, this is due to it count processingTasks as 1.

This fix will force update all attributes to init value. 

Fixes # (issue)

## Type of change

## Checklist

- [x] I have tested locally

### DIFF
--- a/packages/node-core/src/utils/autoQueue.spec.ts
+++ b/packages/node-core/src/utils/autoQueue.spec.ts
@@ -19,7 +19,7 @@ describe('Auto Queue', () => {
     expect(newQ.size).toEqual(3);
 
     newQ.flush();
-
-    expect(newQ.size).toEqual(0);
+    // expect queue should be empty，this not including ongoing processing task
+    expect(newQ.queueSize).toEqual(0);
   });
 });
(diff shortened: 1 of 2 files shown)


packages/node-core/src/utils/autoQueue.spec.ts
```diff
@@ -0,0 +1,25 @@
...
+
+    newQ.put(task1);
+    newQ.put(task2);
+    newQ.put(task3);
+
+    expect(newQ.size).toEqual(3);
+
+    newQ.flush();
+
+    expect(newQ.size).toEqual(0);
```
stwiname, line 23: Maybe we can check pendingPromise is false and that we can restart the autoqueue if a new task is pushed

packages/node-core/src/utils/autoQueue.ts
```diff
@@ -169,6 +168,9 @@ export class AutoQueue<T> implements IQueue {
     // Empty the queue
     // TODO do we need to reject all promises?
     this.queue.takeAll();
+    this.pendingPromise = false;
+    this._abort = false;
+    this.processingTasks = 0;
```
stwiname, line 173: Im not too sure how this fixes the issue. If we empty the queue, then wont this happen https://github.com/subquery/subql/blob/40edb57698f35f42ae567edc087954576d4a0a4e/packages/node-core/src/utils/autoQueue.ts#L145 and these values get set anyway?

I feel like the test would still pass without these changes.

Issue #868, opened 2022-04-02
Support ssl on postgre connect


zjb0807:
https://github.com/brianc/node-postgres/blob/master/packages/pg-pool/README.md#create

darwinsubramaniam:
Hi, how is the progress on this issue?

darwinsubramaniam:
Hi @bighammerw ,
I am made some changes on the fork , but still have not tested it.
https://github.com/darwinsubramaniam/subql -> method used is using environment variable

bighammerw:
hi darwin thanks i will have a look for this issue

jamesbayly:
@darwinsubramaniam can you make a PR so we can review?

darwinsubramaniam:
> @darwinsubramaniam can you make a PR so we can review?

HI @jamesbayly , surely

darwinsubramaniam:
Hi @jamesbayly and @bighammerw , 

Could you please kindly , guide me how to test the features. I could not figure out how can test it. It would be nice if you guide me on that, future bugfix or any help required in I will be able to contribute as well.

darwinsubramaniam:
Hi @jamesbayly  and @bighammerw ,

Figure out how to run the subql-node locally using the source code, I found out that to enable SSL is not as simple as just making the ssl to be true, 

It is also should carry the ca cert at minimum to enable correct authentication to the SSL enabled postgres.  - reference https://stackoverflow.com/questions/43948920/how-to-connect-via-ssl-to-sequelize-db

making modification to the code to so USER can pass is the full path for the cert for now , later maybe can improve more on that. 
Marked the PR as Work In Progress.

PR #1513, opened 2023-02-10
[WIP] Feature/adding postgres ssl connection option

darwinsubramaniam:
# Description
Enabling the SSL connection option for Postgres.

Fixes # (issue)
#868

## Type of change

## Checklist

- [x] I have tested locally

### DIFF
--- a/packages/node-core/src/configure/NodeConfig.ts
+++ b/packages/node-core/src/configure/NodeConfig.ts
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import assert from 'assert';
-import fs from 'fs';
+import fs, {readFileSync} from 'fs';
 import path from 'path';
 import {loadFromJsonOrYaml} from '@subql/common';
 import {last} from 'lodash';
@@ -42,6 +42,9 @@ export interface IConfig {
   readonly multiChain: boolean;
   readonly reindex?: number;
   readonly unfinalizedBlocks?: boolean;
+  readonly pgCa?: string;
+  readonly pgKey?: string;
+  readonly pgCert?: string;
 }
 
 export type MinConfig = Partial<Omit<IConfig, 'subquery'>> & Pick<IConfig, 'subquery'>;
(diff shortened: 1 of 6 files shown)


packages/node-core/src/db/db.module.ts
```diff
@@ -23,6 +32,42 @@ const DEFAULT_DB_OPTION: DbOption = {
...
+USER should provide the path as environment variable
+- DB_CA_CERT: path to the CA file (Required)
+- DB_KEY_CERT: path to the KEY file (Optional)
+- DB_CLIENT_CERT: path to the CERT file (Optional)
+@return the option with ssl option
+@exception throw error if the ssl is enabled but there is not correct file path
+*/
+const dbOptionWithSSLOptions = (option: DbOption): DbOption => {
+  if (option.ssl) {
+    const ca_content = readFileSync(process.env.DB_CA_CERT).toString();
```
stwiname, line 48: Can you please add some error handling for this.

packages/node-core/src/db/db.module.ts
```diff
@@ -23,6 +32,42 @@ const DEFAULT_DB_OPTION: DbOption = {
   username: process.env.DB_USER ?? 'postgres',
   password: process.env.DB_PASS ?? 'postgres',
   database: process.env.DB_DATABASE ?? 'postgres',
+  ssl: process.env.DB_SSL === 'true' ?? false,
```
stwiname, line 35: Rather than having `DB_SSL`, we can use whether `DB_CA_CERT` is defined as the check to use ssl or not.

stwiname (changes requested):
Hi @darwinsubramaniam 

Thanks for making this PR.

I think for the connection options it would be great to do it through the args, [and by extension it adds support for env args](https://github.com/subquery/subql/blob/main/packages/node/src/yargs.ts#L9).

It would also be self documenting with the help command and add some type checking.

packages/node-core/src/db/db.module.ts
darwinsubramaniam: > 

Hi @stwiname ,
It is great suggestion, I will work on the modification as per advice and together with the error handling

darwinsubramaniam: introduced error handling by logging error, but the file reading content is common function for both  node-core and query package, so introduced the new function in common package and leveraged that for both node-core and query.

darwinsubramaniam:
Hi @stwiname ,

Made changes as per suggested , using the args and the required error  handling by logging and exit the program. 

Pending Task: 

- [ ]  Test using the ssl Postgres.

packages/node-core/src/configure/NodeConfig.ts
```diff
@@ -201,6 +204,50 @@ export class NodeConfig implements IConfig {
     return this._config.unfinalizedBlocks;
   }
 
+  get isPostgresSecureConnection(): boolean {
+    if (this._config.pgCa) {
+      return true;
+    }
+    return false;
```
stwiname, line 211:
```suggestion
    return !!this._config.pgCa;
```

stwiname (changes requested):
Nice work!

We will do some testing on our end.

PR #1519, opened 2023-02-15
`dictionary-optimisation` flag 

bz888:
# Description
Instead of using orderBy PK when using dictionary, uses blockheight (while using distinct). And adding a flag to toggle this behaviour

Fixes 
Algorand dictionary timeout

## Checklist

- [x] I have tested locally

### DIFF
--- a/packages/query/src/graphql/plugins/PgDistinctPlugin.ts
+++ b/packages/query/src/graphql/plugins/PgDistinctPlugin.ts
@@ -90,6 +90,7 @@ export const PgDistinctPlugin: Plugin = (builder) => {
             // Dependent on https://github.com/graphile/graphile-engine/pull/805
             (queryBuilder as any).distinctOn(id);
 
+            // set BlockHeight as orderBy key, as by default it uses primaryKey (This will speed up the query speed of dictionaries)
             if (argv('dictionary-optimisation')) {
               queryBuilder.setOrderIsUnique();
             }
--- a/packages/query/src/graphql/plugins/PgDistinctPlugin.ts
+++ b/packages/query/src/graphql/plugins/PgDistinctPlugin.ts
@@ -6,6 +6,7 @@ import {Build, Plugin} from 'graphile-build';
 import type {GraphQLEnumType} from 'graphql';
 import * as PgSql from 'pg-sql2';
 import {SQLNode} from 'pg-sql2';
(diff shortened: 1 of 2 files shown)


stwiname:
Some background on this. We found that dictionary queries where we order by `block_height` will then order by `id`, because `block_height` has no constraints on it. This led to a much slower query and its not necessary to order by `id` with the dictionary. 

There is an alterative solution but it would be a lot more work:
Change the db structure so entities have a pk from `block_height` + `some unique field` within the block. This would require reindexing dictionaries and updating the sdk to support pks that arent `id`

The reason for putting this behind a flag is that there could be scenarios where distinct is used but the results are ordered by a different field.


Example SQL query with id (`__local_1__."id" ASC`):
```
with __local_0__ as (select to_json((json_build_object('__identifiers'::text, json_build_array(__local_1__."id"), 'blockHeight'::text, ((__local_1__."block_height"))::text))) as "@nodes" from (select 
__local_1__.*
from "d-15687/subquery/Algorand-Dictionary"."transactions" as __local_1__
where (TRUE) and (TRUE)
order by __local_1__."block_height" ASC,__local_1__."id" ASC
limit 10
) __local_1__), __local_2__ as (select json_agg(to_json(__local_0__)) as data from __local_0__) select coalesce((select __local_2__.data from __local_2__), '[]'::json) as "data" 
```

stwiname (approved):

packages/query/src/graphql/plugins/PgDistinctPlugin.ts
```diff
@@ -88,6 +89,10 @@ export const PgDistinctPlugin: Plugin = (builder) => {
 
             // Dependent on https://github.com/graphile/graphile-engine/pull/805
             (queryBuilder as any).distinctOn(id);
+
+            if (argv('dictionary-optimisation')) {
+              queryBuilder.setOrderIsUnique();
```
stwiname, line 94: Can you please add a comment as to why we do this.

ianhe8x: Is it always good to do this, not just in case of dictionary?

bz888: what if there is a case where user wants to use `distinct` but orderedBy primaryKey

PR #1551, opened 2023-03-08
Multiple endpoints

DeveloperInProgress:


packages/node/src/indexer/api.service.ts
```diff
@@ -26,15 +29,69 @@ const NOT_SUPPORT = (name: string) => () => {
...
+    //deal with the case where average is 0
+    if (total === 0) {
+      return weights.map((weight) => 1 / this.buffers.length);
+    }
+    return weights.map((weight) => weight / total);
+  }
+}
+
 @Injectable()
 export class ApiService implements OnApplicationShutdown {
```
stwiname, line 87: I think we can simplify this a lot and remove any external api changes. Externally the api service should not show there are multiple endpoints.

Some things to change:

- Have a connection pool that adds instances when they connect and removes when they disconnect
- We don't need `attemptReconnects` to be on an interval. We can remove an interval and retry as soon as it disconnects
- Have a getter for the api that returns the right instance each time. This might mean updating other classes to use `apiService.api` rather than keeping an internal reference.
- I think we can move `fetchBlockBatches` from being a static function to a class member of api.service. We have done this with etheruem and its working well. That way we can better monitor the requests.
- Retry logic if requests fail should live in the api service.

packages/node/src/indexer/blockDispatcher/block-dispatcher.service.ts
```diff
@@ -108,6 +115,31 @@ export class BlockDispatcherService
...
+        batch,
+        specChanged ? undefined : this.runtimeService.parentSpecVersion,
+      );
+      const end = Date.now();
+      this.loadBalancer.addToBuffer(index, (end - start) / batch.length);
+      return blocks;
+    } catch (e) {
+      logger.error(e, 'Failed to fetch blocks');
+      this.apiService.attemptReconnects();
+      return this.fetchBlocksFromFirstAvailableEndpoint(batch, specChanged);
```
stwiname, line 139: This could recurse forever there needs to be a limit.

stwiname (changes requested):
For workers we should assign each worker an endpoint.

packages/cli/src/controller/migrate-controller.ts
```diff
@@ -63,7 +63,11 @@ export async function prepare(
   if (project.runner.node.name === SUBSTRATE_NODE_NAME) {
     cli.action.start('Getting network genesis hash from endpoint for Chain ID');
     try {
-      genesisHash = await getGenesisHash(projectNetwork.endpoint);
+      if (projectNetwork instanceof TerraProjectNetworkV0_3_0) {
```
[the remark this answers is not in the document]
DeveloperInProgress: yeah

packages/node/src/indexer/api.service.ts
```diff
@@ -26,24 +34,117 @@ const NOT_SUPPORT = (name: string) => () => {
...
+
+  average() {
+    if (this.items.length === 0) {
+      return 0;
+    }
+    return this.items.reduce((a, b) => a + b, 0) / this.items.length;
+  }
+}
+
+export class ApiLoadBalancer {
```
stwiname, line 54: Whats the reason for not using a round robin to determine the next endpoint to use? This over complicates things IMO.

I think we should remove this for now, it doesn't take into consideration whether the endpoint is connected

packages/node/src/indexer/api.service.ts
```diff
@@ -26,24 +34,117 @@ const NOT_SUPPORT = (name: string) => () => {
...
+  private connectionPool: Record<number, ApiPromise> = {};
+  private disconnectedApis: Record<number, ApiPromise> = {};
+  private fetchBlocksBatches = SubstrateUtil.fetchBlocksBatches;
+  private loadBalancer: ApiLoadBalancer;
+  private disconnectedApiIndices: number[] = [];
+  private endpoints: string[] = [];
   private currentBlockHash: string;
   private currentBlockNumber: number;
-  private apiOption: ApiOptions;
+  private apiOptions: ApiOptions[] = [];
```
stwiname, line 112: This is unused.

packages/node/src/indexer/api.service.ts
```diff
@@ -62,52 +163,164 @@ export class ApiService implements OnApplicationShutdown {
...
+        if (this.disconnectedApis[i]) {
+          delete this.disconnectApis[i];
+        }
+      });
+      api.on('disconnected', () => {
+        this.eventEmitter.emit(IndexerEvent.ApiConnected, { value: 0 });
+        logger.warn(`Disconnected from ${endpoint}`);
+        this.disconnectedApis[i] = api;
+        delete this.connectionPool[i];
+        this.attemptReconnects();
```
stwiname, line 196: This function can be removed and just call reconnect here.

packages/node/src/indexer/api.service.ts
```diff
@@ -26,24 +34,117 @@ const NOT_SUPPORT = (name: string) => () => {
...
+    });
+    return weightsMap;
+  }
+}
+
 @Injectable()
 export class ApiService implements OnApplicationShutdown {
-  private api: ApiPromise;
+  private connectionPool: Record<number, ApiPromise> = {};
+  private disconnectedApis: Record<number, ApiPromise> = {};
```
stwiname, line 105: I don't think we need to keep track of this. We should only need to keep reference to all and connected apis

packages/node/src/indexer/api.service.ts
```diff
@@ -205,4 +419,98 @@ export class ApiService implements OnApplicationShutdown {
...
+          index,
+          Math.ceil((end - start) / batch.length),
+        );
+        return blocks;
+      } catch (e) {
+        logger.error(
+          e,
+          `Failed to fetch blocks ${batch[0]}...${batch[batch.length - 1]}`,
+        );
+        await this.addToDisconnectedApiIndices(index);
```
stwiname, line 492: There is no idication that the endpoint is disconnected here.

packages/node/src/indexer/api.service.ts
```diff
@@ -205,4 +419,98 @@ export class ApiService implements OnApplicationShutdown {
...
+        logger.error(e, 'Failed to fetch blocks');
+        reconnectAttempts++;
+      }
+    }
+    throw new Error(
+      `Maximum number of retries (${MAX_RECONNECT_ATTEMPTS}) reached.`,
+    );
+  }
+
+  async fetchBlocks(
```
stwiname, line 453: Lets remove all the weights and connection logic here. This function should do 3 things. 

1. Get an api instance that is connected
2. Call `fetchBlocksBatches`
3. If it fails recurse up to N times

Having all this extra logic doesn't make sense to only apply to fetching block when there are other api requests as well.

packages/node/src/indexer/api.service.ts
```diff
@@ -62,52 +163,164 @@ export class ApiService implements OnApplicationShutdown {
...
+
+  async addToDisconnectedApiIndices(index: number): Promise<void> {
+    if (this.disconnectedApiIndices.includes(index)) {
+      return;
+    }
+    await this.connectionPool[index].disconnect();
+    this.disconnectedApiIndices.push(index);
+  }
+
+  getApi(index?: number): ApiPromise {
```
stwiname, line 288:
```suggestion
  get api(): ApiPromise {
```

Outside of this class there should be no concept of more than one api instance

packages/node/src/indexer/api.service.ts
```diff
@@ -26,24 +34,117 @@ const NOT_SUPPORT = (name: string) => () => {
...
+
 @Injectable()
 export class ApiService implements OnApplicationShutdown {
-  private api: ApiPromise;
+  private connectionPool: Record<number, ApiPromise> = {};
+  private disconnectedApis: Record<number, ApiPromise> = {};
+  private fetchBlocksBatches = SubstrateUtil.fetchBlocksBatches;
+  private loadBalancer: ApiLoadBalancer;
+  private disconnectedApiIndices: number[] = [];
+  private endpoints: string[] = [];
```
stwiname, line 109: Can be removed

stwiname (changes requested):
This is all very integrated into the api service and polkadot specific. It should extract new functionality out and generalise it so that it can live in node core and be easily applied to other chains.

packages/node/src/indexer/api.service.ts
```diff
@@ -454,63 +332,71 @@ export class ApiService implements OnApplicationShutdown {
...
+        `Failed to fetch blocks ${blockNums[0]}...${
+          blockNums[blockNums.length - 1]
+        }`,
+      );
+      logger.warn(
+        `disconnecting from endpoint and retrying with another endpoint`,
+      );
+      if (api.isConnected) {
+        await api.disconnect();
+      }
```
stwiname, line 364: We should never purposefully disconnect without some specific error handling

packages/node/src/indexer/api.service.ts
```diff
@@ -57,63 +71,104 @@ export class ApiService implements OnApplicationShutdown {
...
+        }
+
+        const genesisHash = api.genesisHash.toString();
+        if (this.networkMeta.genesisHash !== genesisHash) {
+          throw this.metadataMismatchError(
+            'Genesis Hash',
+            this.networkMeta.genesisHash,
+            genesisHash,
+          );
+        }
```
stwiname, line 159: I think we can just test the `genesisHash`

packages/node/src/indexer/api.service.ts
```diff
@@ -57,63 +71,104 @@ export class ApiService implements OnApplicationShutdown {
...
+        });
+      });
+      api.on('disconnected', () => {
+        this.eventEmitter.emit(IndexerEvent.ApiConnected, {
+          value: 0,
+          apiIndex: i,
+          endpoint: endpoint,
+        });
+        this.connectionPoolService.handleApiDisconnects(i, endpoint);
+      });
```
stwiname, line 110: Can we run this in parallel to speed up connecting

packages/node/src/indexer/fetch.service.ts
```diff
@@ -118,8 +118,8 @@ export class FetchService implements OnApplicationShutdown {
     this.isShutdown = true;
   }
 
-  get api(): ApiPromise {
-    return this.apiService.getApi();
+  api(index: number): ApiPromise {
```
stwiname, line 121:
```suggestion
  api(): ApiPromise {
```

packages/node-core/src/indexer/connectionPool.service.ts
```diff
@@ -0,0 +1,96 @@
...
+
+  get api(): T {
+    const index = this.getNextConnectedApiIndex();
+    if (index === -1) {
+      throw new Error('No connected api');
+    }
+    return this.connectionPool[index];
+  }
+
+  get firstConnectedApi(): T {
```
stwiname, line 56: This seems to be unused, can be removed along with `getNextConnectedApiIndex`

packages/node-core/src/indexer/connectionPool.service.ts
```diff
@@ -0,0 +1,96 @@
...
+      return -1;
+    }
+    return toNumber(Object.keys(this.connectionPool)[0]);
+  }
+
+  get numConnections(): number {
+    return Object.keys(this.connectionPool).length;
+  }
+
+  async handleApiDisconnects(apiIndex: number, endpoint: string): Promise<void> {
```
stwiname, line 86: Maybe its worth adding in here a check that we have at least some connections. It would be worth testing what happens if they all disconnect (turn off your internet)

stwiname (commented):
This is looking really good.

There are a few places that need cleaning up (logs and commented code)

packages/node/src/configure/configure.module.ts
```diff
@@ -108,7 +108,7 @@ export class ConfigureModule {
         config.subquery,
         omitBy<SubstrateProjectNetworkConfig>(
           {
-            endpoint: config.networkEndpoint,
+            endpoint: config.networkEndpoints,
```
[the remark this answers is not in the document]
DeveloperInProgress: it's `endpoint` (type is inherited from `@subql/common` module)

packages/node/src/indexer/blockDispatcher/worker-block-dispatcher.service.ts
```diff
@@ -214,7 +214,13 @@ export class WorkerBlockDispatcherService
 
         const start = new Date();
         await memoryLock.acquire();
-        await worker.fetchBlock(height, blockSpecVersion);
+        try {
+          await worker.fetchBlock(height, blockSpecVersion);
+        } catch (e) {
+          logger.debug(`Failed to fetch block ${height}: ${e.message}`);
+          memoryLock.release();
+          return processBlock();
```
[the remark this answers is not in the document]
DeveloperInProgress: this looks unnecessary after all, there is a try-catch inside `worker.fetchBlock()` and it wouldn't throw an error.

packages/node-core/src/indexer/connectionPool.service.ts
DeveloperInProgress: it retries to fetch about 5 times and exits after that

stwiname:
This command is failing because it thinks no endpoint is provided 
`yarn start:dev -f ipfs://QmcakQw27nvSkFg63qkS8ksLS8GAQTDUqM4i17ohVVAwQ6 --network-endpoint="wss://shiden.api.onfinality.io/public-ws" --batch-size=30`

It should still work as we cant break compatibility

DeveloperInProgress:
> Is there a way to specify multiple endpoints in the cli? e.g `--network-endpoint wss://shiden-rpc.dwellir.com wss://shiden.api.onfinality.io/public-ws` or `--network-endpoint wss://shiden-rpc.dwellir.com --network-endpoint wss://shiden.api.onfinality.io/public-ws`

@stwiname  `--network-endpoint wss://shiden-rpc.dwellir.com --network-endpoint wss://shiden.api.onfinality.io/public-ws` works

stwiname (approved):

PR #1556, opened 2023-03-09
add missing flag for indexCountLimint

bz888:
# Description

Fixes # (issue)

## Type of change

add missing CLI flag for `subql-node`

## Checklist

- [x] I have tested locally

### DIFF
--- a/packages/node/src/yargs.ts
+++ b/packages/node/src/yargs.ts
@@ -232,4 +232,9 @@ export const yargsOptions = yargs(hideBin(process.argv))
         'Postgres client certificate - Path to client certificate e.g /path/to/client-certificates/postgresql.crt',
       type: 'string',
     },
+    'index-count-limit': {
+      demandOption: false,
+      describe: 'indexCountLimit',
+      type: 'number',
+    },
   });


packages/node/src/yargs.ts
```diff
@@ -232,4 +232,9 @@ export const yargsOptions = yargs(hideBin(process.argv))
         'Postgres client certificate - Path to client certificate e.g /path/to/client-certificates/postgresql.crt',
       type: 'string',
     },
+    'index-count-limit': {
```
stwiname, line 235: Does this need to be consumed somewhere?

jiqiang90:
We don't need to create an another flag for this configuration. 

This can be configured use user configuration file. Follow this step.

1. Create a configuration file, name `myConfigure.json` (also support in .yaml)
2. Inside configuration file: 

```json
{
  "indexCountLimit":20
}

```

3. When start the node, pass argument `-c` and point to path of this configuration file. 

It should allow user create up to 20 indexes on each entity.  More configurations can be find here https://github.com/subquery/subql/blob/5ef2e328ea848af5066a1f15665db2e8ac42894e/packages/node-core/src/configure/NodeConfig.ts#L15

We should update document to for how to user custom configuration file, and this is support for user to run from local. 
In host service we still use default configuration to protect DB performance. (This could be open to user in future)

jiqiang90:
Close this now, user can set though config

PR #1563, opened 2023-03-14
Complete cached store methods

jiqiang90:
# Description

Fixes # (issue)

## Type of change

## Checklist

- [ ] I have tested locally

### DIFF
--- a/packages/node-core/src/indexer/storeCache.service.ts
+++ b/packages/node-core/src/indexer/storeCache.service.ts
@@ -3,7 +3,7 @@
 
 import assert from 'assert';
 import {Injectable} from '@nestjs/common';
-import { flatten, isEqual, includes} from 'lodash';
+import {flatten, isEqual, includes} from 'lodash';
 import {CreationAttributes, Model, ModelStatic, Op, Sequelize, Transaction} from 'sequelize';
 import {CountOptions} from 'sequelize/types/model';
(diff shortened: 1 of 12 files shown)


packages/node-core/src/indexer/storeCache.service.ts
```diff
@@ -199,22 +348,62 @@ class CachedModel<
...
+            const latestData = model.getLatest().data;
+            unifiedIds.push(latestData.id);
+            return latestData;
+          }
+        })
+      );
+      // No need search further
+      if (findOne && joinedData.length !== 0) {
+        return joinedData;
+      }
```
stwiname, line 367: I think the getCache has to remain in sync with the setCache so we should just be able to use the get cache here?

packages/node-core/src/indexer/storeCache.service.ts
stwiname, line 1: Is it worth breaking this up into a file for each class?

packages/node-core/src/indexer/storeCache.service.ts
jiqiang90: Yes, this make sense. getCache actually keep all latest record for each entity, search from getCache will be enough.
Also bulkCreate, bulkUpdate, set all sync with getCache.

stwiname (approved):

PR #1565, opened 2023-03-15
Implement cache for metadata entity

stwiname:


packages/node-core/src/indexer/storeCache/cacheModel.ts
stwiname, line 1: The changes in this file seemed like bugs

jiqiang90 (approved):

PR #1568, opened 2023-03-17
Lfu get cache

jiqiang90:
# Description

Fixes # (issue)

## Type of change

## Checklist

- [ ] I have tested locally

### DIFF
--- a/packages/node-core/package.json
+++ b/packages/node-core/package.json
@@ -29,7 +29,7 @@
     "@subql/x-merkle-mountain-range": "^2.0.0-0.1.2",
     "@willsoto/nestjs-prometheus": "^4.4.0",
     "lodash": "^4.17.21",
-    "node-lfu-cache": "5.0.0",
+    "lru-cache": "8.0.4",
     "prom-client": "^14.0.1",
     "sequelize": "6.28.0",
     "vm2": "^3.9.9",
--- a/packages/node-core/package.json
+++ b/packages/node-core/package.json
@@ -29,6 +29,7 @@
     "@subql/x-merkle-mountain-range": "^2.0.0-0.1.2",
     "@willsoto/nestjs-prometheus": "^4.4.0",
     "lodash": "^4.17.21",
+    "node-lfu-cache": "5.0.0",
     "prom-client": "^14.0.1",
     "sequelize": "6.28.0",
     "vm2": "^3.9.9",
--- a/packages/node-core/src/indexer/storeCache/cacheModel.ts
+++ b/packages/node-core/src/indexer/storeCache/cacheModel.ts
@@ -13,13 +13,13 @@ import {
   SetData,
   ICachedModel,
   SetValueModel,
-  GetDataModel,
-  LfuOptions,
+  GetData,
 } from './types';
 
-const getCacheOptions: LfuOptions = {
+const getCacheOptions = {
   max: 500,
-  maxAge: 1000 * 60 * 60, //in ms
+  ttl: 1000 * 60 * 60, // in ms
+  updateAgeOnGet: true, // we want to keep most used record in cache longer
 };
 
 export class CachedModel<
@@ -27,7 +27,7 @@ export class CachedModel<
 > implements ICachedModel<T>, ICachedModelControl<T>
 {
   // Null value indicates its not defined in the db
-  private getCache = new GetDataModel<T>(getCacheOptions);
+  private getCache = new GetData<T>(getCacheOptions);
 
   private setCache: SetData<T> = {};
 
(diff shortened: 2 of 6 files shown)


packages/node-core/src/indexer/storeCache/cacheModel.ts
```diff
@@ -38,26 +44,30 @@ export class CachedModel<
...
-      const record = await this.model.findOne({
-        // https://github.com/sequelize/sequelize/issues/15179
-        where: {id} as any,
-        transaction: tx,
-      });
-
-      this.getCache[id] = {
-        data: record?.toJSON<T>(),
-      };
+    if (this.getCache.get(id) === undefined) {
```
stwiname, line 55:
```suggestion
    if (!this.getCache.has(id)) {
```

packages/node-core/src/indexer/storeCache/cacheModel.ts
```diff
@@ -67,22 +77,25 @@ export class CachedModel<
...
-        // then return all from cache [30,80]
-        return cachedData.slice(options.offset, options.offset + options.limit);
-      } else if (cachedData.length < options.offset + options.limit) {
-        // example cache length 66, offset is 30, limit is 50
-        // then return [30,66] from cache, set new limit and join record from db
-        cachedData = cachedData.slice(options.offset, cachedData.length);
-        options.limit = options.limit - (cachedData.length - options.offset);
-      }
+    if (options?.offset === undefined) {
+      options.offset = 0;
```
stwiname, line 81: `options` could be undefined.

Seems to be other cases in this function that should also check for options being defined

packages/node-core/src/indexer/storeCache/cacheModel.ts
```diff
@@ -254,14 +269,34 @@ export class CachedModel<
...
+        }
+      });
+      // No need search further
+      if (findOne && joinedData.length !== 0) {
+        return joinedData;
       }
-    });
+    }
+    if (this.getCache.length !== 0) {
+      Object.entries(this.getCache).map(([, getValue]) => {
```
stwiname, line 289: How does this work with the cache state? Because every item in the cache is going to be read from.

packages/node-core/src/indexer/storeCache/cacheModel.ts
```diff
@@ -254,14 +269,34 @@ export class CachedModel<
...
+          joinedData.push(latestData);
+        }
+      });
+      // No need search further
+      if (findOne && joinedData.length !== 0) {
+        return joinedData;
       }
-    });
+    }
+    if (this.getCache.length !== 0) {
```
stwiname, line 288: This is unnecessary, map will iterate 0 times if its empty

packages/node-core/src/indexer/storeCache/cacheModel.ts
```diff
@@ -254,14 +269,34 @@ export class CachedModel<
...
-  private getByFieldFromCache(field: keyof T, value: T[keyof T] | T[keyof T][]): T[] {
-    // As getCache always synced the latest record with setCache, from `set` method
-    // We only set getCache here, otherwise we need to search setCache.
-    return Object.entries(this.getCache).map(([, getValue]) => {
-      if ((Array.isArray(value) && includes(value, getValue.data[field])) || isEqual(getValue.data[field], value)) {
-        return getValue.data;
+  private getByFieldFromCache(field: keyof T, value: T[keyof T] | T[keyof T][], findOne?: boolean): T[] {
+    const joinedData: T[] = [];
+    const unifiedIds: string[] = [];
+    if (Object.keys(this.setCache).length !== 0) {
```
stwiname, line 275: This is unnecessary, map will iterate 0 times if its empty

packages/node-core/package.json
```diff
@@ -29,6 +29,7 @@
     "@subql/x-merkle-mountain-range": "^2.0.0-0.1.2",
     "@willsoto/nestjs-prometheus": "^4.4.0",
     "lodash": "^4.17.21",
+    "node-lfu-cache": "5.0.0",
```
stwiname, line 32: What made you choose this version? It seems to be no longer maintained and is based off https://github.com/isaacs/node-lru-cache and this one has TS support as well

packages/node-core/src/indexer/storeCache/cacheModel.ts
jiqiang90: option will be defined here, I think it is best to move this code together

https://github.com/subquery/subql/blob/b9a9c860f45ccb7a25ec793a66db31932bcbb020/packages/node-core/src/indexer/store.service.ts#L721

packages/node-core/src/indexer/storeCache/cacheModel.ts
```diff
@@ -234,7 +236,7 @@ export class CachedModel<
   }
 
   clear(): void {
-    this.getCache = {};
+    this.getCache.clear();
```
stwiname, line 239: Is this something we still want? We want the cache to persist across blocks right? I think clear would get called every time we flush

PR #1569, opened 2023-03-20
generate sourcemap for projects

### LINKED ISSUE #1276: Improved Error Handling and Display
### Description

When we have an error in the mapping function, all we get is a console log printing what mapping function failed. When this is a complicated or big function, it's really hard to identify the bug

telegram-cloud-photo-size-5-6087141313809068407-y

Support sourcemaps for mapping handlers

DeveloperInProgress:
fixes #1276


packages/node-core/src/indexer/sandbox.ts
```diff
@@ -1,10 +1,13 @@
 // Copyright 2020-2022 OnFinality Limited authors & contributors
 // SPDX-License-Identifier: Apache-2.0
 
+import {existsSync, readFileSync} from 'fs';
 import path from 'path';
 import {Store} from '@subql/types';
 import {levelFilter} from '@subql/utils';
 import {merge} from 'lodash';
+import {SourceMapConsumer, NullableMappedPosition} from 'source-map';
+import {decode} from 'vlq';
```
stwiname, line 10: Seems to be unused

jiqiang90 (approved):

packages/node-core/src/indexer/sandbox.ts
```diff
@@ -46,11 +51,65 @@ export class Sandbox extends NodeVM {
         },
       })
     );
+    this.root = option.root;
+
+    const filename = 'dist/index.js';
```
stwiname, line 56: I don't think this will work with published packages. Can we test from IPFS please

packages/node-core/src/indexer/sandbox.ts
```diff
@@ -46,11 +51,65 @@ export class Sandbox extends NodeVM {
...
+    if (existsSync(sourceMapPath)) {
+      this.sourceMap = this.decodeSourceMap(sourceMapPath);
+    }
   }
 
   async runTimeout<T = unknown>(duration: number): Promise<T> {
     return timeout(this.run(this.script), duration);
   }
+
+  async convertStack(stackTrace: any) {
```
stwiname, line 68:
```suggestion
  private async convertStack(stackTrace: string): Promise<string> {
```

packages/node-core/src/indexer/sandbox.ts
```diff
@@ -46,11 +51,65 @@ export class Sandbox extends NodeVM {
...
+
+    return json;
+  }
+
+  async findLineInfo(
+    sourcemap: any,
+    compiledLineNumber: number,
+    compiledColumnNumber: number
+  ): Promise<NullableMappedPosition> {
+    const consumer = await new SourceMapConsumer(sourcemap);
```
stwiname, line 108:
```suggestion
    const consumer = new SourceMapConsumer(sourcemap);
```

packages/node-core/src/indexer/sandbox.ts
DeveloperInProgress: `Error.stack` is of `any` type and I need to retain it for RegEx `matchAll`

packages/node-core/src/indexer/sandbox.ts
DeveloperInProgress: the constructor seems to return a promise: https://github.com/mozilla/source-map/blob/4e304db915cecd42505feb007766a034395c6eb7/lib/source-map-consumer.js#L18

stwiname (approved):

packages/node-core/src/indexer/sandbox.ts
```diff
@@ -75,7 +75,7 @@ export class Sandbox extends NodeVM {
 
     const entryFile = last(this.entry.split('/'));
     const regex = new RegExp(`${entryFile.split('.')[0]}.${entryFile.split('.')[1]}:([0-9]+):([0-9]+)`, 'gi');
-    const matches = [...stackTrace.matchAll(regex)];
+    const matches = [...(stackTrace as any).matchAll(regex)];
```
stwiname, line 78: Is string the wrong type if we have to cast to any?

stwiname (approved):

PR #1572, opened 2023-03-21
remove blocking in process queueing

DeveloperInProgress:


packages/node/src/indexer/blockDispatcher/worker-block-dispatcher.service.ts
```diff
@@ -176,9 +176,11 @@ export class WorkerBlockDispatcherService
           cleanedBlocks.length - startIndex,
           await this.maxBatchSize(workerIdx),
         );
-        cleanedBlocks
-          .slice(startIndex, startIndex + batchSize)
-          .forEach((height) => this.enqueueBlock(height, workerIdx));
+        Promise.all(
```
stwiname, line 179: Missing await

packages/node/src/indexer/blockDispatcher/worker-block-dispatcher.service.ts
```diff
@@ -200,11 +202,14 @@ export class WorkerBlockDispatcherService
     // Used to compare before and after as a way to check if queue was flushed
     const bufferedHeight = this.latestBufferedHeight;
 
+    // get SpecVersion from main runtime service
+    const { blockSpecVersion, syncedDictionary } =
```
stwiname, line 206: Because we are calling `enqueueBlock` from a promise.all will this lead for fetching the same spec version more than once?

packages/node/src/indexer/blockDispatcher/worker-block-dispatcher.service.ts
```diff
@@ -214,7 +219,7 @@ export class WorkerBlockDispatcherService
 
         const start = new Date();
         await memoryLock.acquire();
-        await worker.fetchBlock(height, blockSpecVersion);
+        await pendingBlock;
```
stwiname, line 222: Doesn't moving this call break the memory lock?

DeveloperInProgress: I think locking while fetching blocks will limit the number of fetches to 1 at a time. instead, we can just lock while freeing up memory. Before fetching blocks we call `memoryLock.waitForUnlock()`. 

Also, we can change the `waitForWorkerBatchSize`  function to only lock if memory is unavailable, right now it locks everytime it is called.

packages/node/src/indexer/blockDispatcher/worker-block-dispatcher.service.ts
DeveloperInProgress: possible when runtime service is in the process of syncing with dictionary

packages/node/src/indexer/worker/worker.service.ts
```diff
@@ -76,6 +85,7 @@ export class WorkerService {
         }
 
         const block = this.fetchedBlocks[height];
+        this._numOfFetchedBlocks++;
```
stwiname, line 88: This never decreases

jiqiang90 (approved):

stwiname (approved):

PR #1577, opened 2023-03-22
Update metadata when dictionary skips large number of blocks

stwiname:
Force flush cache when the dictionary skips a large number of blocks. 

This will update metadata with latest block heights

### DIFF
--- a/packages/node/src/indexer/fetch.service.ts
+++ b/packages/node/src/indexer/fetch.service.ts
@@ -96,7 +96,8 @@ export class FetchService implements OnApplicationShutdown {
     private apiService: ApiService,
     private nodeConfig: NodeConfig,
     @Inject('ISubqueryProject') private project: SubqueryProject,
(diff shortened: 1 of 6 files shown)


packages/node/src/indexer/fetch.service.ts
```diff
@@ -96,7 +96,8 @@ export class FetchService implements OnApplicationShutdown {
     private apiService: ApiService,
     private nodeConfig: NodeConfig,
     @Inject('ISubqueryProject') private project: SubqueryProject,
-    @Inject('IBlockDispatcher') private blockDispatcher: ISubstrateBlockDispatcher,
+    @Inject('IBlockDispatcher')
+    private blockDispatcher: ISubstrateBlockDispatcher,
```
jiqiang90, line 100: This been move to a new line, is this intensional?

jiqiang90 (approved):

packages/node/src/indexer/fetch.service.ts
stwiname: This would be prettier/eslint

PR #1584, opened 2023-03-27
Testing Framework

### LINKED ISSUE #1249: Testing Framework
### Goal

A testing framework to allow users to test their projects filters and handler functions without having to index the project until it breaks. The testing framework will allow users to run their handlers and filters on known blocks with relevant data.

### Intended usage

User should be able to run tests via a subcommand to the node: `subql-node test -f ./path/to/project.yaml`

A testing function will be provided to users to help implement their tests with the following interface:

```ts
subqlTest(config: {
    handler: HandlerFunction,
    blockHeight: number,
    dependentEntities: Entity[],
    expectedEntities: Entity[],
    name?: string
})
```

The test functions will be in a specific file and the `subqlTest` function will globally register itself so the test runner can find them.

### Requirements

* The tests should run as close to the node as possible, so that running the project has as little difference as possible
* Tests should be independent of each other. No data should carry between tests.
* Tests should check that a matching filter is defined in the project.yaml. And throw an error if not found
* The expected entities provided by the user should be deeply equal to what is in the store

### Post MVP features:

Future features will largely be driven by user feedback.

* Support for dynamic datasources
* Ability to extract dependent entities from exiting indexed data

DeveloperInProgress:
# Description
A testing framework to allow users to test their projects filters and handler functions without having to index the project until it breaks. The testing framework will allow users to run their handlers and filters on known blocks with relevant data.

Fixes #1249 

## Type of change

## Checklist

- [ ] I have tested locally

Example project with tests:
https://github.com/DeveloperInProgress/acala-subql-tests
Example output from tests:
image (3)


packages/node/src/subcommands/testing.service.ts
```diff
@@ -0,0 +1,161 @@
...
+
+    // Init entities
+    test.dependentEntities.map(async (entity) => {
+      const attrs = entity as unknown as CreationAttributes<Model>;
+      logger.info(JSON.stringify(attrs));
+      await (global as any).store.set(`${entity.name}`, entity.id, entity);
+    });
+
+    // Run handler
+    switch (test.handlerKind) {
```
stwiname, line 132: We should use existing classes where possible. Here we can use `IndexerManager.indexBlock`

packages/types/src/interfaces.ts
```diff
@@ -8,6 +8,7 @@ import {IEvent} from '@polkadot/types/types';
 
 export interface Entity {
   id: string;
+  name: string;
```
stwiname, line 11: I don't think this is practical to add a name to all entities. This will add another column to every db

packages/node-core/src/indexer/sandbox.ts
```diff
@@ -4,7 +4,9 @@
 import path from 'path';
 import {Store} from '@subql/types';
 import {levelFilter} from '@subql/utils';
+import {build} from 'esbuild';
 import {merge} from 'lodash';
+import ts, {transpileModule} from 'typescript';
```
stwiname, line 9: We shouldn't be building any TS here. The project should already be built into JS

packages/node/src/main.ts
```diff
@@ -16,4 +16,6 @@ initLogger(
 // Lazy import, to allow logger to be initialised before bootstrap()
 // As bootstrap runs services that requires logger
 const { bootstrap } = require('./init');
-void bootstrap();
+if (argv._.length === 0 || !(argv._[0] === 'test')) {
```
stwiname, line 19: See yargs and subcommands. This doesn't look right

packages/node/src/subcommands/testing.service.ts
```diff
@@ -0,0 +1,253 @@
...
+      projectPath,
+      'tsconfig.json',
+    ));
+
+    const options: RegisterOptions = {
+      compilerOptions,
+      files: true,
+    };
+
+    register(options);
```
stwiname, line 72: Im not sure why we're doing anything with TS

packages/node/src/subcommands/testing.service.ts
```diff
@@ -0,0 +1,253 @@
...
+    logger.info(`Starting test: ${test.name}`);
+
+    // Fetch block
+    logger.debug('Fetching block');
+    const [block] = await SubstrateUtil.fetchBlocksBatches(
+      this.apiService.getApi(),
+      [test.blockHeight],
+    );
+
+    // Check filters match
```
stwiname, line 135: Still todo?

packages/node/src/subcommands/testing.service.ts
```diff
@@ -0,0 +1,253 @@
...
+
+    logger.info(JSON.stringify(test.dependentEntities));
+
+    // Init entities
+    logger.debug('Initializing entities');
+    test.dependentEntities.map(async (entity) => {
+      const attrs = entity as unknown as CreationAttributes<Model>;
+      logger.info(entity.constructor.name);
+      await store.set(`${(entity as any).name}`, entity.id, entity);
+    });
```
stwiname, line 157: We should use `entity.save()` here. We might need to update the interface

packages/node/src/subcommands/testing.service.ts
```diff
@@ -0,0 +1,253 @@
...
+        script: undefined,
+      },
+      this.nodeConfig,
+    );
+
+    sandbox.freeze(store, 'store');
+
+    // Run handler
+    logger.debug('Running handler');
+    switch (handlerInfo.kind) {
```
stwiname, line 174: We should use indexer.manager here

packages/node/src/subcommands/testing.service.ts
```diff
@@ -0,0 +1,253 @@
...
+      default:
+    }
+
+    // Check expected entities
+    logger.debug('Checking expected entities');
+    let passedTests = 0;
+    let failedTests = 0;
+    for (let i = 0; i < test.expectedEntities.length; i++) {
+      const expectedEntity = test.expectedEntities[i];
+      const actualEntity = await store.get(
```
stwiname, line 201: Here we can use `expectedEntity.get(expectedEntitty.id)`

packages/node/src/subcommands/testing.service.ts
DeveloperInProgress: we don't need these anymore, deleting

packages/node/src/subcommands/testing.service.ts
DeveloperInProgress: `get` is a static method

packages/node/src/indexer/indexer.manager.ts
```diff
@@ -290,7 +290,7 @@ export class IndexerManager {
     }
   }
 
-  private async indexData<K extends SubstrateHandlerKind>(
+  async indexData<K extends SubstrateHandlerKind>(
```
stwiname, line 293: We should leave this private and use indexBlock

DeveloperInProgress: how shall we get the datasource for the test? maybe we can pass a ds filter function as an optional parameter to indexBlock that should do additional filtering on top of datasources obtained from projectService.

```
 async indexBlock(
    blockContent: BlockContent,
    runtimeVersion: RuntimeVersion,
    filterDs?: (ds: SubqlProjectDs[]) => SubqlProjectDs[]
  ) {
       
       
       if (filterDs) {
        datasources = filterDs(datasources);
      }
  }
```

packages/node/src/main.ts
```diff
@@ -16,4 +16,6 @@ initLogger(
 // Lazy import, to allow logger to be initialised before bootstrap()
 // As bootstrap runs services that requires logger
 const { bootstrap } = require('./init');
-void bootstrap();
+if (!(argv._[0] === 'test')) {
```
stwiname, line 19: Whats this change for?

packages/node/src/subcommands/testing.init.ts
```diff
@@ -0,0 +1,26 @@
...
+export async function testingInit(): Promise<void> {
+  try {
+    const app = await NestFactory.create(TestingModule, {
+      logger: new NestLogger(),
+    });
+
+    await app.init();
+    const testingService = app.get(TestingService);
+    await testingService.init();
+    // TODO call function to run tests
```
stwiname, line 19: Comment can go

packages/node/src/subcommands/testing.service.ts
```diff
@@ -0,0 +1,242 @@
...
+    this.testSandboxes.map((sandbox) => {
+      this.tests.push(...sandbox.getTests());
+    });
+  }
+
+  async run() {
+    if (this.tests?.length !== 0) {
+      await Promise.all(
+        this.tests.map(async (test, index) => {
+          await this.runTest(test, this.testSandboxes[index]);
```
stwiname, line 89:
```suggestion
          return this.runTest(test, this.testSandboxes[index]);
```

packages/node/src/subcommands/testing.service.ts
```diff
@@ -0,0 +1,242 @@
...
+    );
+
+    // Check expected entities
+    logger.debug('Checking expected entities');
+    let passedTests = 0;
+    let failedTests = 0;
+    for (let i = 0; i < test.expectedEntities.length; i++) {
+      const expectedEntity = test.expectedEntities[i];
+      const actualEntity = await store.get(
+        (expectedEntity as any)._name,
```
stwiname, line 177: I think we can update the type so we don't need any

packages/node/src/subcommands/testing.service.ts
```diff
@@ -0,0 +1,242 @@
...
+            entry: mapping.file,
+            kind: hnd.kind,
+          };
+        }
+      }
+    }
+    throw new Error(`handler ${handler} not found in the mappings`);
+  }
+
+  private getVM(
```
stwiname, line 229: Unused

packages/node/src/indexer/indexer.manager.ts
```diff
@@ -78,6 +78,7 @@ export class IndexerManager {
   async indexBlock(
     blockContent: BlockContent,
     runtimeVersion: RuntimeVersion,
+    filterDs?: (ds: SubqlProjectDs[]) => SubqlProjectDs[],
```
stwiname, line 81: Ideally we make no changes to the indexer.manager. We shouldn't be changing core functionality for tests

packages/node/src/subcommands/testing.service.ts
stwiname, line 1: We should aim for this to be in node core. There is very little here that is chain specific.

stwiname (commented):
Can you please provide some example output from running tests as well as an example project with tests

packages/node/src/main.ts
DeveloperInProgress: the first check was not necessary. Just had to check if the subcommand was `test`

DeveloperInProgress:
@stwiname  Here is an example project for running tests: https://github.com/DeveloperInProgress/acala-subql-tests

here is an example output:
image

packages/node/src/subcommands/testing.service.ts
```diff
@@ -0,0 +1,255 @@
...
+import { getAllEntitiesRelations } from '@subql/utils';
+import chalk from 'chalk';
+import Pino from 'pino';
+import { CreationAttributes, Model, Sequelize } from 'sequelize';
+import { SubqlProjectDs, SubqueryProject } from '../configure/SubqueryProject';
+import { ApiService } from '../indexer/api.service';
+import { IndexerManager } from '../indexer/indexer.manager';
+import * as SubstrateUtil from '../utils/substrate';
+
+const logger = getLogger('CLI-test-Testing');
```
stwiname, line 25:
```suggestion
const logger = getLogger('subql-testing');
```

packages/node-core/src/indexer/sandbox.ts
```diff
@@ -155,3 +155,27 @@ export class IndexerSandbox extends Sandbox {
     this.freeze(logger, 'logger');
   }
 }
+
+export class TestSandbox extends Sandbox {
```
stwiname, line 159: Do you think we need this extra class? Can we just use the IndexerSandbox class?

We're missing injecting the api here too.

packages/node/src/indexer/indexer.manager.ts
stwiname: Maybe rather than providing the projectService as a dependency we can provide the datasources here. 

It should be fairly easy to move the projectService.init elsewhere

packages/node/src/main.ts
stwiname: I don't think we're guaranteed to have test as the first argument. currently `subql-node -f /path/to/project test` is a valid command but this would fail

stwiname (commented):
We need to update CI to publish the new package

packages/node-core/src/indexer/sandbox.ts
DeveloperInProgress: this sandbox is only used for loading the tests. The tests are later run in the IndexerSandbox with the injected globals.

packages/node-core/src/indexer/indexer.manager.ts
```diff
@@ -0,0 +1,11 @@
+// Copyright 2020-2022 OnFinality Limited authors & contributors
+// SPDX-License-Identifier: Apache-2.0
+
+export abstract class IndexerManager<B, DS> {
```
stwiname, line 4: Lets make this an interface as it doesn't implement any logic

packages/node-core/src/indexer/testing.service.ts
```diff
@@ -0,0 +1,265 @@
...
+    await Promise.all(
+      this.testSandboxes.map(async (sandbox) => {
+        await sandbox.runTimeout(1000);
+      })
+    );
+
+    this.testSandboxes.map((sandbox, index) => {
+      this.tests[index] = sandbox.getTests();
+    });
+  }
```
stwiname, line 87: Could we combine this so that `runTimeout` returns the tests? Or `getTests` internally calls `runTimeout`

packages/node-core/src/indexer/testing.service.ts
```diff
@@ -0,0 +1,265 @@
...
+
+    logger.info(
+      `Test: ${test.name} completed with ${chalk.green(`${passedTests} passed`)} and ${chalk.red(
+        `${failedTests} failed`
+      )} checks`
+    );
+    await this.sequelize.dropSchema(`"${schema}"`, {
+      logging: false,
+      benchmark: false,
+    });
```
stwiname, line 217: I think we could move this into a `finally` so we can guarantee that it always happens. If there's any error checking the entities currently then this won't get called

packages/node-core/src/indexer/testing.service.ts
```diff
@@ -0,0 +1,265 @@
...
+          mapping.handlers = [handlers[i] as any];
+          return [dsCopy];
+        }
+      }
+
+      return [];
+    }
+  }
+
+  private isEqual(expected: any, actual: any): boolean {
```
stwiname, line 239: I think we can replace this with https://lodash.com/docs/4.17.15#isEqual

packages/types/src/interfaces.ts
```diff
@@ -8,6 +8,10 @@ import {IEvent} from '@polkadot/types/types';
 
 export interface Entity {
   id: string;
+  _name?: string;
+  save?: () => Promise<void>;
+
+  [key: string]: any;
```
stwiname, line 14: Can we remove this please, this will break a lot of type checking

packages/node-core/src/indexer/testing.service.ts
```diff
@@ -0,0 +1,252 @@
...
+      );
+    } catch (e) {
+      this.totalFailedTests += test.expectedEntities.length;
+      logger.warn(`Test: ${test.name} field due to runtime error`, e);
+      this.failedTestsSummary.push({
+        testName: test.name,
+        entityId: undefined,
+        entityName: undefined,
+        failedAttributes: [`Runtime Error:\n${e.stack}`],
+      });
```
stwiname, line 205: I think the runtime error should be specific to indexBlock like it was before. 

We should instead throw an error here that the test suite failed

stwiname (approved):

packages/node-core/src/indexer/testing.service.ts
```diff
@@ -0,0 +1,252 @@
...
+      return new TestSandbox(option, this.nodeConfig);
+    });
+
+    logger.info(`Found ${this.testSandboxes.length} test files`);
+
+    await Promise.all(
+      this.testSandboxes.map(async (sandbox, index) => {
+        this.tests[index] = await sandbox.getTests();
+      })
+    );
```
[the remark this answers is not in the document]
DeveloperInProgress: Since we do this parallely, we need to map the tests to the right sandbox index

stwiname:
Can we handle no tests better please 

This is the current error: `2023-04-19T03:02:20.048Z <CLI-Testing> ERROR Testing failed Error: ENOENT: no such file or directory, scandir '/Users/scotttwiname/Projects/subql-starter/Polkadot/Polkadot-starter/dist/test'`

I think we can just log that no tests are found and exit cleanly

packages/node/src/subcommands/testing.init.ts
```diff
@@ -0,0 +1,34 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+import { NestFactory } from '@nestjs/core';
+import { getLogger, NestLogger } from '@subql/node-core';
+import { ApiService } from '../indexer/api.service';
+import { ProjectService } from '../indexer/project.service';
+import { TestingModule } from './testing.module';
+import { TestingService } from './testing.service';
+
+const logger = getLogger('CLI-Testing');
```
stwiname, line 11:
```suggestion
const logger = getLogger('Testing');
```

PR #1597, opened 2023-04-03
Add support for non-historical indexing with cache

stwiname:
Changes:
- Alter constrainsts to be `DEFERRABLE INITIALLY IMMEDIATE`
- On transaction start set deferrable if historical is disabled

### DIFF
--- a/packages/node-core/src/indexer/store.service.ts
+++ b/packages/node-core/src/indexer/store.service.ts
@@ -306,8 +306,7 @@ export class StoreService {
           });
           extraQueries.push(
             commentConstraintQuery(`"${schema}"."${rel.target.tableName}"`, fkConstraint, tags),
-            createUniqueIndexQuery(schema, relatedModel.tableName, relation.foreignKey),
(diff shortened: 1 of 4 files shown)


packages/node-core/src/indexer/store.service.ts
```diff
@@ -303,7 +306,8 @@ export class StoreService {
           });
           extraQueries.push(
             commentConstraintQuery(`"${schema}"."${rel.target.tableName}"`, fkConstraint, tags),
-            createUniqueIndexQuery(schema, relatedModel.tableName, relation.foreignKey)
+            createUniqueIndexQuery(schema, relatedModel.tableName, relation.foreignKey),
+            constraintDeferrableQuery(model.getTableName().toString(), fkConstraint)
```
jiqiang90, line 310: I think this should work without `constraintDeferrableQuery`.

It should pick up the same fkConstraint from `belongTo`, for both `hasOne` and `hasMany`

image

image

jiqiang90 (approved):

PR #1602, opened 2023-04-05
improve blockSize function

DeveloperInProgress:
# Description
The maximum length of a string in JavaScript is fixed at 2^53-1. JSON.stringify() will throw RangeError on large blocks that could exceed this size when serialized. `blockSize` function must be updated to deal with the block one property at a time 

## Type of change

## Checklist

- [ ] I have tested locally

### DIFF
--- a/packages/node-core/src/indexer/smartBatch.service.ts
+++ b/packages/node-core/src/indexer/smartBatch.service.ts
@@ -32,10 +32,49 @@ export class SmartBatchService {
     blocks.forEach((block) => this.blockSizeBuffer.put(this.blockSize(block)));
   }
 
-  blockSize(block: any): number {
-    return Buffer.byteLength(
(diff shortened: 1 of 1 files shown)


packages/node-core/src/indexer/smartBatch.service.ts
```diff
@@ -32,10 +32,49 @@ export class SmartBatchService {
     blocks.forEach((block) => this.blockSizeBuffer.put(this.blockSize(block)));
   }
 
-  blockSize(block: any): number {
-    return Buffer.byteLength(
-      JSON.stringify(block, (key, value) => (typeof value === 'bigint' ? value.toString() : value))
-    );
+  blockSize(block: any) {
```
stwiname, line 35:
```suggestion
  blockSize(block: any): number {
```

stwiname (approved):

PR #1604, opened 2023-04-06
support cockroach db with non-historical

jiqiang90:
# Description

TODO: 

- [x] check whether we still can use bulkCreate under cacheModel, and performance different? bulkCreate seems working with non relationship entity

Fixes # (issue)

## Type of change

## Checklist

- [x] I have tested locally

### DIFF
--- a/packages/node-core/src/indexer/store.service.ts
+++ b/packages/node-core/src/indexer/store.service.ts
@@ -69,6 +69,8 @@ const logger = getLogger('store');
 const NULL_MERKEL_ROOT = hexToU8a('0x00');
 const NotifyTriggerManipulationType = [`INSERT`, `DELETE`, `UPDATE`];
 
+type RemovedIndexes = Record<string, IndexesOptions[]>;
+
 interface IndexField {
   entityName: string;
   fieldName: string;
@@ -94,6 +96,7 @@ export class StoreService {
   historical: boolean;
   private dbType: SUPPORT_DB;
   private useSubscription: boolean;
+  private removedIndexes: RemovedIndexes = {};
 
   poiModel: CachePoiModel;
   metadataModel: CacheMetadataModel;
@@ -116,8 +119,7 @@ export class StoreService {
       this.historical = false;
       logger.warn(`Historical feature is not support with ${this.dbType}`);
     }
-    this.storeCache.setHistorical(this.historical);
-    this.storeCache.setUseCockroach(this.dbType === SUPPORT_DB.cockRoach);
+    this.storeCache.init(this.historical, this.dbType === SUPPORT_DB.cockRoach);
 
     try {
       await this.syncSchema(this.schema);
@@ -258,7 +260,7 @@ export class StoreService {
       this.updateIndexesName(model.name, indexes, existedIndexes as string[]);
 
       // Update index query for cockroach db
-      this.handleCockroachIndex(schema, model.name, indexes, existedIndexes as string[], extraQueries);
+      this.beforeHandleCockroachIndex(schema, model.name, indexes, existedIndexes as string[], extraQueries);
 
       const sequelizeModel = this.sequelize.define(model.name, attributes, {
         underscored: true,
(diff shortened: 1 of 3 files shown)


packages/node-core/src/indexer/store.service.ts
```diff
@@ -117,6 +117,7 @@ export class StoreService {
       logger.warn(`Historical feature is not support with ${this.dbType}`);
     }
     this.storeCache.setHistorical(this.historical);
+    this.storeCache.setUseCockroach(this.dbType === SUPPORT_DB.cockRoach);
```
stwiname, line 120: Maybe these could be combined into an init?

packages/node-core/src/indexer/storeCache/cacheModel.ts
```diff
@@ -227,10 +228,15 @@ export class CachedModel<
       ]);
     } else {
       dbOperation = Promise.all([
-        this.model.bulkCreate(records, {
-          transaction: tx,
-          updateOnDuplicate: Object.keys(records[0]) as unknown as (keyof T)[], // TODO is this right? we want upsert behaviour
-        }),
+        // cockroach having issue with bulkCreate
```
stwiname, line 231: Can we elaborate on this? Is it just the size of the query or something else?

packages/node-core/src/indexer/store.service.ts
```diff
@@ -429,6 +436,28 @@ export class StoreService {
...
 
+  private handleCockroachIndex(
+    schema: string,
+    modelName: string,
+    indexes: IndexesOptions[],
+    existedIndexes: string[],
+    extraQueries: string[]
+  ): void {
+    // remove original indexes avoid direct create from sequelize model
+    // this will create index manually from extraQueries
```
stwiname, line 447: It would be good to explain WHY we need to do this.

stwiname (commented):
This just fixes the deferrable issue right? But we still have the issue with constraints failing because of the order of queries?

jiqiang90:
Checked, performance of bulkCreate (9200 blocks )vs upsert (9900 blocks) in flush with starter project, very similar outcome. I think we can use upsert here.

packages/node-core/src/indexer/store.service.ts
jiqiang90: > This just fixes the deferrable issue right? But we still have the issue with constraints failing because of the order of queries?

This fixes the deferrable issue. constraint with query order issue could still be an issue, test with some relation entity project under substrate appears working, will need to check again with the other network projects to confirm.

packages/node-core/src/indexer/storeCache/cacheModel.ts
jiqiang90: Unconfirmed error:
```
(node:6895) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'unique' of undefined
    at recursiveBulkCreate (/Users/jiatwork/ofn/subql/node_modules/sequelize/src/model.js:2782:19)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (internal/process/task_queues.js:95:5)
```

But I suspect this is due to we remove the indexes from model. I need to check this

jiqiang90: Confirmed this issue due to we remove indexes from model, however after model synced with db, seems like we could not added it back to model.

So I pick use `upsert` here. See https://github.com/subquery/subql/issues/1606

packages/node-core/src/indexer/store.service.ts
```diff
@@ -429,6 +441,46 @@ export class StoreService {
     });
   }
 
+  // Sequelize model will generate follow query to create hash indexes
+  // Example SQL:  CREATE INDEX "accounts_person_id" ON "polkadot-starter"."accounts" USING hash ("person_id")
+  // This will be rejected from cockroach db due to syntax error
+  // To avoid this we need to create index manually and add to extraQueries in order to create index in db
+  private beforeHandleCockroachIndex(
+    schema: string,
+    modelName: string,
```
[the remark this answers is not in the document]
jiqiang90: I think it is a bit confusing here, we are having 2 `models` here 

This model is a type of https://github.com/subquery/subql/blob/1599cec0bec9398c2173f3af73e3f35a15323b42/packages/utils/src/graphql/types.ts#L33.
And we kinda want to sorted the indexes and extra sql queries before the actual model defined. 

And we only get `getTableName()` method after the actual model `sequelizeModel` defined. https://github.com/subquery/subql/blob/1599cec0bec9398c2173f3af73e3f35a15323b42/packages/node-core/src/indexer/store.service.ts#L265

PR #1614, opened 2023-04-13
Flush store cache in operational order

### LINKED ISSUE #1610: Solution for Cockroach FK not support deferrable
### Prerequisites

* [ ] Are you running the latest version(s)?
* [ ] Did you check the [debugging guide](https://doc.subquery.network/academy/tutorials_examples/debug-projects.html#how-to-debug-a-subquery-project)?
* [ ] Did you check the [FAQs](https://doc.subquery.network/faqs/faqs.html) and [Discussions](https://github.com/subquery/subql/discussions)?
* [ ] Are you reporting to the correct repository?
* [ ] Did you search for an existing issue or pull request?

### Description

Due to cockroach db not support FK constraint set to be deferrable https://github.com/cockroachdb/cockroach/issues/31632 , we need to support cache flush entity with orders.

Propose : We can add an `operationIndex` to each records, then we extract records from each entity sort them with `operationIndex`

### Details

**Network Details:**
  * [Network]

### Steps to Reproduce

1. [First Step]
2. [Second Step]
3. [and so on...]

### Any other information

jiqiang90:
# Description

- enable with `store-flush-in-order` flag
- only work with **non-historica**l project, as historical projects has no constraints between entity, we can always use bulk method.
-  it will only sort **relational** model 's record, other non-relational model can be flush with bulk method.
-  We need to flush all records instead of flush the the latest records from model's record only, as record and its fk record could be missing during flush in operational order.

Tests:

- working with both postgres and cockroach db,  worker and non-workers

Fixes #1610

## Type of change

## Checklist

- [x] I have tested locally

### DIFF
--- a/packages/node-core/src/indexer/storeCache/cacheModel.ts
+++ b/packages/node-core/src/indexer/storeCache/cacheModel.ts
@@ -248,7 +248,8 @@ export class CachedModel<
     await dbOperation;
   }
 
-  async flushOperation(operationIndex: number, tx: Transaction, blockHeight?: number): Promise<void> {
+  // Flush relation model in operationIndex order with non-historical db
+  async flushOperation(operationIndex: number, tx: Transaction): Promise<void> {
     const removeRecordKey = Object.keys(this.removeCache).find(
       (key) => this.removeCache[key].operationIndex === operationIndex
     );
--- a/packages/node-core/src/indexer/storeCache/cacheModel.ts
+++ b/packages/node-core/src/indexer/storeCache/cacheModel.ts
@@ -1,7 +1,7 @@
 // Copyright 2020-2022 OnFinality Limited authors & contributors
 // SPDX-License-Identifier: Apache-2.0
 
-import {flatten, includes, isEqual, uniq} from 'lodash';
+import {flatten, forEach, includes, isEqual, uniq} from 'lodash';
 import {CreationAttributes, Model, ModelStatic, Op, Transaction} from 'sequelize';
 import {Fn} from 'sequelize/types/utils';
 import {NodeConfig} from '../../configure';
@@ -13,6 +13,9 @@ import {
   ICachedModel,
   SetValueModel,
   GetData,
+  IndexedFlushableRecord,
+  FilteredHeightRecords,
+  IndexedOperationActionType,
 } from './types';
 
 const getCacheOptions = {
@@ -32,6 +35,12 @@ export class CachedModel<
 
   private removeCache: Record<string, RemoveValue> = {};
 
+  private getNextStoreOperationIndex: () => number;
+
+  private storeFlushInOrder: boolean;
+
+  readonly hasAssociations: boolean;
+
   flushableRecordCounter = 0;
 
   constructor(
@@ -45,6 +54,14 @@ export class CachedModel<
       getCacheOptions.max = this.config.storeGetCacheSize;
     }
     this.getCache = new GetData<T>(getCacheOptions);
+    if (Object.keys(this.model.associations).length > 0) {
+      this.hasAssociations = true;
+    }
+  }
+
+  init(storeFlushInOrder = false, getNextStoreOperationIndex: () => number) {
+    this.storeFlushInOrder = storeFlushInOrder;
+    this.getNextStoreOperationIndex = getNextStoreOperationIndex;
   }
 
   private get historicalModel(): ModelStatic<Model<T & HistoricalModel, T & HistoricalModel>> {
@@ -146,7 +163,7 @@ export class CachedModel<
     if (this.setCache[id] === undefined) {
       this.setCache[id] = new SetValueModel();
     }
-    this.setCache[id].set(data, blockHeight);
+    this.setCache[id].set(data, blockHeight, this.storeFlushInOrder ? this.getNextStoreOperationIndex() : undefined);
     // Experimental, this means getCache keeps duplicate data from setCache,
     // we can remove this once memory is too full.
     this.getCache.set(id, data);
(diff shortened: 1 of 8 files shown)


packages/node/src/yargs.ts
```diff
@@ -201,6 +201,12 @@ export const yargsOptions = yargs(hideBin(process.argv))
         'If enabled writing data to the store is asynchronous with regards to block processing',
       type: 'boolean',
     },
+    'store-flush-in-order': {
+      demandOption: false,
+      describe:
+        'If enabled relational model will flush in original operational order, this only work in non-historical mode',
```
stwiname, line 207:
```suggestion
        'If enabled relational model will flush in original operational order, this only works in non-historical mode',
```

packages/node-core/src/indexer/storeCache/types.ts
stwiname, line 1: It might be worth moving `SetValueModel` to its own file

packages/node-core/src/indexer/storeCache/storeCache.service.ts
```diff
@@ -35,6 +37,24 @@ export class StoreCacheService implements BeforeApplicationShutdown {
   init(historical: boolean, useCockroachDb: boolean): void {
     this._useCockroachDb = useCockroachDb;
     this._historical = historical;
+    if (this.config.storeFlushInOrder) {
+      if (this._historical) {
+        logger.warn(`Flush store cache in order is not support with Historical feature,it will have no effect`);
```
stwiname, line 42:
```suggestion
        logger.warn(`Flush store cache in order is not supported with Historical feature, it will have no effect`);
```

packages/node-core/src/indexer/storeCache/storeCache.service.ts
```diff
@@ -35,6 +37,24 @@ export class StoreCacheService implements BeforeApplicationShutdown {
...
+      } else {
+        this._storeFlushInOrder = this.config.storeFlushInOrder;
+        this.resetStoreOperationIndex();
+        logger.info(`Store cache will flush in order`);
+      }
+    }
+  }
+
+  resetStoreOperationIndex(): void {
```
stwiname, line 51: Can be private

packages/node-core/src/indexer/storeCache/storeCache.service.ts
```diff
@@ -83,9 +102,25 @@ export class StoreCacheService implements BeforeApplicationShutdown {
...
+    );
+    // Handle record in operation order
+    for (const record of sortedFlushableRecords) {
+      const model = this.sequelize.model(record.entity);
+      if (record.action === IndexedOperationActionType.Set) {
+        await model.upsert(record.data, {transaction: tx});
+      } else if (record.action === IndexedOperationActionType.Remove) {
+        await model.destroy({where: {id: record.data.id} as any, transaction: tx});
+      }
+    }
```
stwiname, line 119: I think we can simplify this if we add a method to cache model. Something like `flushOperation(index: number)` then the model can handle the specific store operation and dealing with the model since we already have an instance of it there. 

Then we can have a really simple loop:

```ts
for (const i = 0; i < this._storeOperationIndex; i++) {
    // Flush operation can be a no-op if it doesn't hatve that index
    await Promise.all(this.models.map(m => m.flushOperation(i)));
}
```

I think it will also mean we can drop `IndexedOperationActionType` as the cacheModel knows what type of operation it is.

`flushOperation` should also remove the item from the cache as it has now been flushed, we don't seem to be doing that currently

packages/node-core/src/indexer/storeCache/cacheModel.ts
```diff
@@ -45,6 +54,14 @@ export class CachedModel<
       getCacheOptions.max = this.config.storeGetCacheSize;
     }
     this.getCache = new GetData<T>(getCacheOptions);
+    if (Object.keys(this.model.associations).length > 0) {
+      this.hasAssociations = true;
+    }
+  }
+
+  init(storeFlushInOrder = false, getNextStoreOperationIndex: () => number) {
```
stwiname, line 62: I think we should always keep the operation index even if we don't flush in order. It will simplify the code and doesn't have much overhead.

packages/node-core/src/indexer/storeCache/cacheModel.ts
```diff
@@ -45,6 +54,14 @@ export class CachedModel<
       getCacheOptions.max = this.config.storeGetCacheSize;
     }
     this.getCache = new GetData<T>(getCacheOptions);
+    if (Object.keys(this.model.associations).length > 0) {
```
stwiname, line 57: Nice!

packages/node-core/src/indexer/storeCache/storeCache.service.ts
```diff
@@ -95,15 +130,23 @@ export class StoreCacheService implements BeforeApplicationShutdown {
...
+      } else {
+        await Promise.all(updatableModels.map((model) => model.flush(tx, blockHeight)));
+      }
       await tx.commit();
     } catch (e) {
       logger.error(e, 'Database transaction failed');
       await tx.rollback();
       throw e;
     }
+    this.resetStoreOperationIndex();
```
stwiname, line 149: I don't think this will work. Remember that we take a snapshot (`updatableModels`) that we flush, so in the mean time items can still be added to the cache. That would mean that the storeOperationIndex could keep incrementing while we're flushing. 

We could store another number as to what index we have flushed up to.

I don't think we need to reset the index, its very unlikely there would be enough operations that we reach the limit of `number` and in that case we could just throw an error to restart the node and reset the index.

jiqiang90:
I been done some performance test with two method: for loop with models, and aggregate operations then for loop.

Here is the result with 1461 record to flushes

- for loop with models (not reset op index):  5068ms
- for loop with models (reset op index):  1008ms
- aggregate operations: 2159ms


So we still need to reset op index to optimize performance.

stwiname:
With this is the for loop starting from 0 or starting from the last flushed index?
>     * for loop with models (not reset op index):  5068ms

jiqiang90:
> With this is the for loop starting from 0 or starting from the last flushed index?
> 
> > ```
> > * for loop with models (not reset op index):  5068ms
> > ```

Flush from 0.  I think rather than set to 0,  make sense to store the last flushed index, I will make a change

stwiname (approved):

PR #1618, opened 2023-04-17
postgres based mmr DB

### LINKED ISSUE #1594: Store MMR in DB
### Description

We should provide a DB storage mechanism for MMR rather than file based. This will simplify having to manage storage and will also make creating backups simpler as it can be in the same DB as projects.

We should provide a migration path where any existing file based MMR data can be moved to the DB.

BONUS:
Remove the project Id column from the POI table as this is redundant data and instead use another field in the metadata table

DeveloperInProgress:
# Description
Provide a DB storage mechanism for MMR rather than file based. This will simplify having to manage storage and will also make creating backups simpler as it can be in the same DB as projects.

Fixes #1594 

## Type of change

## Checklist

- [ ] I have tested locally


packages/node/src/yargs.ts
```diff
@@ -147,6 +148,12 @@ export const yargsOptions = yargs(hideBin(process.argv))
       type: 'boolean',
       default: false,
     },
+    'mmr-db-store': {
```
stwiname, line 151: Let's change this to `mmr-store-type` and have it as an enum. We don't know if later on we have the option for another store type.

packages/node-core/src/indexer/postgresBasedDb.ts
```diff
@@ -0,0 +1,142 @@
...
+  id: number;
+  value: number;
+}
+
+class LeafLengthInstance extends Model<LeafLengthAttributes> implements LeafLengthAttributes {
+  id!: number;
+  value!: number;
+}
+
+export class PgBasedDB {
```
stwiname, line 33: We can implement this interface https://github.com/subquery/subql/blob/main/packages/node-core/src/merkle-mountain-range.d.ts#L18

stwiname: The names of everything need to be much clearer. `PgBasedDB` has no indication this relates to MMR, same with `KVStore`

packages/node-core/src/indexer/postgresBasedDb.ts
```diff
@@ -0,0 +1,142 @@
...
+      {
+        sequelize,
+        modelName: 'LeafLength',
+      }
+    );
+    return LeafLengthInstance;
+  }
+
+  async connect() {
+    await this.sequelize.authenticate();
```
stwiname, line 87: This should use the same connection as the rest of the DB operations, I think we should remove this from here. Same with disconnect

packages/node-core/src/indexer/postgresBasedDb.ts
```diff
@@ -0,0 +1,142 @@
...
+        key: {
+          type: DataTypes.INTEGER,
+          primaryKey: true,
+        },
+        value: {
+          type: DataTypes.BLOB,
+          allowNull: false,
+        },
+      },
+      {
```
stwiname, line 57: I think we're missing the schema here.

stwiname (approved):

DeveloperInProgress:
we need this fix for mmr migrate to work: https://github.com/subquery/merkle-mountain-range/pull/4

packages/node-core/src/indexer/postgresBasedMMRDb.ts
stwiname, line 1: I think this file can be `postgessMmrDb.ts`

packages/node/src/subcommands/mmrMigrate.module.ts
```diff
@@ -0,0 +1,32 @@
...
+
+@Module({
+  providers: [
+    {
+      provide: MMRMigrateService,
+      useFactory: (config: NodeConfig, sequelize: Sequelize) => {
+        return new MMRMigrateService(sequelize, config);
+      },
+      inject: [NodeConfig, Sequelize],
+    },
```
stwiname, line 18: Use factory is only needed if we have extra logic

```suggestion
    MMRMigrateService,
```

packages/node/src/yargs.ts
```diff
@@ -63,6 +63,23 @@ export const yargsOptions = yargs(hideBin(process.argv))
       return reindexInit(argv.targetHeight);
     },
   })
+  .command({
```
stwiname, line 66: I think we want to specify the args for this.

packages/node/src/subcommands/mmrMigrate.service.ts
```diff
@@ -0,0 +1,47 @@
...
+
+const logger = getLogger('mmr-migrate');
+
+const keccak256Hash = (...nodeValues: Uint8Array[]) =>
+  Buffer.from(keccak256(Buffer.concat(nodeValues)), 'hex');
+
+export class MMRMigrateService {
+  constructor(private sequelize: Sequelize, private nodeConfig: NodeConfig) {}
+
+  async migrate(): Promise<void> {
```
stwiname, line 23: Is it possible to make this migrate both ways. File -> Db or Db -> File

packages/node-core/src/indexer/mmr.service.ts
```diff
@@ -66,20 +72,22 @@ export class MmrService implements OnApplicationShutdown {
...
+    this._mmrDb =
+      this.nodeConfig.mmrStoreType === MmrStoreType.Postgres
+        ? await this.ensurePostgresBasedMmr()
+        : await this.ensureFileBasedMmr(this.nodeConfig.mmrPath);
     this._blockOffset = blockOffset;
-
-    // The file based database current leaf length
-    const fileBasedMmrLeafLength = await this.fileBasedMmr.getLeafLength();
+    // The mmr database current leaf length
+    const MmrLeafLength = await this.mmrDb.getLeafLength();
```
stwiname, line 81:
```suggestion
    const mmrLeafLength = await this.mmrDb.getLeafLength();
```

packages/node-core/src/indexer/postgresBasedMMRDb.ts
```diff
@@ -0,0 +1,140 @@
...
+      {
+        sequelize,
+        modelName: 'KVStore',
+        schema: schema,
+      }
+    );
+    return MMRIndexValueStoreInstance;
+  }
+
+  private initLeafLengthModel(sequelize: Sequelize, schema: string): ModelStatic<LeafLengthInstance> {
```
stwiname, line 67: It seems like overkill to have a table for a single value. It probably makes more sense to use the metadata table or use a specific key in the other table (maybe -1 as the key).

packages/node-core/src/indexer/postgresBasedMMRDb.ts
```diff
@@ -0,0 +1,140 @@
...
+          primaryKey: true,
+        },
+        value: {
+          type: DataTypes.BLOB,
+          allowNull: false,
+        },
+      },
+      {
+        sequelize,
+        modelName: 'KVStore',
```
stwiname, line 60:
```suggestion
        modelName: '_mmr',
```

packages/node/src/subcommands/mmrMigrate.service.ts
stwiname, line 1: This can live in node core.

stwiname:
I would suggest rebasing. We have changed the ts settings to enable strict

packages/node/src/subcommands/mmrMigrate.module.ts
DeveloperInProgress: > Use factory is only needed if we have extra logic

without useFactory , NodeConfig comes out undefined

packages/node-core/src/indexer/postgresMmrDb.ts
```diff
@@ -0,0 +1,92 @@
...
+    const nodes = await this.mmrIndexValueStore.findAll<MMRIndexValueStoreInstance>();
+    const nodeMap: NodeMap = {};
+    nodes.forEach((node) => {
+      nodeMap[node.key] = node.value;
+    });
+    return nodeMap;
+  }
+
+  async getLeafLength() {
+    const record = await this.mmrIndexValueStore.findByPk(-1);
```
stwiname, line 82: Lets move -1 to a const so we can explain its usage

packages/node/src/subcommands/mmrMigrate.module.ts
stwiname:
```suggestion
    NodeConfig,
    MMRMigrateService,
```

packages/node/src/yargs.ts
```diff
@@ -63,6 +63,30 @@ export const yargsOptions = yargs(hideBin(process.argv))
...
+    command: 'mmr-migrate',
+    describe: 'Migrate MMR data from storage file to postgres DB',
+    builder: (yargs) =>
+      yargs.options('direction', {
```
stwiname, line 70: I think we might need more options. Like `mmr-path`

stwiname (commented):
I'd like to see some error handling for both indexing and the subcommand to handle if the MMR db doesnt exist.

packages/node-core/src/indexer/mmrMigrate.service.ts
```diff
@@ -0,0 +1,76 @@
...
+
+import {existsSync} from 'fs';
+import {DEFAULT_WORD_SIZE} from '@subql/common';
+import {NodeConfig, getExistingProjectSchema, getLogger, PgBasedMMRDB} from '@subql/node-core';
+import {FileBasedDb} from '@subql/x-merkle-mountain-range';
+import {Logging, Sequelize} from 'sequelize';
+
+const logger = getLogger('mmr-migrate');
+
+const DEFAULT_DB_SCHEMA = 'public';
```
jiqiang90, line 12: https://github.com/subquery/subql/pull/1648

We just deprecated localMode, user can always save this in configured schema

jiqiang90 (approved):

packages/node-core/src/indexer/mmrMigrate.service.ts
```diff
@@ -0,0 +1,131 @@
...
+        }
+      }
+
+      let fileBasedMMRDb: FileBasedDb;
+
+      if (existsSync(this.nodeConfig.mmrPath)) {
+        fileBasedMMRDb = await FileBasedDb.open(this.nodeConfig.mmrPath);
+      } else {
+        fileBasedMMRDb = await FileBasedDb.create(this.nodeConfig.mmrPath, DEFAULT_WORD_SIZE);
+      }
```
stwiname, line 82: Can use a ternary here.

packages/node-core/src/indexer/mmrMigrate.service.ts
```diff
@@ -0,0 +1,131 @@
...
+    try {
+      if (direction === MigrationDirection.FileToDb && !existsSync(this.nodeConfig.mmrPath)) {
+        throw new MMRMigrateError(
+          `MMR file not found: ${this.nodeConfig.mmrPath}`,
+          MMRMigrateErrorCode.MMRFileNotFoundError
+        );
+      } else if (direction === MigrationDirection.DbToFile) {
+        const schema = await getExistingProjectSchema(this.nodeConfig, this.sequelize);
+        if (!schema) {
+          throw new MMRMigrateError(`Schema for MMR DB does not exist`, MMRMigrateErrorCode.SchemaCreationError);
```
stwiname, line 59:
```suggestion
          throw new MMRMigrateError(`Project with db schema ${schema} does not exist`, MMRMigrateErrorCode.SchemaCreationError);
```

packages/node-core/src/indexer/mmrMigrate.service.ts
```diff
@@ -0,0 +1,131 @@
...
+        }
+        const tables = await this.sequelize.query('SELECT tablename FROM pg_tables WHERE schemaname = ?', {
+          replacements: [schema],
+          type: QueryTypes.SELECT,
+        });
+        const tableNames = tables.map((table: any) => table.tablename);
+
+        if (!tableNames.includes('_mmrs')) {
```
stwiname, line 67: Shouldn't the table be called `_mmr` like with `_metadata`?

packages/node-core/src/indexer/mmrMigrate.service.ts
```diff
@@ -0,0 +1,131 @@
...
+          throw new MMRMigrateError(
+            'The _mmr table does not exist in the database',
```
stwiname, line 69: Users don't need to know the internal workings.
```suggestion
            'MMR data is not found in the database, unable to migrate.',
```

packages/node/src/yargs.ts
stwiname: It seems there are still other args missing. Db schema and project path are ones i can think of. Rather than redefining them for subcommands we should move them to a common place

stwiname:
Therese a test failing but otherwise this looks good

PR #1657, opened 2023-05-03
multiple endpoints improvement

### LINKED ISSUE #1650: Multiple endpoints improvements
### Description
Be smarter about selecting endpoints to make requests to in order to speed up indexing time

### Requirements
 * Provide a chain agnostic way of rating endpoints based on failed requests
   * Remove or suspend for a period of time (perhaps with exponential backoff) endpoints that offer no service (5xx errors etc)
   * Prioritise endpoints that have good response time and minimal rate limiting (use some kind of weighting system??)
 * Update retry logic to swap between endpoints rather than retrying the same endpoint
 * Improve logging with what endpoints are not used (can log at an interval)

DeveloperInProgress:
# Description
Be smarter about selecting endpoints to make requests to in order to speed up indexing time

Fixes #1650 

## Type of change

## Checklist

- [ ] I have tested locally


stwiname (commented):
This is getting really hard to follow.

I think this might help clear things up:
```
interface IApi<A /* api */, SA /* safe api */, B /* block */ > {
  fetchBlocks(heights: number[]): Promise<B>

  // Api scoped to a specific block height
  safeApi: SA;

  // Api not scoped to any block height
  unsafeApi: A;
}
```

* `ApiService` would implement `IApi`.
* Each function on `IApi` in `ApiService` would request to a `ConnectionPool` to get a specific to a connection and call that function on the inner `IApi`
* `ConenctionPool` would wrap `fetchBlocks` to monitor performance/reliability. I don't think its feasible to monitor performance of other methods.

The call stack would be something like this: 
`ApiService implements IApi` -> `ConnectionPool` -> `Api implements IApi`

packages/node-core/src/api.service.ts
```diff
@@ -2,43 +2,81 @@
...
-type FetchFunction<T> = (batch: number[]) => Promise<T[]>;
-type FetchFunctionProvider<T> = () => FetchFunction<T>;
+type FetchFunction<T, A> = (batch: number[], api: A) => Promise<T[]>;
+type FetchFunctionProvider<T, A> = () => [FetchFunction<T, A>, A];
 
-export abstract class ApiService<P extends ISubqueryProject = ISubqueryProject, A = any, B = any> {
-  constructor(protected project: P) {}
+export interface IApi<A, SA, B> {
+  fetchBlocks(heights: number[], ...args: any): Promise<B[]>;
+  safeApi: SA;
```
stwiname, line 17: This needs to be a function so we can provide the block height

packages/node-core/src/indexer/connectionPool.service.ts
```diff
@@ -1,57 +1,153 @@
...
+  };
+
+  constructor() {
+    this.startLoggingEndpointStatus(LOG_INTERVAL_MS);
+  }
 
   async onApplicationShutdown(): Promise<void> {
     await Promise.all(
-      Object.keys(this.connectionPool)?.map((key) => this.connectionPool[toNumber(key)].apiDisconnect())
+      Object.keys(this.connectionPool).map((key) => this.connectionPool[toNumber(key)].apiDisconnect!())
```
stwiname, line 63: Object.values

packages/node-core/src/indexer/connectionPool.service.ts
```diff
@@ -1,57 +1,153 @@
...
-export class ConnectionPoolService<T extends ApiConnection> implements OnApplicationShutdown {
+export class ConnectionPoolService<T extends IApi<any, any, any>> implements OnApplicationShutdown {
   private allApi: T[] = [];
+  private apiToIndexMap: Map<any, number> = new Map();
+  private indexToEndpointMap: Record<number, string> = {};
   private connectionPool: Record<number, T> = {};
-  private taskCounter = 0;
+  private performanceScores: Record<number, number> = {};
+  private backoffDelays: Record<number, number> = {};
+  private failureCounts: Record<number, number> = {};
```
stwiname, line 49: Rather than having a bunch of mappings it might make sense to have a class/object that contains all of these into one.

packages/node-core/src/indexer/fetch.service.ts
```diff
@@ -55,7 +58,7 @@ export abstract class BaseFetchService<
   protected abstract preLoopHook(data: {valid: boolean; startHeight: number}): Promise<void>;
 
   constructor(
-    protected apiService: ApiService,
+    protected apiService: ApiService<A, SA, BC>,
```
stwiname, line 61: You can set defaults on the ApiService generics so we don't need to provide generics everywhere. This is useful here were fetch.service doesn't really care about the generic types of api service. The same can be applied in other places.

Alternatively if you want to narrow down the type `BaseFetchService` coulld have `API extends ApiService` rather than all the individual generics of `ApiService`

packages/node/src/indexer/api.service.ts
```diff
@@ -34,7 +34,7 @@ const logger = getLogger('api');
 
 @Injectable()
 export class ApiService
-  extends BaseApiService<SubqueryProject, ApiPromise>
+  extends BaseApiService<ApiPromise, ApiPromise, BlockContent>
```
stwiname, line 37: There is a double up of `ApiPromise` here

packages/node/src/indexer/apiPromise.connection.ts
```diff
@@ -39,43 +47,62 @@ export class ApiPromiseConnection implements ApiConnection {
...
+      this.unsafeApi,
+      heights,
+      overallSpecVer,
+    );
+    return blocks;
   }
 
   async apiConnect(): Promise<void> {
-    await this._api.connect();
+    await this.unsafeApi.connect();
```
stwiname, line 66: `IApi` seems to have connect and disconnect methods. I don't think it makes sense to use `unsafeApi` here. It implies that it is a different instance to `safeApi`

stwiname (commented):
This looks really good. Its much easier to follow.

packages/node-core/src/api.service.ts
```diff
@@ -2,43 +2,81 @@
...
+export interface IApi<A, SA, B> {
+  fetchBlocks(heights: number[], ...args: any): Promise<B[]>;
+  safeApi(height: number): SA;
+  unsafeApi: A;
+  handleError?(error: Error): ApiConnectionError;
+  apiConnect?(): Promise<void>;
+  apiDisconnect?(): Promise<void>;
+}
+
+export abstract class Api<A, SA, B> implements IApi<A, SA, B> {
```
stwiname, line 24: I don't think this is used

packages/node-core/src/indexer/connectionPool.service.ts
```diff
@@ -1,66 +1,165 @@
...
+              return result;
+            } catch (error) {
+              this.handleApiError(index, target.handleError!(error as Error));
+              throw error;
+            }
+          };
+        }
+
+        // Delegate other property access and method calls to the original api object
+        return Reflect.get(target, prop, receiver);
```
stwiname, line 117: I didn't know this existed 👍

packages/node-core/src/indexer/fetch.service.ts
```diff
@@ -23,6 +23,7 @@ const DICTIONARY_MAX_QUERY_SIZE = 10000;
 const CHECK_MEMORY_INTERVAL = 60000;
 
 export abstract class BaseFetchService<
+  API extends ApiService,
```
stwiname, line 26: I think this can be, if we use the interface then it simplifies writing tests
```suggestion
  API extends IApi,
```

packages/node/src/indexer/apiPromise.connection.ts
```diff
@@ -3,19 +3,27 @@
...
 // eslint-disable-next-line @typescript-eslint/no-var-requires
 const { version: packageVersion } = require('../../package.json');
 
 const RETRY_DELAY = 2_500;
 
-export class ApiPromiseConnection implements ApiConnection {
-  constructor(private _api: ApiPromise) {}
+export class ApiPromiseConnection extends Api<
+  ApiPromise,
+  ApiPromise,
```
stwiname, line 17: I think the safe api type should be `ApiAt`

packages/node-core/src/api.service.ts
DeveloperInProgress: ApiPromiseConnection extends this

stwiname: Would it make more sense to just have ApiPromiseConnection implement IApi?

packages/node-core/src/indexer/project.service.ts
```diff
@@ -24,7 +24,7 @@ class NotInitError extends Error {
   }
 }
 
-export abstract class BaseProjectService<DS extends {startBlock?: number}> implements IProjectService<DS> {
+export abstract class BaseProjectService<A, SA, B, DS extends {startBlock?: number}> implements IProjectService<DS> {
```
stwiname, line 27:
```suggestion
export abstract class BaseProjectService<Api extends IApi, DS extends {startBlock?: number}> implements IProjectService<DS> {
```

DeveloperInProgress: I think we should let it extend ApiService because it accesses `networkMeta`

stwiname:
Tests are failing, can you please fix them

packages/node-core/src/api.service.ts
```diff
@@ -2,43 +2,57 @@
...
-
-export abstract class ApiService<P extends ISubqueryProject = ISubqueryProject, A = any, B = any> {
-  constructor(protected project: P) {}
+export interface IApi<A = any, SA = any, B = any> {
+  fetchBlocks(heights: number[], ...args: any): Promise<B[]>;
+  safeApi(height: number): SA;
+  unsafeApi: A;
+  handleError?(error: Error): ApiConnectionError;
+  apiConnect?(): Promise<void>;
+  apiDisconnect?(): Promise<void>;
```
stwiname, line 18: I don't think optional properties on an interface is a good idea, either we implement them or we have a different interface that extends this. 

In this case, the exposed IApi that is consumed outside of ApiService/ConnectionPool such as fetch and project service shouldn't be aware of these but the connection specific (apiPromise.connection) these should be required properties that implement an extended interface

packages/node-core/src/indexer/connectionPool.service.ts
```diff
@@ -71,6 +170,84 @@ export class ConnectionPoolService<T extends ApiConnection> implements OnApplica
...
+  }
+
+  private calculatePerformanceScore(responseTime: number, failureCount: number): number {
+    const responseTimeScore = 1 / responseTime;
+    const failureScore = 1 - failureCount / MAX_FAILURES;
+    return RESPONSE_TIME_WEIGHT * responseTimeScore + FAILURE_WEIGHT * failureScore;
+  }
+
+  @Interval(LOG_INTERVAL_MS)
+  logEndpointStatus(): void {
```
stwiname, line 182: It would be good if we could log the weights of the endpoints for debugging purposes

packages/node-core/src/indexer/connectionPool.service.ts
```diff
@@ -1,66 +1,165 @@
...
+
+        // Delegate other property access and method calls to the original api object
+        return Reflect.get(target, prop, receiver);
+      },
+    });
+
+    return wrappedApi as T;
   }
 
   getNextConnectedApiIndex(): number {
```
stwiname, line 124: This can be private. Can you check the visibility of any other methods as well.

packages/node-core/src/indexer/connectionPool.service.ts
```diff
@@ -71,6 +172,92 @@ export class ConnectionPoolService<T extends ApiConnection> implements OnApplica
...
+    logger.info(suspendedEndpointsInfo);
+
+    Object.keys(this.pool)
+      .map(toNumber)
+      .forEach((index) => {
+        const endpoint = new URL(this.pool[index].endpoint).hostname;
+        const weight = this.pool[index].performanceScore;
+        logger.debug(`Weight of endpoint ${endpoint}: ${weight}`);
+      });
+  }
```
stwiname, line 213:
```suggestion
    Object.values(this.pool)
      .forEach((item) => {
        const endpoint = new URL(item.endpoint).hostname;
        const weight = item.performanceScore;
        logger.debug(`Weight of endpoint ${endpoint}: ${weight}`);
      });
  }
```

packages/node-core/src/indexer/connectionPool.service.ts
```diff
@@ -1,66 +1,167 @@
...
   async connectToApi(apiIndex: number): Promise<void> {
     await this.allApi[apiIndex].apiConnect();
+    this.pool[apiIndex].connection = this.allApi[apiIndex];
   }
 
   get api(): T {
     const index = this.getNextConnectedApiIndex();
     if (index === -1) {
-      throw new Error('No connected api');
+      throw new Error('All of the endpoints are suspended at the moment');
```
stwiname, line 96: Is this desired behaviour? If all the endpoints are suspended it doesn't mean they're unusable. If we throw here the node will most likely restart and resume with reset weights then get back to the same point and restart again.

Ideally we can distinguish between an endpoint not working and rate limiting. With rate limiting we should just throttle and with endpoints being down we should exit if they are all failing

packages/node-core/src/indexer/connectionPool.service.ts
```diff
@@ -1,66 +1,167 @@
...
 
-  getNextConnectedApiIndex(): number {
-    if (Object.keys(this.connectionPool).length === 0) {
+  private getNextConnectedApiIndex(): number {
+    const indices = Object.keys(this.pool)
+      .map(Number)
+      .filter((index) => !this.pool[index].backoffDelay);
+
+    if (indices.length === 0) {
       return -1;
```
stwiname, line 132: Why not return undefined?

packages/node/src/indexer/api.service.ts
```diff
@@ -85,14 +85,19 @@ export class ApiService
       );
     }
 
-    const connections: ApiPromiseConnection[] = [];
+    const endpointToApiIndex: Record<string, ApiPromiseConnection> = {};
 
     await Promise.all(
```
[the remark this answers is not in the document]
DeveloperInProgress: we can do it with promise.allSettled() instead of promise.all(). let's try that in another PR

packages/node-core/src/indexer/mmrMigrate.service.ts
```diff
@@ -3,9 +3,10 @@
 
 import {existsSync} from 'fs';
 import {DEFAULT_WORD_SIZE} from '@subql/common';
-import {NodeConfig, getExistingProjectSchema, getLogger, PgBasedMMRDB} from '@subql/node-core';
+import {NodeConfig, getExistingProjectSchema, PgBasedMMRDB} from '@subql/node-core';
```
jiqiang90, line 6: this should use relative path, it been self-import

packages/node-core/src/indexer/project.service.ts
```diff
@@ -8,7 +8,7 @@ import {EventEmitter2} from '@nestjs/event-emitter';
 import {MetadataKeys} from '@subql/node-core/indexer/entities';
```
jiqiang90, line 8: Can you update these two self-import please

packages/node-core/src/indexer/unfinalizedBlocks.service.spec.ts
```diff
@@ -3,7 +3,7 @@
 
 // import { Header } from '@polkadot/types/interfaces';
 import {EventEmitter2} from '@nestjs/event-emitter';
-import { SchedulerRegistry } from '@nestjs/schedule';
+import {SchedulerRegistry} from '@nestjs/schedule';
 import {CacheMetadataModel, Header, StoreCacheService} from '@subql/node-core';
```
jiqiang90, line 7: here self-import as well

jiqiang90 (approved):

packages/node-core/src/indexer/storeCache/cacheMmr.spec.ts
```diff
@@ -46,7 +46,8 @@ describe('CacheMMR', () => {
     const spy = jest.spyOn(db, 'setLeafLength');
 
     await cacheDb.setLeafLength(100);
-    expect(spy).toBeCalledTimes(1);
+    //there will be no call to db.setLeafLength
```
jiqiang90, line 49: ` await cacheDb.flush({afterCommit:jest.fn()} as any,undefined)` 

We can use this, so db will setLeafLength once

PR #1696, opened 2023-05-12
deprecate project id from poi table, track deployment in metadata

### LINKED ISSUE #1662: Remove projectId from _poi table
The poi table currently has a column for the `projectId` but in every row of the table this will be the same value.

We should remove this column from the table and instead we can rely on a value in the _metadata table. This will require a migration to the table to ensure compatibility.

jiqiang90:
# Description

TODO:

- [x] should not use Record type for deployments, user could switch back to same deployment and block height could be override

Fixes #1662 

## Type of change

## Checklist

- [ ] I have tested locally

### DIFF
--- a/packages/node-core/src/indexer/entities/Poi.entity.ts
+++ b/packages/node-core/src/indexer/entities/Poi.entity.ts
@@ -10,7 +10,7 @@ export interface ProofOfIndex {
   parentHash?: Uint8Array;
   operationHashRoot: Uint8Array;
   mmrRoot?: Uint8Array;
-  projectId: string;
+  projectId?: string;
 }
 
 export interface PoiModel extends Model<ProofOfIndex>, ProofOfIndex {}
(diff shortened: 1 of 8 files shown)


packages/node-core/src/indexer/project.service.ts
```diff
@@ -195,6 +196,28 @@ export abstract class BaseProjectService<DS extends {startBlock?: number}> imple
       metadata.set('specName', specName);
     }
 
+    if (!existing.deployments) {
```
stwiname, line 199: I think this can be moved into a function

packages/utils/src/query/types.ts
```diff
@@ -18,6 +18,7 @@ export type MetaData = {
   queryNodeVersion: string;
   startHeight?: number;
   rowCountEstimate: [TableEstimate];
+  deployments: Record<string, number>;
```
stwiname, line 21: Do we need to make this queryable?

packages/node-core/src/indexer/entities/Poi.entity.ts
```diff
@@ -10,16 +10,16 @@ export interface ProofOfIndex {
...
 
 export interface PoiModel extends Model<ProofOfIndex>, ProofOfIndex {}
 
 export type PoiRepo = typeof Model & {
   new (values?: unknown, options?: BuildOptions): PoiModel;
 };
-
-export function PoiFactory(sequelize: Sequelize, schema: string): PoiRepo {
+// Project Id is deprecated
+export function PoiFactoryDeprecate(sequelize: Sequelize, schema: string): PoiRepo {
```
stwiname, line 22: I don't think we should keep a deprecated version. We should drop the column instead

packages/utils/src/query/types.ts
jiqiang90: let's make it queryable, useful user track deployment history

stwiname (approved):

Issue #1689, opened 2023-05-11
Update codegen to work with TS strict setting


stwiname:
If `strict` is set to true in a project ts config then the project will fail to compile, we should support the strictest TS settings

PR #1713, opened 2023-05-16
strict codegen

bz888:
# Description
Update `subql codegen` code to be compatible with strict typescript
Fixes https://github.com/subquery/subql/issues/1689

This will be a breaking change,
for all project that uses

e.g.
```
  let record = new StarterEntity(block.block.header.hash.toString());
```
would not build, all required fields must be added to the constructor when using `new`

https://github.com/subquery/subql-starter/pull/80

## Type of change
- Breaking change
## Checklist

- [x] I have tested locally

### DIFF
--- a/packages/cli/src/template/model.ts.ejs
+++ b/packages/cli/src/template/model.ts.ejs
@@ -22,7 +22,7 @@ export class <%= props.className %> implements Entity {
     }
 
 <% props.fields.forEach(function(field){ %>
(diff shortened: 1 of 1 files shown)


stwiname (commented):
For historical reference. I think we should update the constructor rather than have a non-null assertion

stwiname (approved):

stwiname:
There are a few tests failing

PR #1755, opened 2023-05-25
Fix subcommand could escape, set profiler at init

jiqiang90:
# Description

Fixed subcommand could escape from app init

Also moved setup profiler from app init

Fixes # (issue)

## Type of change

## Checklist

- [ ] I have tested locally

### DIFF
--- a/packages/node/src/main.ts
+++ b/packages/node/src/main.ts
@@ -1,29 +1,10 @@
 // Copyright 2020-2022 OnFinality Limited authors & contributors
 // SPDX-License-Identifier: Apache-2.0
 
-import { initLogger } from '@subql/node-core/logger';
 import { yargsOptions } from './yargs';
 
 const { argv } = yargsOptions;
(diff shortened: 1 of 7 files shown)


packages/node/src/main.ts
```diff
@@ -1,29 +1,10 @@
...
-    argv._[0] === 'mmr-regen' ||
-    argv._[0] === 'force-clean' ||
-    argv._[0] === 'reindex'
-  )
-) {
-  void bootstrap();
-}
+// Lazy import, to allow logger in yargsOptions to be initialised before setProfiler()
+const { setProfiler } = require('@subql/node-core');
+setProfiler(argv.profiler);
```
stwiname, line 10: If this is here, will it mean that only indexing can support profiler?

I think if it was at the bottom of yargs.ts it would be for any command

stwiname (approved):

PR #1760, opened 2023-05-26
add a cache to WsProvider

### LINKED ISSUE #1751: Improve Substrate api requests
Here are some screenshots of RPC requests for dictionary projects. It seems we are fetching headers and runtime versions twice for every block we fetch. 

We should only need to fetch the header and runtime version once for each block.

Screenshot 2023-05-24 at 4 59 12 PM
Screenshot 2023-05-24 at 4 59 00 PM

DeveloperInProgress:
# Description
Cache runtimeVersion and block headers to avoid duplicate requests to the rpc node.

Fixes #1751 

## Type of change

## Checklist

- [ ] I have tested locally

### DIFF
--- a/packages/node/src/indexer/apiPromise.connection.spec.ts
+++ b/packages/node/src/indexer/apiPromise.connection.spec.ts
@@ -20,14 +20,17 @@ describe('ApiPromiseConnection', () => {
       'User-Agent': `SubQuery-Node ${packageVersion}`,
     };
 
-    //wsProvider = new WsProvider('wss://kusama.api.onfinality.io/public-ws', RETRY_DELAY, headers);
+    wsProvider = new WsProvider(
+      'wss://kusama.api.onfinality.io/public-ws',
+      RETRY_DELAY,
+      headers,
+    );
     httpProvider = new HttpProvider('https://kusama.api.onfinality.io/public');
 
-    //jest.spyOn(wsProvider, 'send');
+    jest.spyOn(wsProvider, 'send');
     jest.spyOn(httpProvider, 'send');
   });
 
-  /*
   afterEach(async () => {
     if (wsProvider.isConnected) {
       await wsProvider.disconnect();
@@ -35,7 +38,8 @@ describe('ApiPromiseConnection', () => {
   });
 
   it('should not make duplicate requests for state_getRuntimeVersion on wsProvider', async () => {
-    const cachedProvider = ApiPromiseConnection.createCachedProvider(wsProvider);
+    const cachedProvider =
+      ApiPromiseConnection.createCachedProvider(wsProvider);
 
     await cachedProvider.send('state_getRuntimeVersion', [TEST_BLOCKHASH]);
     await cachedProvider.send('state_getRuntimeVersion', [TEST_BLOCKHASH]);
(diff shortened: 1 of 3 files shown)


packages/node/src/indexer/apiPromise.connection.ts
```diff
@@ -53,6 +60,50 @@ export class ApiPromiseConnection implements ApiConnection {
...
+          return result;
+        },
+      );
+    };
+
+    return new Proxy(provider, {
+      get: function (target, prop, receiver) {
+        if (prop === 'send') {
+          return function (method, params, isCacheable, subscription) {
+            if (
```
stwiname, line 85: It would be good to have a comment as to why we do these 2 methods

packages/node/src/indexer/apiPromise.connection.ts
```diff
@@ -53,6 +60,50 @@ export class ApiPromiseConnection implements ApiConnection {
     await this._api.disconnect();
   }
 
+  /* eslint-disable prefer-rest-params */
+  static createCachedProvider(provider: WsProvider | HttpProvider) {
```
stwiname, line 64: I don't think this needs to be static. It can be a function outside of a class.

```suggestion
  static createCachedProvider(provider: ProviderInterface): ProviderInterface {
```

packages/node/src/indexer/apiPromise.connection.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+
+describe('ApiPromiseConnection', () => {
+  let wsProvider;
+  let httpProvider;
+
+  beforeEach(async () => {
+    wsProvider = new WsProvider('wss://kusama.api.onfinality.io/public-ws');
+    httpProvider = new HttpProvider('https://kusama.api.onfinality.io/public');
+
+    const api = await ApiPromise.create({ provider: wsProvider });
```
stwiname, line 23: Unused

packages/node/src/indexer/apiPromise.connection.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const { version: packageVersion } = require('../../package.json');
+const RETRY_DELAY = 2_500;
+
+const TEST_BLOCKHASH =
+  '0x3b712c10ba98c5421a468d1411c94479381be307b7ecb5b2602cf0d395eb4292';
+
+describe('ApiPromiseConnection', () => {
+  let wsProvider;
+  let httpProvider;
```
stwiname, line 17: Missing types

packages/node/src/indexer/apiPromise.connection.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+  });
+
+  it('should not make duplicate requests for state_getRuntimeVersion on wsProvider', async () => {
+    const cachedProvider =
+      ApiPromiseConnection.createCachedProvider(wsProvider);
+
+    await cachedProvider.send('state_getRuntimeVersion', [TEST_BLOCKHASH]);
+    await cachedProvider.send('state_getRuntimeVersion', [TEST_BLOCKHASH]);
+
+    expect(wsProvider.send).toHaveBeenCalledTimes(1);
```
stwiname, line 36: All of these tests would fail if we run multiple requests at once. I know the usage is not like this currently but with a simple change we can also support this use case

e.g failing test
```
    await Promise.all([
          cachedProvider.send('state_getRuntimeVersion', [TEST_BLOCKHASH]),
          cachedProvider.send('state_getRuntimeVersion', [TEST_BLOCKHASH]),
    ]);

    expect(wsProvider.send).toHaveBeenCalledTimes(1);
```

To solve this we should cache the promise instead of the result

stwiname (approved):

jiqiang90 (approved):

Issue #1763, opened 2023-05-29
Failing test gives non-useful error


stwiname:
```
2023-05-29T02:41:12.620Z <subql-testing> WARN Test Contract Creation failed to run
TypeError: Cannot convert undefined or null to object
    at Function.keys (<anonymous>)
    at TestingService.runTest (/Users/scotttwiname/.nvm/versions/node/v18.16.0/lib/node_modules/@subql/node/node_modules/@subql/node-core/dist/indexer/testing.service.js:128:24)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async TestingService.run (/Users/scotttwiname/.nvm/versions/node/v18.16.0/lib/node_modules/@subql/node/node_modules/@subql/node-core/dist/indexer/testing.service.js:57:21)
    at async testingInit (/Users/scotttwiname/.nvm/versions/node/v18.16.0/lib/node_modules/@subql/node/dist/subcommands/testing.init.js:26:9)
```

I think it means that there is no entity with a matching id. Not sure if that's coming from the db or the entities in the test

We should give a better message so the user can figure out how to fix the test

PR #1766, opened 2023-05-29
Handle entitiy not found, flush cache after running handler

stwiname:
# Description
Handles when an entity doesn't exist in the DB and returns an appropriate error. It also will flush the cache to ensure the data exists in the DB so that constraint checks are run

Fixes https://github.com/subquery/subql/issues/1763

## Type of change

## Checklist

- [ ] I have tested locally

### DIFF
--- a/packages/node-core/src/indexer/testing.service.ts
+++ b/packages/node-core/src/indexer/testing.service.ts
@@ -119,7 +119,7 @@ export abstract class TestingService<B, DS> {
 
   private async runTest(test: SubqlTest, sandbox: TestSandbox) {
     logger.info(`Starting test: ${test.name}`);
-    const schema = `test-${this.nodeConfig.dbSchema}`;
+    const schema = this.nodeConfig.dbSchema;
 
     try {
       // Fetch block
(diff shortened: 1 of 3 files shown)


stwiname:
@DeveloperInProgress is any of this code relevant now https://github.com/subquery/subql/blob/5c09434643d1c4eca0537b588edb043806ba5317/packages/node-core/src/indexer/testing.service.ts#L122

DeveloperInProgress:
> @DeveloperInProgress is any of this code relevant now
> 
> https://github.com/subquery/subql/blob/5c09434643d1c4eca0537b588edb043806ba5317/packages/node-core/src/indexer/testing.service.ts#L122

looks like we needs this since we drop and recreate schemas between tests

packages/node-core/src/indexer/testing.service.ts
```diff
@@ -119,7 +119,7 @@ export abstract class TestingService<B, DS> {
 
   private async runTest(test: SubqlTest, sandbox: TestSandbox) {
     logger.info(`Starting test: ${test.name}`);
-    const schema = `test-${this.nodeConfig.subqueryName}`;
+    const schema = `test-${this.nodeConfig.dbSchema}`;
```
stwiname, line 122: Won't this now end up being `test-test-*` because node config will update the db schema?

jiqiang90 (approved):

Issue #1687, opened 2023-05-11
Project upgrades


stwiname:
### Description

In order to make projects on the network upgradable as well as keeping track of versioning of projects there is a need to be able to upgrade projects.

### Requirements

To do this an update to the project manifest is required. It will mean adding a reference to the previous project and the block height at which the project should update.

Example addition to a project manifest:
```yaml
parent:
    reference: <ipfs-cid>
    blockHeight: 1234
```

This will require updating the indexer to follow back parent references to the start as there could be multiple versions for various block ranges and upgrade to the appropriate version at the right time

It should also be possible to rewind historical projects to the height specified by an upgrade in order to not need to reindex from the start

PR #1797, opened 2023-06-13
Project Upgrades

stwiname:
# Description

## Motivation
Projects are not always set and forget, things on chain change, contracts are upgraded, runtimes are upgraded and there needs to be a way to upgrade projects. Currently when developing you can run updated projects and use the same schema. This doesn't work with the network and it is also not possible to make sure changes are applied from specific block heights.

## Usage

In order to use the feature we have new options added to the project manifest:
```yaml
parent:
  block: 1000 # the block to switch from the referenced project to this current project
  reference: QmXw6FN6eScxvYXYceuCjKMpqmnuCxwY3Cx4HPhDXgUWe5 # the previous project to upgrade from
```

### Example project

Testing project (Polkadot Dictionary project)

Start: `QmPPPB292SKjMB2Ps66LD973RpgQuAhncdsicpgm15KfBj`
100: `QmXMwMDjixwGbufd6kMtz8ouPSByeK5irAcQh4bFz2iS4N`
1000: `QmXw6FN6eScxvYXYceuCjKMpqmnuCxwY3Cx4HPhDXgUWe5`
    Throws error at 1234
1234: `QmSP3bsrkcKGBGj8YfKKiPaxQnRSdZCuvDCSYVHDp4x35w`
    Fixes error at 1234

1050: `QmcYMx6AsAfjUW3NX5a9rcU2n5Gz1pfmjXnmTEpi4BAsea`
    Use to test rewinding to a fork. Parent: `QmXw6FN6eScxvYXYceuCjKMpqmnuCxwY3Cx4HPhDXgUWe5`

> TODO update this further with regard to db schemas

## Architectural changes

Some high level overview of internal changes:
* Generalise configure module to move more code into node-core
* Getting datasources now comes from a single source of truth in the project service. This ensures we are using the same datasources everywhere, previously there were places that excluded dynamic datasources
* Moves applying cron filters to datasources on to SubqueryProject
* Introduces `IProjectUpgradeService`/`ProjectUpgradeService` to use the right project version at the right height. This also creates a proxy version of `ISubqeryProject` that will change depending on what the current indexing height is

Fixes https://github.com/subquery/subql/issues/1687

## Type of change

- [x] This change requires a documentation update https://github.com/subquery/documentation/pull/398

## Checklist

- [x] I have tested locally
- [x] Updated any relevant documentation https://github.com/subquery/documentation/pull/398


stwiname:
I think testing might be broken. I'll to look into it shortly

packages/node-core/src/configure/ProjectUpgrade.service.ts
```diff
@@ -0,0 +1,296 @@
...
+      if (currentProject.parent.block) {
+        addProject(currentProject.parent.block, currentProject);
+      }
+      // At the limit of the start height for Parent
+      if (startHeight !== undefined && startHeight >= currentProject.parent.block) {
+        break;
+      }
+
+      // Load the next project and repeat
+      const nextProject = await loadProject(currentProject.parent.reference);
```
jiqiang90, line 207: Is that possible when we load the parent reference failed?

packages/node-core/src/indexer/dictionary.service.test.ts
```diff
@@ -233,38 +301,60 @@ describe('DictionaryService', () => {
...
+          {field: 'event', value: 'Bonded'},
+        ],
+      },
+    ]);
+
+    // Then end block of the selected queries map item
+    expect(results?.queryEndBlock).toEqual(499);
   });
+
+  // it('sort map', () => {
```
jiqiang90, line 353: Is this test working?

packages/node-core/src/utils/project.ts
```diff
@@ -243,3 +250,78 @@ export async function initDbSchema(
 export async function initHotSchemaReload(schema: string, storeService: StoreService): Promise<void> {
   await storeService.initHotSchemaReloadQueries(schema);
 }
+
+type IsRuntimeDs = (ds: BaseDataSource) => boolean;
+
+// eslint-disable-next-line @typescript-eslint/require-await
+export async function insertBlockFiltersCronSchedules<DS extends BaseDataSource = BaseDataSource>(
```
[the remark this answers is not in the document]
stwiname: I don't think ive fixed anything, just moved code to node core

packages/node-core/src/indexer/dictionary.service.test.ts
stwiname: Removed because its no longer relevant

packages/node-core/src/configure/ProjectUpgrade.service.ts
stwiname: I have a better error message now

jiqiang90 (approved):

PR #1809, opened 2023-06-20
transform undefined to null in actual entities

DeveloperInProgress:
# Description
Records that gets stored in getCache at the time of `set` will remain undefined while TestingService expects null values from database instead of undefined. Undefined values must be transformed to null before entity checks happen in TestingService. 

Fixes # (issue)

## Type of change

## Checklist

- [ ] I have tested locally

### DIFF
--- a/packages/node-core/src/indexer/testing.service.ts
+++ b/packages/node-core/src/indexer/testing.service.ts
@@ -175,7 +175,7 @@ export abstract class TestingService<A, SA, B, DS> {
           const attributes = actualEntity;
           Object.keys(attributes).map((attr) => {
             const expectedAttr = (expectedEntity as Record<string, any>)[attr] ?? null;
-            const actualAttr = (actualEntity as Record<string, any>)[attr];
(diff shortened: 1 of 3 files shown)


stwiname (approved):

stwiname:
It would be great to have a test for this

packages/node-core/src/indexer/testing.service.ts
```diff
@@ -52,6 +52,10 @@ export abstract class TestingService<A, SA, B, DS> {
     protected readonly apiService: IApi<A, SA, B>,
     protected readonly indexerManager: IIndexerManager<B, DS>
   ) {
+    if (!nodeConfig.disableHistorical) {
+      throw new Error(`Historical data not supported in testing service`);
```
stwiname, line 56: I don't think we need to throw an error. Just set it to null

stwiname:
Can you please update the changelog

stwiname (approved):

PR #1829, opened 2023-06-21
fix retry logic for workers in connection pool

DeveloperInProgress:
# Description

Fixes # (issue)

## Type of change

## Checklist

- [ ] I have tested locally


packages/node-core/src/indexer/connectionPoolState.manager.ts
```diff
@@ -0,0 +1,243 @@
...
+  lastRequestTime: number;
+  connected: boolean;
+  timeoutId?: NodeJS.Timeout;
+}
+
+const logger = getLogger('connection-pool-state');
+
+export class ConnectionPoolStateManager<T extends IApiConnectionSpecific<any, any, any>> {
+  private pool: Record<number, ConnectionPoolItem<T>> = {};
+  private errorTypeToScoreAdjustment = {
```
stwiname, line 31: I think this can be moved out of the class into a const

packages/node-core/src/indexer/worker/worker.connectionPoolState.manager.ts
```diff
@@ -0,0 +1,88 @@
...
+  'hostGetSuspendedIndices',
+  'hostDeleteFromPool',
+  'hostHandleApiError',
+  'hostHandleApiSuccess',
+  'hostGetDisconnectedIndices',
+  'hostShutdownPoolState',
+];
+
+@Injectable()
+export class WorkerConnectionPoolStateManager<T> {
```
stwiname, line 38: This should implement an interface so we know that it matches the non-worker version

packages/node/src/indexer/blockDispatcher/worker-block-dispatcher.service.ts
```diff
@@ -71,6 +78,30 @@ async function createIndexerWorker(
...
+      hostDeleteFromPool:
+        connectionPoolState.deleteFromPool.bind(connectionPoolState),
+      hostHandleApiError:
+        connectionPoolState.handleApiError.bind(connectionPoolState),
+      hostHandleApiSuccess:
+        connectionPoolState.handleApiSuccess.bind(connectionPoolState),
+      hostGetDisconnectedIndices:
+        connectionPoolState.getDisconnectedIndices.bind(connectionPoolState),
+      hostShutdownPoolState:
+        connectionPoolState.shutdown.bind(connectionPoolState),
```
stwiname, line 104: Does a worker need access to all of these? It seems like the interface could be simplified

packages/query/src/graphql/plugins/GetMetadataPlugin.ts
```diff
@@ -219,8 +219,10 @@ export const GetMetadataPlugin = makeExtendSchemaPlugin((build: Build, options)
       extend type Query {
         _metadata(chainId: String): _Metadata
 
-        _metadatas(after: Cursor, before: Cursor): # distinct: [_mmr_distinct_enum] = null
-        # filter: _MetadataFilter
+        _metadatas(
```
stwiname, line 222: I'm not sure what this change is for

packages/node/src/indexer/blockDispatcher/worker-block-dispatcher.service.ts
DeveloperInProgress: we call all of these functions from `ConnectionPoolService`

packages/node/src/indexer/blockDispatcher/worker-block-dispatcher.service.ts
```diff
@@ -71,6 +78,34 @@ async function createIndexerWorker(
         unfinalizedBlocksService.processUnfinalizedBlockHeader.bind(
           unfinalizedBlocksService,
         ),
+      hostAddToConnections:
```
stwiname, line 81: It would be great if we could move this to node-core

packages/node/src/indexer/fetch.module.ts
```diff
@@ -36,7 +38,20 @@ import { UnfinalizedBlocksService } from './unfinalizedBlocks.service';
...
+    ConnectionPoolStateManager,
+    {
+      provide: ConnectionPoolService,
+      useFactory: (
+        nodeConfig: NodeConfig,
+        poolStateManager: ConnectionPoolStateManager<ApiPromiseConnection>,
+      ) => {
+        return new ConnectionPoolService(
+          poolStateManager,
+          nodeConfig.batchSize,
```
stwiname, line 50: Maybe if the constructor takes NodeConfig then we can remove the need for a factory

packages/node-core/src/indexer/connectionPoolState.manager.ts
```diff
@@ -0,0 +1,269 @@
...
+  rateLimited: boolean;
+  failed: boolean;
+  lastRequestTime: number;
+  connected: boolean;
+  timeoutId?: NodeJS.Timeout;
+}
+
+const logger = getLogger('connection-pool-state');
+
+export interface IConnectionPoolStateManager<T extends IApiConnectionSpecific<any, any, any>> {
```
stwiname, line 29: I think this interface should be much simpler. We only really need what would be called from the workers. To me this would be getting the state and updating the state.

E.g. Im not sure why these are in the interface:
* `shutdown`
*  `deleteFromPool`
*  `addToConnections`

stwiname (commented):
There's some tests failing. They seem unrelated to changes though

packages/node-core/src/indexer/connectionPoolState.manager.ts
DeveloperInProgress: they get called from `ConnectionPoolService` in workers

stwiname: That doesn't seem right to me. Maybe more logic needs to be in connection pool state manager.

To me only the main thread should decide if connections are added or removed.

DeveloperInProgress: we keep all the apis in workers because that is where the fetching happens. so, some operations like attempting to reconnect to endpoints will be done in workers, add/delete will be called in the host pool state manager depending on the result of reconnection in workers.

stwiname (commented):
I think this is looking really good.

One thing to change, we don't need to log this for each worker, it can just be done on the main thread:

Screenshot 2023-07-07 at 9 25 12 AM

@jiqiang90 can you please give this a review too.

jiqiang90 (approved):

Issue #1858, opened 2023-07-02
Add upper limit to cache size


stwiname:
In order to not run into OOM issues we should set an upper limit on the cache size.

When this limit is reached, indexing should stop until the cache is flushed. 

This limit may be exceeded slightly if a block produced more entities than the limit, but any following blocks will be stopped

PR #1859, opened 2023-07-03
Set upper limit on store cache threshold

stwiname:
# Description
This sets an upper limit to the store cache size, when this happens we await the cache being flushed to stop the cache further increasing leading and leading to OOM errors.

I have also updated jest to fix some issues running locally.

Fixes https://github.com/subquery/subql/issues/1858

## Type of change

## Checklist

- [x] I have tested locally


packages/node-core/src/configure/NodeConfig.ts
```diff
@@ -80,6 +81,7 @@ const DEFAULT_CONFIG = {
   multiChain: false,
   unfinalizedBlocks: false,
   storeCacheThreshold: 1000,
+  storeCacheUpperLimit: 10000,
```
stwiname, line 84: Im unsure if this is the right default value.

jiqiang90 (approved):

PR #1870, opened 2023-07-05
inti db schema manually during test run

### LINKED ISSUE #1844: Unable to run multiple tests, failing to initialise db-schema
### Description

- Failing to run multiple tests with the `subqlTest` framework

### Details

**Network Details:**
  * [Network]

### Steps to Reproduce

Have multiple subqlTests, when running the tests, the first one would succeed. However, all tests following will throw `db` errors, struggling to find tables.

### Any other information

DeveloperInProgress:
# Description
Initialize schema entities at the start of each test run as it gets deleted when one test run finishes

Fixes #1844 #1877

## Type of change

## Checklist

- [ ] I have tested locally


packages/node-core/src/indexer/testing.service.ts
```diff
@@ -134,6 +135,8 @@ export abstract class TestingService<A, SA, B, DS> {
         await this.sequelize.createSchema(`"${schema}"`, {});
       }
```
stwiname, line 136: Is this necessary now?

DeveloperInProgress: yeah, schema needs to be create before calling `initDbSchema`

stwiname:
@bz888 can you please test and review this fixes your issue.

@DeveloperInProgress can you please update changelog

bz888 (approved):

bz888 (commented):
when creating a new file for tests, it doesnt seem to be correct ? but this is potentially a different issue

this is for the polkadot starter project

Screenshot 2023-07-07 at 5 47 08 PM

both files are exported under `test/index.ts`

```
export * from './mappingHandlers.test';
export * from './secondMappingHandler.test';
```

Screenshot 2023-07-07 at 5 46 19 PM

this is expected to fail, however, it is passing.

packages/node-core/src/indexer/testing.service.ts
```diff
@@ -204,6 +208,17 @@ export abstract class TestingService<A, SA, B, DS> {
...
 
+      //remove tested entities from DB
+
+      await Promise.all(
+        entitiesCreated.map(async (entity) => {
+          // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+          await store.remove(entity._name!, entity.id);
+        })
+      );
```
stwiname, line 218: Why is this necessary? Dropping the schema should be enough

packages/node-core/src/indexer/testing.service.ts
```diff
@@ -216,10 +231,54 @@ export abstract class TestingService<A, SA, B, DS> {
...
+    try {
+      // drop existing project schema and metadata table
       await this.sequelize.dropSchema(`"${schema}"`, {
         logging: false,
         benchmark: false,
       });
+
+      // TODO, remove this soon, once original enum are cleaned
+      // Deprecate, now enums are moved under schema, drop schema will remove project enums
+      await Promise.all(
```
stwiname, line 250: I think all new projects should be creating enums under the schema. Seeing as the testing data is ephemeral I don't see why enums should ever be under public.

packages/node-core/src/indexer/testing.service.ts
```diff
@@ -216,10 +231,54 @@ export abstract class TestingService<A, SA, B, DS> {
...
+          if (resultsDeprecated.length !== 0) {
+            await this.sequelize.query(`
+            DROP TYPE "${enumTypeNameDeprecated}";
+          `);
+          }
+        })
+      );
+
+      // remove schema from subquery table (might not exist)
+      const checker = await this.sequelize.query(
```
stwiname, line 263: Again, this shouldn't exist either. What's causing this to exist?

stwiname (commented):
Tests please

packages/node-core/src/indexer/testing.service.ts
DeveloperInProgress: dropping the schema and recreating it immediately seems to retain the records from before drop. I'm unable to find any docs related to this but the references are still intact.

packages/node-core/src/indexer/testing.service.ts
DeveloperInProgress: I copied the logic from force-clean

packages/node-core/src/indexer/testing.service.ts
stwiname: We should try and figure out a solution rather than a work around

packages/node/src/subcommands/testing.service.ts
```diff
@@ -22,21 +25,28 @@ export class TestingService extends BaseTestingService<
...
+    this.apiService = this.app.get(ApiService);
+
+    await (this.apiService as ApiService).init();
+    await projectService.init();
+
+    this.storeService = this.app.get(StoreService);
+    this.sequelize = this.app.get(Sequelize);
+    this.indexerManager = this.app.get(IndexerManager);
+    this.nodeConfig = this.app.get(NodeConfig);
+    this.project = this.app.get('ISubqueryProject');
```
stwiname, line 49: I don't think this is the right approach. We should use nestjs to achieve dependency injection

packages/node-core/src/indexer/testing.service.ts
```diff
@@ -44,14 +46,13 @@ export abstract class TestingService<A, SA, B, DS> {
...
-    protected readonly storeService: StoreService,
-    @Inject('ISubqueryProject') protected project: ISubqueryProject<any, DS>,
-    protected readonly apiService: IApi<A, SA, B>,
-    protected readonly indexerManager: IIndexerManager<B, DS>
-  ) {
+  protected apiService: IApi<A, SA, B> | undefined;
+  protected sequelize: Sequelize | undefined;
+  protected storeService: StoreService | undefined;
+  protected indexerManager: IIndexerManager<B, DS> | undefined;
+  protected app: TestingModule | undefined;
```
stwiname, line 53: Why are these all optional?

packages/node-core/src/indexer/testing.service.ts
```diff
@@ -119,23 +123,20 @@ export abstract class TestingService<A, SA, B, DS> {
...
 
-      // Init db
-      const schemas = await this.sequelize.showAllSchemas({});
-      if (!(schemas as unknown as string[]).includes(schema)) {
-        await this.sequelize.createSchema(`"${schema}"`, {});
-      }
-
-      this.storeService.setBlockHeight(test.blockHeight);
-      const store = this.storeService.getStore();
+      this.storeService!.setBlockHeight(test.blockHeight);
```
stwiname, line 138: Please don't use non-null assertions

packages/node/src/subcommands/testing.service.ts
DeveloperInProgress: created a new class `TestRunner` to run tests in seperate contexts. this will use nestjs dependency injection

packages/node-core/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 ### Fixed
+- Fix schema and its entities not being recreated between test runs in `TestingService` (#1870)
```
stwiname, line 9: I think we can update this now, the scope has changed a bit.

packages/node/src/subcommands/test.runner.ts
```diff
@@ -0,0 +1,33 @@
...
+  StoreService,
+} from '@subql/node-core';
+import { Sequelize } from '@subql/x-sequelize';
+import { SubqlProjectDs } from '../configure/SubqueryProject';
+import { ApiService } from '../indexer/api.service';
+import { IndexerManager } from '../indexer/indexer.manager';
+import { ApiAt, BlockContent } from '../indexer/types';
+
+@Injectable()
+export class TestRunner extends BaseTestRunner<
```
stwiname, line 18: I don't think we need this class. It doesn't extend BaseTestRunner in any way

stwiname (commented):
This looks really good. Just need to update the changelogs

jiqiang90 (approved):

PR #1894, opened 2023-07-20
[SKIP CI] benchmark

### LINKED ISSUE #1793: Automated Performance Benchmarking Process
The following projects
- ENS project
- Polkadot dictionary
- https://github.com/jamesbayly/subql_transfers/

Reqs
- Run from a high start block (same each time)
- Run for a fixed block range (try to aim for an hour runtime)
  - Potentially need to support end block in handlers
- Track block heights and timestamps
- At the end, record the final run time
- Automated so it is on the same VM spec after every full release
- Look at using a custom GitHub runner in our cluster
- Output a report somewhere (potentially in GitHub), probably send a notification to a Slack channel

jiqiang90:
# Description

Fixes #1793 

## Type of change

## Checklist

- [ ] I have tested locally


.github/workflows/benchmark-substrate.yml
```diff
@@ -0,0 +1,200 @@
...
+          # For posting a rich message using Block Kit
+          payload: |
+            {
+              "text": "Benchmark result ${{steps.current-time.outputs.formattedTime}}-v${{steps.extract_version.outputs.version}}",
+              "blocks": [
+                {
+                  "type": "section",
+                  "text": {
+                    "type": "mrkdwn",
+                    "text": "Results: ${{ job.status }}\n Indexer:@subql/node@${{steps.extract_version.outputs.version}}\n workers:${{github.event.inputs.workers}}\n batch-size:${{github.event.inputs.batch-size}}\n block ${{steps.query_meta.outputs.start_height}} to ${{steps.query_meta.outputs.last_processed_height}}, total ${{steps.query_meta.outputs.total_height}} in ${{github.event.inputs.benchmark-time}}\n More details: https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}"
```
stwiname, line 194: Maybe we can add the deployment to this?

.github/workflows/benchmark-substrate.yml
```diff
@@ -0,0 +1,200 @@
...
+        run: psql -h postgres -d postgres -U postgres -c 'CREATE EXTENSION IF NOT EXISTS btree_gist;'
+        env:
+          PGPASSWORD: postgres
+
+      - name: Setup Node.js environment
+        uses: actions/setup-node@v2
+        with:
+          node-version: 18
+
+      - name: Install dependencies
```
stwiname, line 85: Is it worth moving this to the script so we can have 1 place to update this to other sdks?

.github/workflows/benchmark-substrate.yml
```diff
@@ -0,0 +1,200 @@
...
+          # Use cut to extract the version from the line
+          version=$(echo "$version_line" | grep -oE '[0-9]+\.[0-9]+\.[0-9]+')
+          echo "::set-output name=version::$version"    
+
+
+      - name: Query metadata get processed height
+        id: query_meta
+        env:
+          PGPASSWORD: postgres
+        run: |
```
stwiname, line 126: Is it worth moving these scripts to separate files? Then we at least get syntax hilighting

stwiname (commented):
This looks awesome. Do you think there's a way we could get this to run automatically when release PRs are made?

jiqiang90:
> This looks awesome. Do you think there's a way we could get this to run automatically when release PRs are made?

Happy to do that, but need to give more tries first 

It suppose only triggered when new version node is published

https://dev.to/moaaz_adel/github-actions-a-simple-way-to-trigger-workflow-from-another-workflow-1b55

stwiname:
How do we limit the CPU and Memory to make sure we are benchmarking with the same resources each time?

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,184 @@
...
+          path: output/benchmark/
+
+
+      - name: Post to a Slack channel
+        id: slack
+        uses: slackapi/slack-github-action@v1.24.0
+        with:
+          # Slack channel id, channel name, or user id to post message.
+          # See also: https://api.slack.com/methods/chat.postMessage#channels
+          channel-id: 'C02156BV42V'
```
stwiname, line 167: Can you comment the name of this channel please. I cant remember what one to look in slack.

.github/workflows/scripts/benchmark/Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+
+ENV DB_USER=postgres
+ENV DB_PASS=postgres
+ENV DB_DATABASE=postgres
+ENV DB_HOST=postgres
+ENV DB_PORT=5432
+
+
+# Set CPU shares and memory limit based on build arguments
+CMD ["--cpus=$CPUS", "--memory=$MEMORY"]
```
stwiname, line 20: Does this work? Are we sure it doesn't just set the env var which would now report differently to the number of cpu?

My understanding is that these are set like `docker run --cpus=2 --memory=4g <rest-of-command>` which would mean setting here https://github.com/subquery/subql/blob/063701c727b733fd0480b6aa3671551863f0c73b/.github/workflows/benchmark.yml#L107

stwiname (commented):
Are you able to run a benchmark so we can see the latest changes working?

.github/workflows/scripts/benchmark/Dockerfile
jiqiang90: it been part of build options https://docs.docker.com/engine/reference/commandline/build/#options, and when `docker run` it would use this as part of the args. 

I can change this so easy for people to read.

.github/workflows/benchmark.yml
jiqiang90: It is a App(bot) we can invite to all channel

jiqiang90:
> Are you able to run a benchmark so we can see the latest changes working?

It is been forward in slack, https://github.com/jiqiang90/auction/actions/runs/5640430644

stwiname (approved):

PR #1895, opened 2023-07-21
Performance scoring fix

DeveloperInProgress:
# Description
Convert response time scoring connection pool to use logarithmic scaling to handle response time tending to 0. And move api connection error handling to node-core.

## Type of change

## Checklist

- [ ] I have tested locally


packages/node-core/src/api.connection.ts
```diff
@@ -0,0 +1,43 @@
...
+  static handleDisconnectionError(e: Error): ApiConnectionError {
+    const formatted_error = new ApiConnectionError('ConnectionError', e.message, ApiErrorType.Connection);
+    return formatted_error;
+  }
+
+  static handleLargeResponseError(e: Error): ApiConnectionError {
+    const newMessage = `Oversized RPC node response. This issue is related to the network's RPC nodes configuration, not your application. You may report it to the network's maintainers or try a different RPC node.\n\n${e.message}`;
+
+    return new ApiConnectionError('RpcInternalError', newMessage, ApiErrorType.Default);
+  }
```
stwiname, line 42: I think we can just define these specific error types as classes, they don't need to be static methods.

packages/node-core/src/api.connection.ts
```diff
@@ -0,0 +1,43 @@
...
+import {NetworkMetadataPayload} from './events';
+import {ApiConnectionError, ApiErrorType} from './indexer';
+
+export interface IApiConnectionSpecific<A = any, SA = any, B = any> extends IApi<A, SA, B> {
+  handleError(error: Error): ApiConnectionError;
+  apiConnect(): Promise<void>;
+  apiDisconnect(): Promise<void>;
+}
+
+export abstract class ApiConnectionSpecific<A = any, SA = any, B = any> implements IApiConnectionSpecific<A, SA, B> {
```
stwiname, line 14: When we remove the errors we can do without the abstract class

stwiname (approved):

PR #1899, opened 2023-07-25
Fix IPFS repeat cat same cid

jiqiang90:
# Description

We have an implementation in project service, some file has fetched and saved into local. However we still face when use promise.all fetch multiple assets within multiple datasource, it could fetch same cid simultaneously.

We add a cache to save promises and if it is waiting to be fetched we can avoid to fetch it again. 

Fixes # (issue)

## Type of change

## Checklist

- [ ] I have tested locally

### DIFF
--- a/packages/common/src/project/readers/ipfs-reader.spec.ts
+++ b/packages/common/src/project/readers/ipfs-reader.spec.ts
@@ -16,4 +16,11 @@ describe('IPFSReader', () => {
 
     expect(data.network.chainId).toBe('43114');
   });
+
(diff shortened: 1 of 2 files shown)


stwiname (commented):
If we're saving them to a tmp directory we should use that as a cache rather than in memory. Some of these files could be pretty large

ianhe8x:
> If we're saving them to a tmp directory we should use that as a cache rather than in memory. Some of these files could be pretty large

Should be fine, the ipfsReader will be gc after loading the project.

stwiname (approved):

PR #1900, opened 2023-07-25
fix unable query latest mmr issue

jiqiang90:
# Description

Fixed previous issue `MMR is completing,please try again later` when use `get` method from MMR.

This is due to getLeafLength always use dirty length https://github.com/subquery/merkle-mountain-range/blob/9d0f2fdf7fcf4e1c8a7231f4d41f7474ec403758/merkleMountainRange.js#L142

And it never get update after this mmr instance inited in mmrQuery service, therefore, each time when `get`, it will return [Error. ](https://github.com/subquery/merkle-mountain-range/blob/9d0f2fdf7fcf4e1c8a7231f4d41f7474ec403758/merkleMountainRange.js#L66)

This fix instead use current instance, but construct a new one to avoid this problem. 

Fixes # (issue)

## Type of change

## Checklist

- [ ] I have tested locally

### DIFF
--- a/packages/node-core/CHANGELOG.md
+++ b/packages/node-core/CHANGELOG.md
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- Fix mmrQuery should always read mmr leaf length from Db. 
+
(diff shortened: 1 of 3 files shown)


stwiname (commented):
Looks good to me but a test would be good.

stwiname (approved):

PR #1927, opened 2023-08-07
add primary network endpoint

### LINKED ISSUE #1924: Introduce primary network endpoint
Users have requested being able to specify a primary end point that is always used unless there are issues with that endpoint. 

We can introduce a new flag `--primary-network-endpoint` to achieve this. 

image (7)

DeveloperInProgress:
# Description
add `--primary-network-endpoint` to specify a primary end point that is always used unless there are issues with that endpoint

Fixes #1924 

## Type of change

## Checklist

- [ ] I have tested locally

### DIFF
--- a/packages/node/src/indexer/api.service.ts
+++ b/packages/node/src/indexer/api.service.ts
@@ -85,81 +85,100 @@ export class ApiService
       );
     }
 
-    const endpointToApiIndex: Record<string, ApiPromiseConnection> = {};
-
     if (chainTypes) {
       logger.info('Using provided chain types');
     }
 
     for await (const [i, endpoint] of network.endpoint.entries()) {
-      const connection = await ApiPromiseConnection.create(
-        endpoint,
-        this.fetchBlocksBatches,
-        {
-          chainTypes,
-        },
-      );
-
-      const api = connection.unsafeApi;
-
-      this.eventEmitter.emit(IndexerEvent.ApiConnected, {
-        value: 1,
-        apiIndex: i,
-        endpoint: endpoint,
-      });
-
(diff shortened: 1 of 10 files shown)


packages/node/src/indexer/api.service.ts
```diff
@@ -85,81 +85,100 @@ export class ApiService
...
     }
 
     for await (const [i, endpoint] of network.endpoint.entries()) {
-      const connection = await ApiPromiseConnection.create(
-        endpoint,
-        this.fetchBlocksBatches,
-        {
-          chainTypes,
-        },
+      const connection = await this.createAndValidateConnection(endpoint, i);
```
stwiname, line 93: Is there any way we can move some of this to node core? Ideally we don't need to make changes to node so that we can easily update other SDKs

packages/node/src/indexer/api.service.ts
```diff
@@ -72,6 +72,10 @@ export class ApiService
     try {
       chainTypes = this.project.chainTypes;
       network = this.project.network;
+
+      if (this.nodeConfig.primaryNetworkEndpoint) {
+        network.endpoint.push(this.nodeConfig.primaryNetworkEndpoint);
```
stwiname, line 77: How does connection pool service know this is a primary endpoint?

DeveloperInProgress: The check happens here:

https://github.com/subquery/subql/blob/6e57956ffebc52f7a97255edf519c1688b8fb10a/packages/node-core/src/indexer/connectionPool.service.ts#L54

stwiname (approved):

PR #1952, opened 2023-08-16
delay getPoiBlocksByRange when fully synced

jiqiang90:
# Description

Fixes # (issue)

Poi is keep make query when project is fully synced, this could be the cause of increase cpu usage for indexer and database.

When getPoiBlocksByRange didn't get more than `DEFAULT_FETCH_RANGE`, which means poi is almost fully synced, we add a delay here to ensure `getPoiBlocksByRange` only get called again after few seconds. 

## Checklist

- [ ] I have tested locally

### DIFF
--- a/packages/node-core/src/indexer/mmr.service.ts
+++ b/packages/node-core/src/indexer/mmr.service.ts
@@ -112,36 +112,30 @@ export class MmrService extends baseMmrService implements OnApplicationShutdown 
     this._poi = poi;
   }
 
-  private async syncPoiJob(logging?: boolean): Promise<void> {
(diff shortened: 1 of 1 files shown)


packages/node-core/src/indexer/mmr.service.ts
```diff
@@ -134,6 +140,9 @@ export class MmrService extends baseMmrService implements OnApplicationShutdown
         );
       }
     }
+    if (poiBlocks.length < DEFAULT_FETCH_RANGE) {
```
stwiname, line 143: I think it would make more sense for this to live in the while loop.

stwiname (approved):

PR #1966, opened 2023-08-22
Rework poi

### LINKED ISSUE #1905: Rework proof of index
### Description
Currently we offer two levels of cryptographic proof for index, 
1. `hash` in the poi table, which is `HashFn(height,chainBlockHash,parentHash,operationHash,projectId)`
    1. poi table is not continuous (only when there is operation in the block)
    2. parentHash is the hash of last block with operation
    3. if a block is skipped, we use a fixed hash for that block
2. We also build a mmr and calculate the root hash, because this proof is continuous we use if for the final proof

Theoretical, hash is enough to tell if two indexers are generating same data (because parenthash is included in the digest)
the only problem for it is not continuous.
So we can actually simplify poi by removing mmr and only use the hash, as long as we can solve the only blocking issue.

### Details
1. Make data in _poi table continuous
    1. We can keep the current place we insert poi record with block of operations, but we should leave `hash` empty
    2. Change where we used to insert mmr table to insert poi table, fill in empty blocks in the table. This step we can generate `hash` for records inserted in step 1
    3. For skipped blocks, we use empty hash for chainblockHash and operationHash, but the parentHash is no more the last block with operation, we use its parent block literally.
2. further tidy up _poi table, deprecating parentHash column since it is redundant data and now can be easily found by id.
3. change api, add `/poi/` and deprecate `/mmrs/`, change response body.

**Network Details:**
  * [Network]

### Steps to Reproduce

1. [First Step]
5. [Second Step]
6. [and so on...]

### Any other information

jiqiang90:
# Description

Fixes #1905 

## Type of change

## Checklist

- [x] I have tested locally


packages/node-core/src/indexer/poi/poi.service.ts
```diff
@@ -1,55 +1,261 @@
...
+    if (!this._latestSyncedPoi) {
+      throw new Error('_latestSyncedPoi in Poi service is not inited');
+    }
+    if (this._latestSyncedPoi.parentHash === undefined) {
+      throw new Error(`Invalid latest synced poi at height ${this._latestSyncedPoi.id}, parentHash is missing`);
+    }
+    if (this._latestSyncedPoi.hash === undefined) {
+      throw new Error(`Invalid latest synced poi at height ${this._latestSyncedPoi.id}, hash is missing`);
+    }
+    return <SyncedProofOfIndex>this._latestSyncedPoi;
```
stwiname, line 60: You could use a typeguard here which would avoid the cast:

```ts
function isSyncedProofOfIndex(poi?: ProofOfIndex | SyncedProofOfIndex): poi is SyncedProofOfIndex {
  return !!poi && (poi as SyncedProofOfIndex).parentHash && (poi as SyncedProofOfIndex).hash;
}
```

packages/node-core/src/indexer/poi/poi.service.ts
```diff
@@ -1,55 +1,261 @@
...
+  get projectId(): string {
+    if (!this._projectId) {
+      throw new Error(`No project Id inited`);
+    }
+    return this._projectId;
+  }
+
+  get latestSyncedPoi(): SyncedProofOfIndex {
+    if (!this._latestSyncedPoi) {
+      throw new Error('_latestSyncedPoi in Poi service is not inited');
```
stwiname, line 52:
```suggestion
      throw new Error('_latestSyncedPoi in Poi service is not initalized');
```

packages/node-core/src/indexer/poi/poi.service.ts
```diff
@@ -1,55 +1,261 @@
...
+    }
+    if (this._latestSyncedPoi.parentHash === undefined) {
+      throw new Error(`Invalid latest synced poi at height ${this._latestSyncedPoi.id}, parentHash is missing`);
+    }
+    if (this._latestSyncedPoi.hash === undefined) {
+      throw new Error(`Invalid latest synced poi at height ${this._latestSyncedPoi.id}, hash is missing`);
+    }
```
stwiname, line 59: What are the chances of these happening? If a user faced this error can we give them some instructions how to resolve the issue?

packages/node-core/src/indexer/poi/poi.service.ts
```diff
@@ -1,55 +1,261 @@
...
+  async init(schema: string, projectId: string): Promise<void> {
+    this._poiRepo = this.storeCache.poi ?? undefined;
+    this._projectId = projectId;
+    const latestSyncedPoiHeight = await this.storeCache.metadata.find('latestSyncedPoiHeight');
+    if (latestSyncedPoiHeight !== undefined) {
+      const recordedPoi = await this.poiRepo.getPoiById(latestSyncedPoiHeight);
+      if (recordedPoi) {
+        await this.setLatestSyncedPoi(recordedPoi);
+      } else {
+        // TODO, could use sql to find latestSyncedPoiHeight in db
```
stwiname, line 72: Still a TODO?

packages/node-core/src/indexer/poi/poi.service.ts
```diff
@@ -1,55 +1,261 @@
...
+            throw e;
+          }
+        }
+        logger.info(`Successful migrate Poi`);
+      }
+      // Remove and Change column from sequelize not work, it only applies to public schema
+      // https://github.com/sequelize/sequelize/issues/13365
+      // await this.poiRepo?.model.sequelize?.getQueryInterface().changeColumn(tableName,'mmrRoot',{})
+    } catch (e) {
+      throw new Error(`Failed to migrate poi table}`);
```
stwiname, line 150:
```suggestion
      throw new Error(`Failed to migrate poi table. {e}`);
```

packages/node-core/src/indexer/poi/poi.service.ts
```diff
@@ -1,55 +1,261 @@
...
+          } catch (e) {
+            logger.error(`Migration poi failed with query: ${`query`}`);
+            throw e;
+          }
+        }
+        logger.info(`Successful migrate Poi`);
+      }
+      // Remove and Change column from sequelize not work, it only applies to public schema
+      // https://github.com/sequelize/sequelize/issues/13365
+      // await this.poiRepo?.model.sequelize?.getQueryInterface().changeColumn(tableName,'mmrRoot',{})
```
stwiname, line 148: Can we run raw sql?

packages/node-core/src/indexer/poi/poi.service.ts
```diff
@@ -1,55 +1,261 @@
...
+          );
+        }
+      }
+
+      if (queries.length) {
+        for (const query of queries) {
+          try {
+            await this.poiRepo?.model.sequelize?.query(query, {type: QueryTypes.SELECT});
+          } catch (e) {
+            logger.error(`Migration poi failed with query: ${`query`}`);
```
stwiname, line 140:
```suggestion
            logger.error(`Migration poi failed with query: ${query}`);
```

packages/node-core/src/indexer/poi/poi.service.ts
```diff
@@ -1,55 +1,261 @@
...
+      } else {
+        // TODO, could use sql to find latestSyncedPoiHeight in db
+        throw new Error(`Can not find latestSyncedPoiHeight ${latestSyncedPoiHeight}`);
+      }
+    }
+    await this.migratePoi(schema);
+  }
+
+  // This could use for a subcommand later
+  async migratePoi(schema: string): Promise<void> {
```
stwiname, line 80: Does this need to clean up the MMR table? Or file based MMR?

packages/node-core/src/indexer/poi/poi.service.ts
```diff
@@ -1,55 +1,261 @@
...
+      }
+      this.poiRepo?.bulkUpsert(appendedBlocks);
+    }
+  }
+
+  private addDefaultPoiBlocks(endHeight: number, appendedBlocks: ProofOfIndex[]): void {
+    for (let i = this.latestSyncedPoi.id + 1; i <= endHeight; i++) {
+      const syncedPoiBlock = PoiBlock.create(i, null, null, this.projectId, this.latestSyncedPoi.hash);
+      appendedBlocks.push(syncedPoiBlock);
+      void this.setLatestSyncedPoi(syncedPoiBlock);
```
stwiname, line 258: Why is this void?

packages/node-core/src/indexer/poi/poi.service.ts
jiqiang90: It cleans up mmr table if exist when migration in the first time. 

For fileBased mmr, because we have removed relate flags for mmr, we could ask user remove it manually.

packages/node-core/src/indexer/poi/poi.service.ts
jiqiang90: Current implementation is using raw sql, I left this comment here due to this bug here, we could not alter poi table columns with sequelize method `getQueryInterface().changeColumn`

stwiname:
It would be good for Poi Service to have coverage over 80%
Screenshot 2023-08-24 at 3 01 59 PM

packages/node-core/src/indexer/poi/poi.service.ts
```diff
@@ -1,55 +1,261 @@
...
-    if (!this.latestPoiBlockHash || !isMainThread) {
-      const poiBlockHash = await this.fetchPoiBlockHashFromDb();
-      if (poiBlockHash === null || poiBlockHash === undefined) {
-        this.latestPoiBlockHash = DEFAULT_PARENT_HASH;
+  async syncPoi(exitHeight?: number, logging?: boolean): Promise<void> {
+    if (this.isSyncing) return;
+    this.isSyncing = true;
+    while (!this.isShutdown) {
+      if (!this._latestSyncedPoi) {
+        await delay(10);
```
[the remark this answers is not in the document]
jiqiang90: If _latestSyncedPoi is undefined, which means we haven't sync any Poi yet, no genesis poi been found.  It delay for genesis poi to be created, delay will stop after `ensureGenesisPoi`, https://github.com/subquery/subql/blob/bfe6de09f6743db219f5e46f2022bf41711feb3a/packages/node-core/src/indexer/poi/poi.service.ts#L160

packages/node-core/src/indexer/blockDispatcher/base-block-dispatcher.ts
```diff
@@ -197,27 +190,21 @@ export abstract class BaseBlockDispatcher<Q extends IQueue, DS> implements IBloc
     }
   }
 
-  private async updatePOI(height: number, blockHash: string, operationHash: Uint8Array): Promise<void> {
+  // First creation of POI
+  // TODO, add test here, even block been fetched, if no operation, we still treat it as empty block
```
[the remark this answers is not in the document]
jiqiang90: no longer need

packages/node-core/src/indexer/poi/poi.service.ts
```diff
@@ -1,55 +1,261 @@
...
 
-  async getLatestPoiBlockHash(): Promise<Uint8Array> {
-    if (!this.latestPoiBlockHash || !isMainThread) {
-      const poiBlockHash = await this.fetchPoiBlockHashFromDb();
-      if (poiBlockHash === null || poiBlockHash === undefined) {
-        this.latestPoiBlockHash = DEFAULT_PARENT_HASH;
+  async syncPoi(exitHeight?: number, logging?: boolean): Promise<void> {
+    if (this.isSyncing) return;
+    this.isSyncing = true;
+    while (!this.isShutdown) {
```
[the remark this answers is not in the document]
jiqiang90: Good point !

packages/node-core/src/indexer/poi/poi.service.ts
stwiname: Maybe we can log that they should clean it up?

packages/node-core/src/indexer/poi/poi.service.ts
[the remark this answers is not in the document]
stwiname: I don't think we should have random delays like this. Its near impossible to follow code with things like that. e.g. if for some reason `ensureGenesisPoi` fails or a refactor means its not called we could be scratching our heads for hours trying to figure out why this isn't running. 

We're better off calling functions in the correct order

packages/node-core/src/indexer/poi/poi.service.ts
jiqiang90: the `setLatestSyncedPoi ` provide any option to force flush the cache, so we only await if we need to force flush.

When we create default poi blocks with a continuous range height, we can just use void here.

stwiname: Could we just set it at the end of the for loop?

jiqiang90: it get parentHash from `this.latestSyncedPoi`, so it need to be set at every single block

packages/node-core/src/indexer/poi/poi.service.ts
jiqiang90: We don't have any feature to reset syncedPoi, once we have this feature, we could log here let user to reset to previous valid poi height, also complete https://github.com/subquery/subql/blob/bfe6de09f6743db219f5e46f2022bf41711feb3a/packages/node-core/src/indexer/poi/poi.service.ts#L72

packages/node-core/src/indexer/poi/poi.service.ts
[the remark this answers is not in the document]
jiqiang90: Genesis Poi need to create one time only. 
Once in metadata stored `latestSyncedPoi` and in the `init` will set this object, the sync process will immediately start. 

 Before I think we could make sure sync process start only after when the non-null operation block been processed, rather than checking every 10 seconds, but we also could face an issue here: 
Assume we restart the app, in current poi, latestSyncedPoi is 1000, and we have poi 1002,1005 awaiting to be synced. 
Then after 1005 ,there is no blocks with operation. which means 1002 and 1005 will be never been synced.. 

In current implementation will solve this issue. 

I will leave a comment with this method

packages/node-core/src/indexer/poi/poi.service.ts
ianhe8x: Can we treat this same as `_latestSyncedPoi` not exist?

packages/node-core/src/indexer/poi/PoiBlock.ts
```diff
@@ -12,57 +12,84 @@ import {
...
   readonly id: number;
-  readonly chainBlockHash: Uint8Array;
-  readonly hash: Uint8Array;
-  readonly parentHash: Uint8Array;
-  readonly operationHashRoot: Uint8Array;
+  readonly chainBlockHash: Uint8Array | null;
+  readonly operationHashRoot: Uint8Array | null;
   readonly projectId: string;
+  readonly parentHash: Uint8Array | undefined;
+  readonly hash: Uint8Array | undefined;
```
stwiname, line 48: I found these names confusing, i thought they referred to the block hashes but they are actually the POI hashes.

packages/node-core/src/indexer/poi/PoiBlock.ts
```diff
@@ -12,57 +12,84 @@ import {
...
 
   constructor(
     id: number,
-    chainBlockHash: Uint8Array,
-    hash: Uint8Array,
-    parentHash: Uint8Array,
-    operationHashRoot: Uint8Array,
-    projectId: string
+    chainBlockHash: Uint8Array | null,
```
stwiname, line 52: Could we change this to a string? Currently we need to ensure we can convert any block hash format to Uint8Array so when we add support for a new chain if they provide a different encoding we need to make changes here. 

It would help remove the big if/else block in create

packages/node-core/src/indexer/poi/poi.service.ts
stwiname: So the code above this is used instead of this?

stwiname (commented):
It would be good to have some documentation of how POI works, maybe just some comments in PoiBlock.ts 

The other concern I have is that we're computing POI for blocks that have no operations. Optimism/Arbitrum for example have 100M+ blocks. That is a large amount of data to store just for POI. This is due to wanting to query the POI at any block height. 

Is there something better we could do here? Like:
* Say the POI for any height is the same as the last height there was an operation
* Compute the POI at a specific height when requested as opposed to storing it and it never being queried. 
    * We could also have checkpoints every N blocks so we don't have to compute too much

packages/node-core/src/indexer/poi/PoiBlock.ts
jiqiang90: I added some comments for this

packages/node-core/src/indexer/poi/PoiBlock.ts
jiqiang90: This will require convert any exiting chainBlockHash in db convert from `bytea` to `text`. It will increase complexity for migration, we can leave this now

packages/node-core/src/indexer/poi/poi.service.ts
```diff
@@ -29,9 +30,12 @@ export class PoiService implements OnApplicationShutdown {
   private _poiRepo?: CachePoiModel;
   private _latestSyncedPoi?: ProofOfIndex;
   private isSyncing = false;
-  private _projectId?: string;
 
-  constructor(private storeCache: StoreCacheService, private eventEmitter: EventEmitter2) {}
+  constructor(
+    private storeCache: StoreCacheService,
+    private eventEmitter: EventEmitter2,
+    private project: ISubqueryProject
```
stwiname, line 37: You can use `@Inject('ISubqueryProject')` here instead of having to use a factory

stwiname (approved):

PR #1443, opened 2022-11-30
Switch cluster endpoint from cluster api to normal api

stwiname:
# Description
The updated IPFS endpoint has been restricted and the cluster api no longer works. This changes to using the normal api

## Type of change

Please delete options that are not relevant.

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist

- [x] I have tested locally
- [x] I have performed a self review of my changes
- [x] Updated any relevant documentation
- [x] Linked to any relevant issues
- [ ] I have added tests relevant to my changes
- [x] Any dependent changes have been merged and published in downstream modules
- [x] My code is up to date with the base branch

PR #1967, opened 2023-08-23
CLI Publish improvements

stwiname:
# Description
- Switch from cluster api to normal api (need to check that pinning works)
- Fix uploading files twice by uploading manifest and deployment version
- Fix fileMap type not caching promisies causing repeated file references to be uploading multiple times
- Update to latest ipfs-http-api that doesn't use ESM
- Replaces https://github.com/subquery/subql/pull/1443

## Type of change

## Checklist

- [x] I have tested locally

### DIFF
--- a/packages/common/src/constants.ts
+++ b/packages/common/src/constants.ts
@@ -8,7 +8,7 @@ export const DEFAULT_PORT = 3000;
 export const IPFS_READ = 'https://unauthipfs.subquery.network';
 export const IPFS_WRITE = 'https://authipfs.subquery.network';
 export const IPFS_NODE_ENDPOINT = `${IPFS_READ}/ipfs/api/v0`;
-export const IPFS_CLUSTER_ENDPOINT = `${IPFS_WRITE}/cluster/add`;
+export const IPFS_CLUSTER_ENDPOINT = `${IPFS_WRITE}/ipfs/api/v0`;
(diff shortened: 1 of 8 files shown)


packages/common/src/constants.ts
```diff
@@ -8,7 +8,7 @@ export const DEFAULT_PORT = 3000;
 export const IPFS_READ = 'https://unauthipfs.subquery.network';
 export const IPFS_WRITE = 'https://authipfs.subquery.network';
 export const IPFS_NODE_ENDPOINT = `${IPFS_READ}/ipfs/api/v0`;
-export const IPFS_CLUSTER_ENDPOINT = `${IPFS_WRITE}/cluster/add`;
+export const IPFS_CLUSTER_ENDPOINT = `${IPFS_WRITE}/ipfs/api/v0`;
```
ianhe8x, line 11: this doesn't look like an ipfs_cluster endpoint

stwiname: I've updated the pinning now. I think this is the right way to do it.

ianhe8x: Then let's not call it `IPFS_CLUSTER_ENDPOINT`, maybe `IPFS_WRITE_ENDPOINT`

ianhe8x (approved):

Issue #1889, opened 2023-07-18
Provide an option to skip fetch block & extrinsics for event handler


ianhe8x:
### Description

For `substrate/EventHandler` & `ethereum/LogHandler`, we always fetch the corresponding block and transaction, so we can offer access from the wrapped event object.
What if user has no use of things like `event.block`, `event.extrinsic` or `event.transaction`, can we only fetch header without the full detailed blocks and transactions?

### Details
* Add a new runner option for node, `skipBlock(default: false)`
    * when enabled, if project only contains `substrate/EventHandler` or `ethereum/LogHandler`, then node will not fetch block. 
    * for substrate project, extrinsics will not be fetched as well, since extrinsics are embedded in the block.
* Add a new runner option for node, `skipTransaction(default: false)`
    * when enabled, if project only contains `ethereum/LogHandler`, then node will not fetch transactions.
    * enable `skipTransaction` implies `skipBlock`, `skipBlock is set to false` will cause a conflict and should leads to exit(1)
* when project has other type of handlers that don't work with skipBlock or skipTransaction, but user set them true. the node will print a warning to console to tell user that these two options will be ignored.

These details can help to reproduce the environment the issue is occurring

**Local Environment:** [You can get this information from executing `subql version`.]
**Query Version:** [What is the version of the query service?]
**Indexer Version:** [What is the version of the indexer service?]
**Network Details:**
  * [Network]
  * [Block height, a block height where the issue is happening]
  * [Dictionary endpoint, if used]

### Steps to Reproduce

1. [First Step]
2. [Second Step]
3. [and so on...]

**Example project:** [A link to a minimal example that can reproduce the issue]

**Expected behavior:** [What you expected to happen]

**Actual behavior:** [What actually happened]

### Any other information

Is there any other information you would like to add?

stwiname:
Im going to close this. We have a minimal implementation for Substrate and Ethereum, there can be further improvements with Ethereum and other SKDs but that will have another ticket for them

PR #1968, opened 2023-08-23
Add skipBlock option to improve performance with event handlers

stwiname:
# Description
If a project only has event handlers there is not always a need to fetch all the block content. This can lead to ~10% performance increase in indexing speed.

`skipBlock` can be set in node runner options or in cli flags. It will automatically disable this if handlers other than event handlers are found.

I also found a bug where node runner options in project.yaml were broken because of default values in yargs. This will need to be fixed in other SDKs

Completes https://github.com/subquery/subql/issues/1889 for Substrate

## Type of change

- [x] This change requires a documentation update https://github.com/subquery/documentation/pull/400

## Checklist

- [x] I have tested locally
- [x] Updated any relevant documentation https://github.com/subquery/documentation/pull/400

### DIFF
--- a/packages/node/src/utils/project.ts
+++ b/packages/node/src/utils/project.ts
@@ -11,11 +11,17 @@ import {
   SubstrateCustomHandler,
   SubstrateHandler,
   SubstrateHandlerKind,
+  isRuntimeDs,
+  isCustomDs,
 } from '@subql/common-substrate';
 import { saveFile } from '@subql/node-core';
 import { SubstrateDatasource } from '@subql/types';
 import yaml from 'js-yaml';
 import { NodeVM, VMScript } from 'vm2';
+import {
+  SubqueryProject,
+  SubstrateProjectDs,
+} from '../configure/SubqueryProject';
 
 export function isBaseHandler(
   handler: SubstrateHandler,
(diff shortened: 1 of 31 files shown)


packages/node/src/utils/project.ts
```diff
@@ -112,3 +118,26 @@ export function loadChainTypesFromJs(
...
+      (handler) => handler.kind !== SubstrateHandlerKind.Event,
+    );
+  } else if (isCustomDs(ds)) {
+    // TODO this can be improved upon in the future.
+    return true;
+  }
+  return true;
+}
+
+export function isOnlyEventHandlers(project: SubqueryProject): boolean {
```
jiqiang90, line 134: We could define event only handler in this way, but we could face another issue, we could not know whether user try to access block content from event which we have wrapped previously. 

const wrappedEvents = wrapEvents(wrappedExtrinsics, events, wrappedBlock);

We can test this scenarios with starter, and we found error was thrown, but not clearly enough, could be very difficult to debug.

https://github.com/subquery/subql-starter/blob/96b09b6aa3a0c6506c423acf628f117cd47e9f83/Polkadot/Polkadot-starter/src/mappings/mappingHandlers.ts#L41

image

I think we could implement within wrapped event, if block content is accessed in handler, we could provide more friendly warning/error. Or even refetch the block content.

stwiname: This is why there is the `skipBlock` option, users have to opt into it and test it themselves. They should update their handler types accordingly

jiqiang90 (approved):

PR #1969, opened 2023-08-23
add block filter to common

DeveloperInProgress:
# Description
Block filter is the same for all SDKs so, move it to the common package

Fixes # (issue)

## Type of change

## Checklist

- [ ] I have tested locally


packages/common/src/project/models.ts
```diff
@@ -0,0 +1,14 @@
+// Copyright 2020-2023 SubQuery Pte Ltd authors & contributors
+// SPDX-License-Identifier: GPL-3.0
+
+import {IsString, IsOptional, IsInt} from 'class-validator';
+import {BlockFilter} from './types';
+
+export class BlockFilterImpl implements BlockFilter {
```
stwiname, line 7: Why is this not under versions?

packages/common/src/project/versioned/v1_0_0/types.ts
```diff
@@ -26,6 +26,10 @@ export interface NodeOptions {
   unfinalizedBlocks?: boolean;
 }
 
+export interface FileReference {
```
stwiname, line 29: Not sure why these have moved, they don't relate to the PR

stwiname (approved):

PR #1970, opened 2023-08-24
handle failed api inits

DeveloperInProgress:
# Description
Avoid exiting process when one of the API initialization fails. Instead mark it is as unusable and continue initializing other APIs.

Fixes # (issue)

## Type of change

## Checklist

- [ ] I have tested locally


packages/node-core/src/indexer/connectionPoolState.manager.ts
```diff
@@ -53,17 +54,23 @@ export class ConnectionPoolStateManager<T extends IApiConnectionSpecific<any, an
...
+
     const poolItem: ConnectionPoolItem<T> = {
       primary: primary,
       performanceScore: 100,
       failureCount: 0,
       endpoint: endpoint,
       backoffDelay: 0,
       rateLimited: false,
-      failed: false,
+      failed: initFailed,
```
stwiname, line 70: Do we need to add `initFailed` can't we use this?

packages/node-core/src/indexer/connectionPoolState.manager.ts
```diff
@@ -79,9 +86,15 @@ export class ConnectionPoolStateManager<T extends IApiConnectionSpecific<any, an
...
     }
 
-    const indices = Object.keys(this.pool)
+    const initedIndices = Object.keys(this.pool)
       .map(Number)
-      .filter((index) => !this.pool[index].backoffDelay && this.pool[index].connected);
+      .filter((index) => !this.pool[index].initFailed);
+
+    if (initedIndices.length === 0) {
+      throw new Error(`Initialization failed for all endpoints. Please add healthier endpoints.`);
```
stwiname, line 94: This doesn't seem like the right place to throw this error. This isn't an init function

packages/node/src/indexer/api.service.ts
```diff
@@ -96,66 +96,71 @@ export class ApiService
...
+            );
+          }
         }
-      }
 
-      endpointToApiIndex[endpoint] = connection;
+        endpointToApiIndex[endpoint] = connection;
+      } catch (e) {
+        logger.error(`failed to init ${endpoint}: ${e}`);
+        endpointToApiIndex[endpoint] = null;
```
stwiname, line 162: Can we move some of this to node core? Its pretty much the same on all chains

stwiname (commented):
If the initial connection fails will it ever be retried again?

DeveloperInProgress:
> If the initial connection fails will it ever be retried again?

I'm not sure how to handle this when there are multiple workers involved. the initialization may fail in a subset of workers and we should only allow using an endpoint when all of the workers are connected to it.

stwiname:
> > If the initial connection fails will it ever be retried again?
> 
> I'm not sure how to handle this when there are multiple workers involved. the initialization may fail in a subset of workers and we should only allow using an endpoint when all of the workers are connected to it.

Why would we only use an endpoint if all workers are connected? The shared state is the weighting not what endpoints can be used

packages/node-core/src/api.service.ts
```diff
@@ -58,6 +62,104 @@ export abstract class ApiService<A = any, SA = any, B = any> implements IApi<A,
...
+    getChainId: (connection: IApiConnectionSpecific) => Promise<string>,
+    network: ProjectNetworkConfig & {chainId: string},
+    index: number,
+    endpoint: string,
+    attempt: number,
+    postConnectedHook?: (connection: IApiConnectionSpecific, endpoint: string, index: number) => void
+  ): void {
+    if (attempt < 5) {
+      //eslint-disable-next-line @typescript-eslint/no-misused-promises
+      setTimeout(async () => {
```
stwiname, line 129: It might be worth having a backoff function as we use the same concept a few places now. Also there needs to be a reference to this timeout

packages/node-core/src/indexer/connectionPool.service.ts
```diff
@@ -62,12 +62,25 @@ export class ConnectionPoolService<T extends IApiConnectionSpecific<any, any, an
     }
   }
 
+  async updateConnection(api: T, index: number): Promise<void> {
+    if (index >= this.allApi.length) {
+      throw new Error(`Attempting to update connection that does not exist.`);
+    }
+
+    this.allApi[index] = api;
```
stwiname, line 70: Do you think this is safe? What happens if the wrong index is provided?

stwiname (commented):
This is looking like a good improvement

packages/node-core/src/indexer/connectionPool.service.ts
DeveloperInProgress: 1.) the order of the api will be different for different workers, making the connection pool state invalid because only the index is used to reference api. 
2.) one api might overwrite another

so, it's important that we get the index right.

stwiname: Yes its important to get the index right. Is using an int the right thing to key these by? If we use the endpoint it would be much safer

packages/node/src/indexer/api.service.ts
```diff
@@ -47,8 +48,9 @@ export class ApiService
     connectionPoolService: ConnectionPoolService<ApiPromiseConnection>,
     eventEmitter: EventEmitter2,
     private nodeConfig: NodeConfig,
+    retryManager: RetryManager,
```
stwiname, line 51: I was just meaning a util function. Adding another dependency ads a lot of unnecessary complexity. 

Its already hard enough to write tests with ApiService because we need all the connectionPool stuff. We should be able to mock an api service without all that when we want a single endpoint. But thats another issue that shouldn't be addressed here

stwiname (changes requested):

stwiname (approved):
Awesome! I think changing from index to endpoint was a good idea, it seems to mak code more readable

stwiname:
I tried this by adding `wss://fake.example.com/public-ws` to the polkadot starter and the project wont start. The OnFinality endpoint is there still

packages/node-core/src/api.service.ts
```diff
@@ -98,24 +99,31 @@ export abstract class ApiService<A = any, SA = any, B extends Array<any> = any[]
...
         failedConnections.set(i, endpoint);
+        throw e;
       }
-    }
-
-    if (Object.values(endpointToApiIndex).every((value) => value === null)) {
+    });
+
+    try {
+      const {fulfilledIndex, result} = (await raceFulfilled(connectionPromises)) as {
```
stwiname, line 112: Why is there a cast needed here?

DeveloperInProgress: to rule out `undefined`

stwiname (approved):

PR #1987, opened 2023-09-03
block height filter

### LINKED ISSUE #1962: Allow Customers to set Initial Store State on Project Genesis
### Why

Customers want to be able to insert 'genesis data' into their SubQuery project. This could include seed data or constants that they want to reference.
 
### Description

Create some kind of Genesis handler that is always run when a project starts for the first time on a fresh database.
- Create genesis data in Subquery projects.
- Get called before all other handlers
- Only called once , in metadata we could record if it been called once, it should never call again. 
- Should have access to all standard store queries

jiqiang90:
# Description

Add a height filter for block handler, allow trigger mapping at specific height. 

Also can use a genesis handler, eg 

```yaml
        - handler: handleGenesisBlock
          kind: substrate/BlockHandler
          filter:
            height: 1
#            modulo: 50

```

This much handy than only a genesis handler 

Fixes #1962

TODO:

- [ ] Unit test

## Type of change

## Checklist

- [x] I have tested locally

### DIFF
--- a/packages/common/src/project/versioned/v1_0_0/models.ts
+++ b/packages/common/src/project/versioned/v1_0_0/models.ts
@@ -63,6 +63,9 @@ export class BlockFilterImpl implements BlockFilter {
   @IsOptional()
   @IsString()
   timestamp?: string;
+  @IsOptional()
+  @IsInt()
+  height?: number;
 }
 
 export class ParentProjectModel implements ParentProject {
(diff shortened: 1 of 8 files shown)


jamesbayly:
Can we make sure this still allows dictionaries to be used, e.g. the presence of a block handler with a specific block height does not stop dictionaries

jiqiang90:
> Can we make sure this still allows dictionaries to be used, e.g. the presence of a block handler with a specific block height does not stop dictionaries

Dictionary still can be used, same as the modulo filter

stwiname:
I think we can close this. Users can achieve the same result with the [endBlock feature](https://github.com/subquery/subql/pull/2064)

Issue #1756, opened 2023-05-25
Improved GetByField


jiqiang90:
```
enum Expressions = {
'=',
'<',
'<=',
'>',
'>='
'in'
}
```

`getByFields(field_name: string, value: any, expression: Expression = '=')`
`getByFields(field_name: string, value: any, expression: Expression, field_name: string, value: any, expression: Expression)`

Always treat multiple conditions as an AND

jiqiang90:
https://github.com/subquery/subql/issues/1531 Relate to this task

ianhe8x:
@jiqiang90 are you proposing a chained builder pattern query method?
```
store.buildQuery().getByFields('field1', value).getByFields('field2',value).query();
```
They is new to the project and may cause some code style concerns.
We could just do this, align with the gql filter styles (and simpler)
```
getByFields({
[field]: {
  [comparer]: value
  } 
})
```

PR #1993, opened 2023-09-05
Add getByFields function to store interface

stwiname:
# Description
Adds the ability to query entities from mapping handlers by multiple fields. When this is used, it will flush the cache if there is any pending data for the entity. This provides a single source of truth so we don't need to combine cache data with db data. This could lead to some slower performance if used.

Example usage:
```ts
const res = await store.getByFields('Extrinsic', [
    ['module', '=', extrinsic.extrinsic.method.section],
    ['call', '=', extrinsic.extrinsic.method.method]
], { limit: 5 });

const res = await Extrinsic.getByFields([
    ['module', '=', extrinsic.extrinsic.method.section],
    ['call', '=', extrinsic.extrinsic.method.method]
]);

const res = await Extrinsic.getByFields([
    ['module', 'in', ['timestamp', 'parachains']],
    ['call', 'in', ['set', 'setHeads']],
  ], { limit: 5 });
```

Fixes https://github.com/subquery/subql/issues/1756

## Type of change

- [x] This change requires a documentation update https://github.com/subquery/documentation/pull/409

## Checklist

- [x] I have tested locally
- [x] Updated any relevant documentation https://github.com/subquery/documentation/pull/409

### DIFF
--- a/packages/cli/CHANGELOG.md
+++ b/packages/cli/CHANGELOG.md
@@ -11,12 +11,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - Support for cosmwasm contract abi codegen (#1989)
-
-### Added
 - Update `subql init` to use proxy templates (#1991)
 
 ### Changed
 - Update common packages version from `latest`, to ensure they will be bumped (#1992)
+- Update model codegen to include `getByFields` and produce prettier code (#1993)
 
 ## [3.6.1] - 2023-08-29
 ### Fixed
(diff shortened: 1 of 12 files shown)


jiqiang90:
I'm happy with this,  as currently we only support `AND` here https://github.com/subquery/subql/blob/d5da407cc7bcd52fba717fdf4378de43de85fd5c/packages/node-core/src/indexer/storeCache/cacheModel.ts#L179, we need to mention this in the documents too.

Will be nice if we can provide `Op.or` in the future, current implementation will require user to make multiple requests and then aggregate the results.

jiqiang90 (approved):

PR #1997, opened 2023-09-06
migrate to oclif v2

### LINKED ISSUE #1990: Look into dependency sizes for CLI and project
The polkadot start installs about 800MB of dependencies right now. This includes the CLI as a dependency. We should look at reducing this.

This file is the output of `du -h -d 1 | sort -n` on `node_modules` with the sorting fixed with the units
[modules_size.txt](https://github.com/subquery/subql/files/12517022/modules_size.txt)

Starting with the larger dependencies I think there are a few easy wins, these changes will probably need to be made to `common` and `types` packages for each chain.

* [x] Ensure dependencies are actual required dependencies, if not they should be moved to devDependencies (e.g prettier)
* [x] Update oclif and use correct sub packages. Currently we have `aws-cli` as a transient dependency which is 90MB
* [x] Move ethereum codegen to `@subql/common-ethereum`. This will ensure versions of ethers remain the same between different packages
* [x] More specific imports. Do we need `@polkadot/api` as a dependency if we only use `@polkadot/types`
* [x] Investigate a way to make the CLI more modular, currently it supports every chain which brings along the associated sdks for those chains. [typechain](https://github.com/dethcrypto/TypeChain) could be a good example where you need to also install the target
* [ ] Look at bundling the CLI into a single js file (e.g webpack)

bz888:
# Description
- Migrating from oclif v1 to v1
- Removed Oclif cli library to use oclif/core
- Removed oclif/command and other oclif packages as in v2 they exist in oclif/core
- Refactor in accordance with v2 migration (Flags.enum was removed and replaced with custom, however, I struggle with implementing custom)

- move more packages into dev dependencies

Fixes #1990

## Type of change

## Checklist

- [x] I have tested locally


packages/cli/src/utils.ts
```diff
@@ -103,3 +103,7 @@ export function resolveToAbsolutePath(inputPath: string): string {
   const regex = new RegExp(`^~(?=$|[/\\\\])`);
   return path.resolve(inputPath.replace(regex, os.homedir()));
 }
+
+export function isValidateEnum<T extends Record<string, string>>(enumType: T, input: string): input is T[keyof T] {
```
stwiname, line 107: Can you write a test for this please

stwiname (approved):

PR #2005, opened 2023-09-07
update endpoint health listener

### LINKED ISSUE #1965: Update Ethereum SDK health checks
### Description

With some layer 2 networks block finalization behaves differently to other chains. E.g Arbitrum will finalize ~1000 blocks at once every few minutes. This is dependent on a layer 1 transaction confirming. You can see this activity here https://arbiscan.io/batches.

The same thing happens with Optimism https://optimistic.etherscan.io/batches

### Changes

We should update the health check on the node to accommodate for this. We currently expect the finalized height to be increasing at a block interval but this isn't the case with these networks.

bz888:
# Description
Due to networks like Arbitrum and Optimism where the latest finalized block is batched and outputs every couple of mins. 
The endpoint health should not be listening to the latestFinalized block, rather it should be listening to the latest unFinalizedBlock

Fixes #1965

## Type of change

## Checklist

- [x] I have tested locally

### DIFF
--- a/packages/node-core/src/meta/health.service.ts
+++ b/packages/node-core/src/meta/health.service.ts
@@ -44,6 +44,8 @@ export class HealthService {
     }
   }
 
+  // Not all networks have a consistent interval for Finalized Blocks
+  // Hence, the endpoint's health will be dictated by the best block (as it is a consistent interval)
   @OnEvent(IndexerEvent.BlockBest)
   handleTargetBlock(blockPayload: TargetBlockPayload): void {
     if (this.recordBlockHeight !== blockPayload.height) {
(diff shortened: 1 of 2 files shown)


packages/node-core/src/meta/health.service.ts
```diff
@@ -44,7 +44,7 @@ export class HealthService {
     }
   }
 
-  @OnEvent(IndexerEvent.BlockTarget)
+  @OnEvent(IndexerEvent.BlockBest)
```
stwiname, line 47: Can you please add a comment why we've chosen this block

stwiname (approved):

Issue #1960, opened 2023-08-19
new dictionary limit flag, decouple with batchSize


ianhe8x:
### Prerequisites

* [ ] Are you running the latest version(s)?
* [ ] Have you searched the [documentation for your issue](https://academy.subquery.network/)?
* [ ] Did you search for an existing issue or pull request?

### Description
When we construct dictionary query, we set `first: batchSize.toString()`, which may leads to more query requests if batchSize is small.
The thing is we change batchSize mostly considering memory usage and rpc ratelimits, not dictionary queries.

### Details
We should use a separated flag for dictionary limit, and default it to a larger value like 100

PR #2015, opened 2023-09-12
support cursor pagination in dictionary query

### LINKED ISSUE #1815: Investigate updating queries to use cursors for dictionaries
It is best practice to use [graphql cursors](https://relay.dev/graphql/connections.htm) as a way of paginating results.

We should test updating the queries to use cursors to see if this can improve query performance. This may also need some updates to the query service to make sure that cursors are being used properly

DeveloperInProgress:
# Description
It is best practice to use [graphql cursors](https://relay.dev/graphql/connections.htm) as a way of paginating results (instead of take/skip).
We should update subql-node to use cursors instead of other pagination tools, and also allow it to fall back to the old way if not supported

Fixes #1815 

## Type of change

## Checklist

- [ ] I have tested locally


packages/node-core/src/indexer/dictionary.service.ts
```diff
@@ -112,6 +112,60 @@ function extractVars(entity: string, conditions: DictionaryQueryCondition[][]):
...
+  };
+
+  if (afterCursor) {
+    //eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    node.args!.after = afterCursor;
+  }
+
+  if (useDistinct) {
+    // Non null assertion here because we define the object explicitly
+    assert(node.args, 'Args should be defined in the above definition of node');
```
stwiname, line 162: Maybe we can add this check for the afterCursor too

packages/node-core/src/indexer/dictionary.service.ts
```diff
@@ -112,6 +112,60 @@ function extractVars(entity: string, conditions: DictionaryQueryCondition[][]):
...
+        entity: 'pageInfo',
+        project: ['hasNextPage', 'endCursor'],
+      },
+    ],
+    args: {
+      filter: {
+        ...filter,
+        blockHeight: {
+          greaterThanOrEqualTo: `"${startBlock}"`,
+          lessThan: `"${endBlock}"`,
```
stwiname, line 148: I don't think this will work. Having a cursor means there shouldn't be an end.

packages/node-core/src/indexer/dictionary.service.ts
```diff
@@ -112,6 +112,60 @@ function extractVars(entity: string, conditions: DictionaryQueryCondition[][]):
...
+      {
+        entity: 'pageInfo',
+        project: ['hasNextPage', 'endCursor'],
+      },
+    ],
+    args: {
+      filter: {
+        ...filter,
+        blockHeight: {
+          greaterThanOrEqualTo: `"${startBlock}"`,
```
stwiname, line 147: It would be great to convert this to a graphql variable

packages/node-core/src/indexer/dictionary.service.ts
```diff
@@ -250,45 +305,65 @@ export class DictionaryService {
...
         }
-      }
-      const _metadata = resp.data._metadata;
+
+        _metadata = resp.data._metadata;
+
+        if (!this.dictionaryValidation(_metadata, startBlock)) {
+          return undefined;
+        }
+      } while (this.supportsCursor && hasNextPage);
```
stwiname, line 348: This isn't the right approach. The goal is to always use the following cursor unless the datasources changes. So each time `scopedDictionaryEntries` is called it should use the previous cursor if the datasources haven't changed.
This would align with my other comment that there should be no end block.

packages/node-core/src/indexer/dictionary.service.ts
```diff
@@ -250,45 +305,65 @@ export class DictionaryService {
...
-      }
-
       return {
         _metadata,
         batchBlocks,
       };
     } catch (err: any) {
-      // Check if the error is about distinct argument and disable distinct if so
-      if (JSON.stringify(err).includes(distinctErrorEscaped)) {
-        this.useDistinct = false;
```
stwiname, line 289: This functionality has disappeared

packages/node-core/src/indexer/dictionary.service.ts
```diff
@@ -250,45 +305,65 @@ export class DictionaryService {
...
-        logger.warn(`Dictionary doesn't support distinct query.`);
-        // Rerun the qeury now with distinct disabled
+      // If server doesn't support cursor-based queries, fallback to non cursor-based
+      if (
+        JSON.stringify(err).includes('Cannot query field "cursor"') ||
```
stwiname, line 362: We should use a const for these messages.

packages/node-core/src/indexer/dictionary.service.ts
```diff
@@ -250,45 +305,65 @@ export class DictionaryService {
...
+          this.client.query({
+            query: gql(query),
+            variables: {...variables, afterCursor},
+          }),
+          this.nodeConfig.dictionaryTimeout
+        );
+
+        for (const entity of Object.keys(resp.data)) {
+          if (entity !== '_metadata') {
+            const data = this.supportsCursor ? resp.data[entity].edges : resp.data[entity].nodes;
```
stwiname, line 330: Rather than having various checks for `supportsCursor` it would be more readable to have a single check and keep the differences separate

stwiname:
This needs tests too

packages/node-core/src/indexer/fetch.service.ts
```diff
@@ -235,6 +236,14 @@ export abstract class BaseFetchService<
         continue;
       }
 
+      if (cachedDictBlocks.length > 0) {
```
stwiname, line 239: I think this is starting to do what this issue is wanting to achieve https://github.com/subquery/subql/issues/1960.

This is effectively like adding another queue which is unnecessary. We should look at changing the behaviour of the block number queue to compensate for this. I also think we should set a limit on the number of dictionary results (100 per entity is the default so lets stick with that)

packages/node-core/src/indexer/dictionary.service.ts
stwiname: This is still gone.

packages/node-core/src/indexer/dictionary.service.ts
```diff
@@ -250,45 +324,66 @@ export class DictionaryService {
...
+            this.afterCursor = resp.data[entity].pageInfo.endCursor;
+          }
+        }
+      } else {
+        for (const entity of Object.keys(resp.data)) {
+          if (entity !== '_metadata' && resp.data[entity].nodes.length > 0) {
+            for (const node of resp.data[entity].nodes) {
+              blockHeightSet.add(Number(node.blockHeight));
+              entityEndBlock[entity] = Number(node.blockHeight);
+            }
```
stwiname, line 362: Maybe this was better before, this adds a lot of duplicate code

packages/node-core/src/indexer/dictionary.service.ts
```diff
@@ -250,45 +324,66 @@ export class DictionaryService {
...
+        queryEndBlock,
+        batchSize,
+        conditions,
+        this.afterCursor
+      );
       const resp = await timeout(
         this.client.query({
           query: gql(query),
-          variables,
+          variables: {...variables},
```
stwiname, line 341: Whats this change for?

packages/node-core/src/indexer/dictionary.service.ts
```diff
@@ -174,6 +247,7 @@ export class DictionaryService {
     let link: ApolloLink;
 
     if (this.nodeConfig.dictionaryResolver) {
+      this.supportsCursor = false;
```
stwiname, line 250: This wont work, if the dictionary resolver flag is set it doesn't mean that a network dictionary is used. The network might not offer a dictionary for that chain so it will use the fallbackServiceUrl which would be a traditional dictionary.

DeveloperInProgress: The fallback happens inside apollo-link, do you think we can alter the query at this level?

image

stwiname: Yes, I don't think its really possible for us to check

ianhe8x:
@stwiname For dictionary, I think cursor won't make query more efficient since we update block range condition for each query which is basically what cursor do.
Plus there will be extra efforts needed to make dict-tyk support cursor. I think we just pause this PR.

BTW, cursor should work for a postgraphile cluster. (nothing need to be shared)
E.g a cursor that looks like `WyJibG9ja19oZWlnaHRfYXNjIixbNDEyMDQ3MjEsIjQxMjA0NzIxLTI2MyJdXQ==`, if you decode it with base64 it is `["block_height_asc",[41204721,"41204721-263"]]`

stwiname:
Im closing this for now as there doesn't seem to be much gain and there are problems with load balanced endpoints hitting multiple instances

Issue #2020, opened 2023-09-13
chain is undefined, Testing Framework


bz888:
### Prerequisites

* [x] Are you running the latest version(s)?
* [x] Have you searched the [documentation for your issue](https://academy.subquery.network/)?
* [x] Did you search for an existing issue or pull request?

### Description
When running the testing framework on main SDK (`subql-node`) 
Getting Error:
```
2023-09-13T11:18:54.165Z <Testing> ERROR Testing failed 
TypeError: Cannot destructure property 'chain' of 'this.apiService.networkMeta' as it is undefined.
```
### Steps to Reproduce

- Create any substrate family subql project
- Add test, can be a failing test

- When running `subql-node` test

The test should fail, but it is failing to even initialise the test 

https://github.com/subquery/testing-projects/tree/main/testing-framework

### Any other information

Is there any other information you would like to add?

PR #2028, opened 2023-09-14
Fix dependency injection for testing framework and not initing api service

stwiname:
# Description
Fixes testing framework not running. Removes duplicate ApiService and ProjectService. 

Fixes https://github.com/subquery/subql/issues/2020

## Type of change

## Checklist

- [x] I have tested locally

### DIFF
--- a/packages/node-core/src/api.service.ts
+++ b/packages/node-core/src/api.service.ts
@@ -1,6 +1,7 @@
 // Copyright 2020-2023 SubQuery Pte Ltd authors & contributors
 // SPDX-License-Identifier: GPL-3.0
 
+import assert from 'assert';
 import {EventEmitter2} from '@nestjs/event-emitter';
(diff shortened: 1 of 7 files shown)


jiqiang90 (approved):

DeveloperInProgress (commented):
There are two instances of ApiService being created, one of them calls `init()` and one of them does not. `IApi` references to the instance that did not call `init()`. maybe we should export ApiService from FetchModule, so it doesn't have to be provided again in the TestingModule context.

stwiname:
> There are two instances of ApiService being created, one of them calls `init()` and one of them does not. `IApi` references to the instance that did not call `init()`. maybe we should export ApiService from FetchModule, so it doesn't have to be provided again in the TestingModule context.

I don't think that FetchModule is being used with testing though. And I don't think it should either

DeveloperInProgress (approved):

PR #2029, opened 2023-09-14
distribute worker load based on number of blocks loaded

DeveloperInProgress:
# Description
Current method of distribute fetch loads to workers uses the value of heap memory left in a worker as a metric for distribution. This leads to overloading of first worker before fetching begins, or when it is throttled because of a ratelimited endpoint

image

Instead we should use the number of blocks loaded to a worker to distribute further loads. This metric also serves as an approximate comparison for memory used by workers since majority of the heap memory will be occupied by fetched blocks. 

## Type of change

## Checklist

- [ ] I have tested locally

### DIFF
--- a/packages/node-core/src/indexer/blockDispatcher/worker-block-dispatcher.ts
+++ b/packages/node-core/src/indexer/blockDispatcher/worker-block-dispatcher.ts
@@ -42,6 +42,7 @@ export abstract class WorkerBlockDispatcher<DS, W extends Worker>
   protected workers: W[] = [];
   private numWorkers: number;
   private isShutdown = false;
+  private currentWorkerIndex = 0;
 
   protected abstract fetchBlock(worker: W, height: number): Promise<void>;
 
@@ -190,22 +191,15 @@ export abstract class WorkerBlockDispatcher<DS, W extends Worker>
     });
   }
 
+  //assuming all workers have same constraints, distribute load by
+  //round-robin based on number of blocks already loaded to ensure even distribution
   private async getNextWorkerIndex(): Promise<number> {
-    return Promise.all(this.workers.map((worker) => worker.getBlocksLoaded())).then((blocksLoaded) => {
-      const minBlocks = Math.min(...blocksLoaded);
-
-      // Collect all indices with minimum blocks loaded
-      const minIndices = [];
-      for (let i = 0; i < blocksLoaded.length; i++) {
-        if (blocksLoaded[i] === minBlocks) {
-          minIndices.push(i);
-        }
-      }
-
-      // Return a random index among the minIndices
-      const randomIndex = Math.floor(Math.random() * minIndices.length);
-      return minIndices[randomIndex];
-    });
+    this.currentWorkerIndex = (this.currentWorkerIndex + 1) % this.workers.length;
+    const memLeft = await this.workers[this.currentWorkerIndex].getMemoryLeft();
+    if (memLeft < this.minimumHeapLimit) {
+      return this.getNextWorkerIndex();
+    }
+    return this.currentWorkerIndex;
   }
 
   private async maxBatchSize(workerIdx: number): Promise<number> {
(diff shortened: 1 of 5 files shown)


stwiname (commented):
Code looks good but I think this needs tests

packages/node-core/src/indexer/blockDispatcher/worker-block-dispatcher.ts
```diff
@@ -190,9 +192,12 @@ export abstract class WorkerBlockDispatcher<DS, W extends Worker>
   }
 
   private async getNextWorkerIndex(): Promise<number> {
-    return Promise.all(this.workers.map((worker) => worker.getMemoryLeft())).then((memoryLeftValues) => {
-      return memoryLeftValues.indexOf(Math.max(...memoryLeftValues));
-    });
+    this.currentWorkerIndex = (this.currentWorkerIndex + 1) % this.workers.length;
```
ianhe8x, line 195: This looks very tricky, you changed a class variable in a `get` method, which only suppose to be readonly.

packages/node-core/src/indexer/blockDispatcher/worker-block-dispatcher.ts
```diff
@@ -190,9 +192,12 @@ export abstract class WorkerBlockDispatcher<DS, W extends Worker>
...
+    const memLeft = await this.workers[this.currentWorkerIndex].getMemoryLeft();
+    if (memLeft < this.minimumHeapLimit) {
+      return this.getNextWorkerIndex();
```
ianhe8x, line 198: Can this in certain circumstance become a infinite loop?

ianhe8x:
I think the memory based adjustment on block fetch distribution might end up a less efficiency strategy.

In the beginning, we've tested two strategy, one is distribute jobs to works evenly. But it turned out slower comparing with the other strategy that we always distribute the batch to same worker, plus, when a worker is indexing, it doesn't fetch blocks, so all cpu of a worker can be maximized to spent on indexing rather than splited to block fetching.

Now, this memory based adjustment breaks a batch again.

From my pov, if we are worried about memory usage in block fetching, we can do some control in worker's block fetch logic, pause fetching the blocks in the queue when we find memory is reaching limit.

DeveloperInProgress:
> In the beginning, we've tested two strategy, one is distribute jobs to workers evenly. But it turned out slower comparing with the other strategy that we always distribute a full batch to same worker, plus, when a worker is indexing, it doesn't get more blocks (unless it is the only worker), so all cpu of a worker can be maximized to spent on indexing rather than splited to block fetching.
> 
> Now, this memory based adjustment breaks a batch again.

smart batch sizes would help in the scenarios where a system has limited memory and 

> 
> From my pov, if we are worried about memory usage in block fetching, we can do some control in worker's block fetch logic, pause fetching the blocks in the queue when we find memory is reaching limit.




> @DeveloperInProgress I think the memory based adjustment on block fetch distribution might end up a less efficiency strategy.
> 
> In the beginning, we've tested two strategy, one is distribute jobs to workers evenly. But it turned out slower comparing with the other strategy that we always distribute a full batch to same worker, plus, when a worker is indexing, it doesn't get more blocks (unless it is the only worker), so all cpu of a worker can be maximized to spent on indexing rather than splited to block fetching.
> 
> Now, this memory based adjustment breaks a batch again.
> 
> From my pov, if we are worried about memory usage in block fetching, we can do some control in worker's block fetch logic, pause fetching the blocks in the queue when we find memory is reaching limit.
> 
> `packages/node/src/indexer/blockDispatcher/worker-block-dispatcher.service.ts:L151`

yes it's good to let the GC handle memory management, and we also have a logic to wait for memory if it goes below a limit
https://github.com/subquery/subql/blob/d8383bedca09806bdfb23f94091999b674b74fae/packages/node-core/src/utils/batch-size.ts#L42

On the other hand, should we consider the downsides of loosing smart batch sizing? for example, we would still be on the lookout for OOM error. Sticking to original batch size and waiting for memory to be safe from OOM would increase the idle time.

ianhe8x:
@DeveloperInProgress The safe free memory checks are applied in different layers currently. I think there's a lot of space for simplification and improvement.
Improvements can be in separated PR. For this one, I think my only left concern is the infinite loop it may introduce.

ianhe8x (approved):

PR #2044, opened 2023-09-25
Rename skipBlock to skipTransaction, make NodeConfig expandable

stwiname:
# Description
Renames the `skipBlock` feature to `skipTransactions` to better represent the feature. 

It also makes some minor changes to NodeConfig to do the following:
* Allow extending so the different SDKs can have their own options
* Support any type of node runner options so we can have SDK specific options.

## Checklist

- [x] I have tested locally

### DIFF
--- a/packages/node-core/src/utils/configure.ts
+++ b/packages/node-core/src/utils/configure.ts
@@ -12,7 +12,7 @@ export interface ArgvOverrideOptions {
   unsafe?: boolean;
   disableHistorical?: boolean;
   unfinalizedBlocks?: boolean;
-  skipBlock?: boolean;
+  skipTransaction?: boolean;
 }
 
 export function defaultSubqueryName(config: Partial<IConfig>): MinConfig {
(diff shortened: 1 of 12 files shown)


packages/node-core/src/utils/configure.ts
```diff
@@ -30,28 +30,26 @@ export function defaultSubqueryName(config: Partial<IConfig>): MinConfig {
...
-  // we override them if they are not provided in args/flag
-  if (argvs.disableHistorical === undefined && options.historical !== undefined) {
-    // THIS IS OPPOSITE
-    argvs.disableHistorical = !options.historical;
-  }
-  applyArgs(argvs, options, 'unsafe');
-  applyArgs(argvs, options, 'unfinalizedBlocks');
-  applyArgs(argvs, options, 'skipBlock');
+  // Do this with any options as other SDKs might want to implement custom options
+  (Object.entries(options) as [keyof Omit<ArgvOverrideOptions, 'disableHistorical'> | 'historical', any][]).map(
```
stwiname, line 40: I'm unsure if this is the right move to support any option from the manifest. It should still be caught by the class validation. But if a user can override anything it could lead to trouble

jiqiang90 (approved):

PR #2046, opened 2023-09-26
Fix project upgrades in workers

stwiname:
# Description
Updates project upgrade service to work within workers when the store is not available

Also adds a minor change to move main thread checks from worker fetch module to constructors

Fixes # (issue)

## Type of change

## Checklist

- [x] I have tested locally

### DIFF
--- a/packages/node-core/src/indexer/project.service.ts
+++ b/packages/node-core/src/indexer/project.service.ts
@@ -78,7 +78,7 @@ export abstract class BaseProjectService<API extends IApi, DS extends BaseDataSo
     return getExistingProjectSchema(this.nodeConfig, this.sequelize);
   }
 
-  async init(): Promise<void> {
+  async init(startHeight?: number): Promise<void> {
     for await (const [, project] of this.projectUpgradeService.projects) {
       await project.applyCronTimestamps(this.getBlockTimestamp.bind(this));
     }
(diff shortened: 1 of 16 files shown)


packages/node-core/src/indexer/project.service.ts
```diff
@@ -115,6 +115,12 @@ export abstract class BaseProjectService<API extends IApi, DS extends BaseDataSo
 
       // Flush any pending operations to set up DB
       await this.storeService.storeCache.flushCache(true, true);
+    } else {
+      assert(startHeight, 'ProjectService must be initalized with a start height in workers');
```
jiqiang90, line 119: image
I was little confused until I found the usage.

packages/node/src/indexer/indexer.manager.spec.ts
```diff
@@ -227,30 +227,28 @@ describe('IndexerManager', () => {
...
+    //     nodeConfig,
+    //     new ConnectionPoolStateManager(),
+    //   ),
+    //   nodeConfig,
+    // );
+    // await expect(indexerManager.start()).resolves.toBe(undefined);
+    // expect(Object.keys((indexerManager as any).vms).length).toBe(1);
   });
 
   it.skip('should be able to start the manager (v0.2.0)', async () => {
```
jiqiang90, line 242: I think we can remove these tests

jiqiang90 (approved):

packages/node/src/indexer/indexer.manager.spec.ts
stwiname: It would be good to bring back some tests for the indexer manager at some point so i left them commented

PR #2050, opened 2023-09-28
[release] 20230928

stwiname:
Release for CLI can come after we have updated the other SDKs and the common-* dependencies in the cli

### DIFF
--- a/packages/types-core/package.json
+++ b/packages/types-core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@subql/types-core",
-  "version": "0.0.0-2",
+  "version": "0.0.0",
   "description": "",
   "homepage": "https://github.com/subquery/subql",
   "repository": "github:subquery/subql",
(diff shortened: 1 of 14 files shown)


packages/types-core/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@subql/types-core",
-  "version": "0.0.0-2",
+  "version": "0.0.0",
```
jiqiang90, line 3: I wonder do we need a changelog for init release?

jiqiang90:
I think we missed changelog for ts manifest ?

jiqiang90 (changes requested):

packages/types-core/package.json
stwiname: I don't think so, there needs to be a reference point from what to base changes off.

stwiname:
> I think we missed changelog for ts manifest ?
> 
> Will come after Cli release?

Ah was there no changelogs in the TS Manifest pr? Could you add changes to this PR?

jiqiang90 (approved):

PR #2061, opened 2023-10-03
update ProjectManifestV1 interface

bz888:
ProjectManifest V1 interface should have a generic network typing


packages/types-core/src/project/versioned/v1_0_0/types.ts
```diff
@@ -136,6 +137,7 @@ export interface ParentProject {
 }
 
 export interface ProjectManifestV1_0_0<
+  C extends IProjectNetworkConfig,
```
stwiname, line 140: Can we make this the last generic and default to `IProjectNetworkConfig`

jiqiang90 (approved):

PR #2064, opened 2023-10-03
Endblock feature

### LINKED ISSUE #2008: endblock on DataSource
Allow an optional `endBlock` on the datasource
Ensure compatibility with dictionary

Filters in that datasource will only run from `startBlock` to `endBlock`

DeveloperInProgress:
# Description
Allow an optional endBlock on the datasource. Handlers in that datasource will only run from startBlock to endBlock

Fixes #2008 

## Type of change

## Checklist

- [ ] I have tested locally

### DIFF
--- a/packages/node-core/src/indexer/blockDispatcher/base-block-dispatcher.ts
+++ b/packages/node-core/src/indexer/blockDispatcher/base-block-dispatcher.ts
@@ -208,7 +208,7 @@ export abstract class BaseBlockDispatcher<Q extends IQueue, DS> implements IBloc
       await this.storeCacheService.flushCache(false, true);
     }
 
-    if (!(await this.projectService.hasDataSourceWithEndBlockGreaterThan(height))) {
+    if (!(await this.projectService.hasDataSourcesAfterHeight(height))) {
       logger.info(`All data sources have been processed up to block number ${height}. Exiting gracefully...`);
       await this.storeCacheService.flushCache(false, true);
       process.exit(0);
--- a/packages/node-core/src/indexer/blockDispatcher/base-block-dispatcher.ts
+++ b/packages/node-core/src/indexer/blockDispatcher/base-block-dispatcher.ts
@@ -209,7 +209,7 @@ export abstract class BaseBlockDispatcher<Q extends IQueue, DS> implements IBloc
       await this.storeCacheService.flushCache(false, true);
     }
 
-    if (!(await this.projectService.hasDataSourcesAfterHeight(height))) {
+    if (!this.projectService.hasDataSourcesAfterHeight(height)) {
       logger.info(`All data sources have been processed up to block number ${height}. Exiting gracefully...`);
       await this.storeCacheService.flushCache(false, true);
       process.exit(0);
--- a/packages/node-core/src/indexer/blockDispatcher/base-block-dispatcher.ts
+++ b/packages/node-core/src/indexer/blockDispatcher/base-block-dispatcher.ts
@@ -209,7 +209,7 @@ export abstract class BaseBlockDispatcher<Q extends IQueue, DS> implements IBloc
     }
 
     if (!(await this.projectService.hasDataSourceWithEndBlockGreaterThan(height))) {
-      logger.info(`No data sources found with endBlock greater than the block number ${height}`);
+      logger.info(`All data sources have been processed up to block number ${height}. Exiting gracefully...`);
       await this.storeCacheService.flushCache(false, true);
       process.exit(0);
     }
--- a/packages/node-core/src/indexer/blockDispatcher/base-block-dispatcher.ts
+++ b/packages/node-core/src/indexer/blockDispatcher/base-block-dispatcher.ts
@@ -207,6 +207,12 @@ export abstract class BaseBlockDispatcher<Q extends IQueue, DS> implements IBloc
       // Flush all data from cache and wait
       await this.storeCacheService.flushCache(false, true);
     }
+
+    if (!(await this.projectService.hasDataSourceWithEndBlockGreaterThan(height))) {
+      logger.info(`No data sources found with endBlock greater than the block number ${height}`);
+      await this.storeCacheService.flushCache(false, true);
+      process.exit(0);
+    }
   }
 
   // First creation of POI
--- a/packages/node-core/src/indexer/dictionary.service.ts
+++ b/packages/node-core/src/indexer/dictionary.service.ts
@@ -156,10 +156,7 @@ function buildDictQueryFragment(
 export class DictionaryService {
   private _client?: ApolloClient<NormalizedCacheObject>;
 
-  queriesMap?: BlockHeightMap<{
-    query: DictionaryQueryEntry[];
-    maxDsEndBlock: number; //maximum value of endBlock obtained from datasources for current project
-  }>;
+  queriesMap?: BlockHeightMap<DictionaryQueryEntry[]>;
   private useDistinct = true;
   private useStartHeight = true;
   protected _startHeight?: number;
@@ -345,11 +342,19 @@ export class DictionaryService {
     dataSources: BlockHeightMap<DS[]>,
     buildDictionaryQueryEntries: (dataSources: DS[]) => DictionaryQueryEntry[]
   ): void {
-    this.queriesMap = dataSources.map((dataSources) => {
-      const query = buildDictionaryQueryEntries(dataSources);
-      const maxDsEndBlock = maxEndBlockHeight(dataSources);
-      return {query, maxDsEndBlock};
+    this.queriesMap = dataSources.map(buildDictionaryQueryEntries);
+    const newQueriesMap = this.queriesMap?.getAll() || new Map();
+
+    dataSources.getAll().forEach((ds, height) => {
+      const endBlock = maxEndBlockHeight(ds);
+      const queryDetails = this.queriesMap?.getDetails(height);
+
+      if (!queryDetails?.endHeight || queryDetails.endHeight > endBlock) {
+        newQueriesMap.set(endBlock + 1, []);
+      }
     });
+
+    this.queriesMap = new BlockHeightMap(newQueriesMap);
   }
 
   async scopedDictionaryEntries(
@@ -358,13 +363,7 @@ export class DictionaryService {
     scaledBatchSize: number
   ): Promise<(Dictionary & {queryEndBlock: number}) | undefined> {
     const queryDetails = this.queriesMap?.getDetails(startBlockHeight);
-    const queryEntry: DictionaryQueryEntry[] = queryDetails?.value.query ?? [];
-
-    // Update end block if query changes
-    queryEndBlock =
-      queryDetails?.value.maxDsEndBlock && queryDetails?.value?.maxDsEndBlock < queryEndBlock
-        ? queryDetails.value.maxDsEndBlock
-        : queryEndBlock;
+    const queryEntry: DictionaryQueryEntry[] = queryDetails?.value ?? [];
 
     queryEndBlock =
       queryDetails?.endHeight && queryDetails?.endHeight < queryEndBlock ? queryDetails.endHeight : queryEndBlock;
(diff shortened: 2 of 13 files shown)


packages/node-core/src/indexer/fetch.service.ts
```diff
@@ -243,10 +243,15 @@ export abstract class BaseFetchService<
...
+        /* queryEndBlock needs to be limited by the latest height or the maximum value of endBlock in datasources.
          * Dictionaries could be in the future depending on if they index unfinalized blocks or the node is using an RPC endpoint that is behind.
          */
-        const queryEndBlock = Math.min(startBlockHeight + DICTIONARY_MAX_QUERY_SIZE, this.latestFinalizedHeight);
+        const maxEndBlock = await this.projectService.maxEndBlockHeight();
+        const queryEndBlock = Math.min(
+          startBlockHeight + DICTIONARY_MAX_QUERY_SIZE,
+          this.latestFinalizedHeight,
+          maxEndBlock
+        );
```
stwiname, line 254: I don't think this is right here. The `queriesMap` should be updated to take into consideration the changes

packages/node-core/src/indexer/dictionary.service.ts
```diff
@@ -155,7 +156,10 @@ function buildDictQueryFragment(
 export class DictionaryService {
   private _client?: ApolloClient<NormalizedCacheObject>;
 
-  queriesMap?: BlockHeightMap<DictionaryQueryEntry[]>;
+  queriesMap?: BlockHeightMap<{
+    query: DictionaryQueryEntry[];
+    maxDsEndBlock: number; //maximum value of endBlock obtained from datasources for current project
```
stwiname, line 161: Block height map should already work out the end block https://github.com/subquery/subql/blob/main/packages/node-core/src/utils/blockHeightMap.ts#L4

We just need to change the entries in the map based on the end blocks

packages/node-core/src/indexer/fetch.service.spec.ts
```diff
@@ -152,7 +152,9 @@ describe('Fetch Service', () => {
   const enableDictionary = () => {
     // Mock the remainder of dictionary service so it works
     (dictionaryService as any).useDictionary = true;
-    dictionaryService.queriesMap = new BlockHeightMap(new Map([[1, [{entity: 'mock', conditions: []}]]]));
+    dictionaryService.queriesMap = new BlockHeightMap(
+      new Map([[1, {query: [{entity: 'mock', conditions: []}], maxDsEndBlock: 1000}]])
```
stwiname, line 156: The correct data for this should be

```ts
new Map([
  [1, [{entity: 'mock', conditions: []}]],
  [1001, []]
])
```

packages/node-core/src/indexer/dictionary.service.ts
```diff
@@ -345,11 +342,19 @@ export class DictionaryService {
...
+    dataSources.getAll().forEach((ds, height) => {
+      const endBlock = maxEndBlockHeight(ds);
+      const queryDetails = this.queriesMap?.getDetails(height);
+
+      if (!queryDetails?.endHeight || queryDetails.endHeight > endBlock) {
+        newQueriesMap.set(endBlock + 1, []);
+      }
     });
+
+    this.queriesMap = new BlockHeightMap(newQueriesMap);
```
stwiname, line 357: I think it would make sense to build to datasources blockheight map correctly, then we know it matches up everywhere

packages/node-core/src/indexer/project.service.ts
```diff
@@ -283,7 +291,16 @@ export abstract class BaseProjectService<API extends IApi, DS extends BaseDataSo
         });
     }
 
-    return new BlockHeightMap(dsMap);
+    const newDsMap = cloneDeep(dsMap);
```
stwiname, line 294: Im not sure why this needs to be cloned.

packages/node-core/src/indexer/project.service.ts
```diff
@@ -283,7 +291,16 @@ export abstract class BaseProjectService<API extends IApi, DS extends BaseDataSo
...
+    [...dsMap.entries()].forEach(([height, ds], i, arr) => {
+      const nextStart = arr[i + 1]?.[0];
+      const endBlock = maxEndBlockHeight(ds);
+      if (endBlock < Number.MAX_SAFE_INTEGER && (nextStart === undefined || nextStart > endBlock + 1)) {
+        newDsMap.set(endBlock + 1, []);
+      }
```
stwiname, line 300: This only seems to consider at the end. Can you please write tests for this.

I don't think this would work:

```ts
const dataSources =  [{ startBlock: 1}, { startBlock: 10, endBlock 20, }];

// before this pr
const map = new Map([
  [1, [/* ds 1*/]],
  [10, [/* ds 1 and 2 */],
])


// what it should be
const map = new Map([
  [1, [/* ds 1*/]],
  [10, [/* ds 1 and 2 */],
  [21, [/* ds 1*/]]
]);

// What it looks like its doing
const map = new Map([
  [1, [/* ds 1*/]],
  [10, [/* ds 1 and 2 */],
  [21, [/* empty */]]
]);
```

It also needs to accommodate more datasources that could start and end anywhere inbetween

packages/node-core/src/indexer/project.service.ts
```diff
@@ -260,31 +260,65 @@ export abstract class BaseProjectService<API extends IApi, DS extends BaseDataSo
...
+    // sort events by block in ascending order, start events come before end events
+    events.sort((a, b) => a.block - b.block || Number(b.start) - Number(a.start));
+
+    for (const event of events) {
+      if (event.start) {
+        activeDataSources.add(event.ds);
+      } else {
+        activeDataSources.delete(event.ds);
+      }
+      dsMap.set(event.block, Array.from(activeDataSources));
```
stwiname, line 321: I think this needs to be done in a single loop. An issue I can see is datasources being mixed between project versions. It would also be much simpler

packages/node-core/src/indexer/project.service.spec.ts
```diff
@@ -0,0 +1,137 @@
...
+      null as unknown as any,
+      null as unknown as any,
+      {unsafe: false} as unknown as NodeConfig,
+      {getDynamicDatasources: jest.fn()} as unknown as DynamicDsService<any>,
+      null as unknown as any,
+      null as unknown as any
+    );
+  });
+
+  test('hasDataSourcesAfterHeight', async () => {
```
stwiname, line 42: Please follow this pattern

```suggestion
  it('has data sources after height', async () => {
```

packages/node-core/src/indexer/project.service.spec.ts
```diff
@@ -0,0 +1,137 @@
...
+  async getBlockTimestamp(height: number): Promise<Date> {
+    return Promise.resolve(new Date());
+  }
+
+  onProjectChange(project: any): void {
+    return;
+  }
+}
+
+describe('BaseProjectService', () => {
```
stwiname, line 22: Can you please add tests over project versions to ensure there is only datasources from one version at any one time

packages/node-core/src/indexer/project.service.spec.ts
```diff
@@ -0,0 +1,167 @@
...
+          ],
+          [
+            101,
+            [
+              {startBlock: 1, endBlock: 300},
+              {startBlock: 50, endBlock: 200},
+            ],
+          ],
+          [201, [{startBlock: 1, endBlock: 300}]],
+          [301, []],
```
stwiname, line 133: If this can happen then the indexer will exit because of https://github.com/subquery/subql/blob/5c3142bb41f9196e4fbe433ede8f66021dfbf3cc/packages/node-core/src/indexer/blockDispatcher/base-block-dispatcher.ts#L212-L215

We either need to support this correctly in the fetch service when enqueuing blocks or throw an error if this happens

packages/node-core/src/indexer/project.service.spec.ts
```diff
@@ -0,0 +1,167 @@
...
+    });
+
+    it('should contain datasources from current project only', () => {
+      (service as any).dynamicDsService.dynamicDatasources = [];
+      (service as any).projectUpgradeService = {
+        projects: [
+          [
+            1,
+            {
+              dataSources: [{startBlock: 1}],
```
stwiname, line 146: This is a pre-existing bug buy if this change is made then it will produce a different result but the expected result is as it is now
```suggestion
              dataSources: [{startBlock: 1}, {startBlock: 200}],
```

packages/node-core/src/indexer/project.service.spec.ts
DeveloperInProgress: If no datasources are available to index a specific blockheight we throw this error:

https://github.com/subquery/subql/blob/5c3142bb41f9196e4fbe433ede8f66021dfbf3cc/packages/node-core/src/indexer/indexer.manager.ts#L145C1-L159C4

should we change this behaviour to somehow skip blocks that has no datasources?

packages/node-core/src/indexer/project.service.ts
```diff
@@ -309,12 +309,18 @@ export abstract class BaseProjectService<API extends IApi, DS extends BaseDataSo
...
-          event.start ? activeDataSources.add(event.ds) : activeDataSources.delete(event.ds);
-          dsMap.set(event.block, Array.from(activeDataSources));
-        });
+      const sortedEvents = events.sort((a, b) => a.block - b.block || Number(b.start) - Number(a.start));
+
+      // remove all dsMap entries after this height because it will be replaced by current project datasources
+      const minStartHeight = sortedEvents[0].block;
+      [...dsMap.entries()].forEach(([height, ds]) => {
+        if (height >= minStartHeight) dsMap.delete(height);
+      });
```
stwiname, line 318: Rather than deleting, why not look forward to the next project and filter based on that?

packages/node-core/src/indexer/project.service.spec.ts
stwiname: I think that happens too late on having to index to that height to see the error.

I think we can skip blocks, its should be relatively easy to do

packages/node-core/src/indexer/blockDispatcher/base-block-dispatcher.ts
```diff
@@ -208,6 +208,12 @@ export abstract class BaseBlockDispatcher<Q extends IQueue, DS> implements IBloc
       // Flush all data from cache and wait
       await this.storeCacheService.flushCache(false, true);
     }
+
+    if (!(await this.projectService.hasDataSourcesAfterHeight(height))) {
```
stwiname, line 212: I'm not sure this is right. What happens if we run into bypassblocks at the next height?

packages/node-core/src/indexer/fetch.service.ts
```diff
@@ -99,9 +99,27 @@ export abstract class BaseFetchService<
...
+    for (let i = 0; i < sortedHeights.length - 1; i++) {
+      const currentHeight = sortedHeights[i];
+      const nextHeight = sortedHeights[i + 1];
+
+      const currentDS = datasources.get(currentHeight);
+      // If the value for the current height is an empty array, then it's a gap
+      if (currentDS && currentDS.length === 0) {
+        this.bypassBlocks.push(...range(currentHeight, nextHeight));
+      }
+    }
```
stwiname, line 121: Can you please move this to a function

packages/node-core/src/indexer/fetch.service.ts
```diff
@@ -99,9 +99,27 @@ export abstract class BaseFetchService<
...
+    const sortedHeights = Array.from(datasources.keys()).sort((a, b) => a - b);
+
+    for (let i = 0; i < sortedHeights.length - 1; i++) {
+      const currentHeight = sortedHeights[i];
+      const nextHeight = sortedHeights[i + 1];
+
+      const currentDS = datasources.get(currentHeight);
+      // If the value for the current height is an empty array, then it's a gap
+      if (currentDS && currentDS.length === 0) {
+        this.bypassBlocks.push(...range(currentHeight, nextHeight));
```
stwiname, line 119: Does this need to be sorted?

DeveloperInProgress: I suppose it's already sorted, removing the sort

packages/node-core/src/indexer/blockDispatcher/base-block-dispatcher.ts
DeveloperInProgress: maybe we can do this check with datasources map to make sure we are not exiting prematurely

stwiname (approved):

jamesbayly:
Docs for review https://github.com/subquery/documentation/pull/443

PR #2084, opened 2023-10-10
check for ts manifest before building it

DeveloperInProgress:
# Description
Currently, codegen assumes that the default ts manifest exists and tries to build it. We should check if a ts manifest exists before building it

Fixes # (issue)

## Type of change

## Checklist

- [ ] I have tested locally


stwiname (approved):

jiqiang90 (commented):
This is rollback https://github.com/subquery/subql/pull/2080

The idea is when run `build` command each time ,should always rebuild ts manifest.

let's imagine if user made change to manifest file after the codegen, like change a filter name etc. Before we use yaml file it doesn't require a build step for manifest, but now it should have a rebuild.

jiqiang90 (approved):

PR #2110, opened 2023-10-18
Support built in cache feature

### LINKED ISSUE #2022: Support built in cache feature
### Description
I sometimes use in-memory cache to save rpc queries. The sandbox we use allow it, since we reuse the instance.
It works like a charm until some error report and I realized this will break when multiple workers are used.

As a solution for this, I switch to store the value I want to cache in db. But I think there's better ways since I don't really want it persisted in db.

We've done a great job for `store` and make it work with multi workers by moving the store to main thread and use inter-thread communication. I think we can do same for cache.
### Details
* introduce a global var `cache`
* cache lives in main thread
* cache supports two functions
  * set(key: string, value: string): Promise<void>
  * get(key: string): Promise<string|undefined>

DeveloperInProgress:
# Description
Implement in-memory cache and inject it to sandbox. workers will use the cache service from host to allow using same instance of cache in the sandbox.

Fixes #2022

## Type of change

## Checklist

- [ ] I have tested locally


packages/node-core/src/indexer/worker/worker.store.service.ts
```diff
@@ -19,7 +19,7 @@ function unwrapProxy<T = any>(input: T): T {
 }
 
 /* Unwraps any arguments to a function that are proxy objects */
-function unwrapProxyArgs<T extends Array<any>, R>(fn: (...args: T) => R): (...args: T) => R {
+export function unwrapProxyArgs<T extends Array<any>, R>(fn: (...args: T) => R): (...args: T) => R {
```
stwiname, line 22: If we're using this in multiple places it should be moved to a common place.

packages/types-core/src/interfaces.ts
```diff
@@ -2,3 +2,8 @@
 // SPDX-License-Identifier: GPL-3.0
 
 export type DynamicDatasourceCreator = (name: string, args: Record<string, unknown>) => Promise<void>;
+
+export interface Cache {
+  set<D = string>(key: string, value: D): Promise<void>;
+  get<D = string>(key: string): Promise<D | undefined>;
```
stwiname, line 8: Whats the reason for having default generics here?

packages/node/src/indexer/sandbox.service.ts
```diff
@@ -31,11 +34,16 @@ export class SandboxService<Api> {
       ? this.storeService.getStore()
       : hostStoreToStore((global as any).host); // Provided in worker.ts
 
+    const cache: Cache = isMainThread
+      ? this.cacheService.getCache()
+      : hostCacheToCache((global as any).host);
```
stwiname, line 39: Can we use the approach used by unfinalized blocks service instead please. We should inject the worker version in worker-fetch.module.ts

Ideally we would do the same with the store serverice

packages/node-core/src/indexer/inMemoryCache.service.ts
```diff
@@ -0,0 +1,25 @@
+// Copyright 2020-2023 SubQuery Pte Ltd authors & contributors
+// SPDX-License-Identifier: GPL-3.0
+
+import {Injectable} from '@nestjs/common';
+import {Cache} from '@subql/types-core';
+
+@Injectable()
+export class InMemoryCacheService {
```
stwiname, line 8: Can you please add a comment here to say the purpose of this cache

packages/node-core/src/indexer/inMemoryCache.service.ts
```diff
@@ -0,0 +1,30 @@
...
+  private cache: Record<string, any> = {};
+
+  getCache(): Cache {
+    return {
+      //eslint-disable-next-line @typescript-eslint/require-await
+      get: async <D = string>(key: string): Promise<D | undefined> => {
+        if (this.cache[key]) {
+          return this.cache[key] as D;
+        }
```
stwiname, line 22:
```suggestion
        return this.cache[key] as D;
```

stwiname (approved):

PR #2124, opened 2023-10-25
build docker image from source

### LINKED ISSUE #2114: Build docker image from sources
Currently docker images are built from npm packages, but there are downsides to this:
* We have to publish to npm in order to make an image
* There are potential delays with the package being available on npm so that we cant automate making a docker image after a release
* Dependencies can be updated with breaking or incompatible changes causing more issues

We should update the docker image to instead build from sources so we can resolve a lot of these issues
Then we can also automate publishing images after publishing to npm. We should also make the docker image a small as possible by not including any unnecessary files

DeveloperInProgress:
# Description
Build docker image for node from source rather than building it from published package. And add docker publish to release workflow

Fixes #2114

## Type of change

## Checklist

- [ ] I have tested locally


.github/workflows/node-docker.yml
```diff
@@ -79,11 +96,15 @@ jobs:
...
+      - run: yarn
+
+      - name: build
+        run: yarn build
+
+      # Replace workspace:* with actual versions
+      - name: Replace workspace dependencies with actual versions
         run: |
-          sh .github/workflows/scripts/nodeVersion.sh
+          sh .github/workflows/scripts/replace-versions.sh
```
stwiname, line 107: It would be great if we could run this from the dockerfile instead. That way its really easy to build locally and the changes are contained to the image

packages/node/Dockerfile
```diff
@@ -1,14 +1,28 @@
...
+# Build the application
+#RUN yarn build
 
 FROM node:18-alpine
 ENV TZ utc
 
 RUN apk add --no-cache tini git curl
-COPY --from=builder /usr/local/lib/node_modules /usr/local/lib/node_modules
+# Copy the built application from the previous stage
+COPY --from=builder /app .
```
stwiname, line 24: Ideally we can cut this down so we only include files that would be in the NPM package. e.g nothing from /src

stwiname (commented):
It would be good to compare the size of this image to the current way of producing images. Ideally they should be the same or smaller

.github/workflows/node-docker.yml
DeveloperInProgress: That was my go to option at first but, we only copy node package into docker, so we cannot access the versions of the other packages in the workspace.

stwiname: I think we could copy in the package.json files (or other packages) in the first step of the docker image

stwiname: I still think we need to do this in the docker image. From a user perspective its going to make building images much easier

packages/node/Dockerfile
```diff
@@ -1,14 +1,24 @@
...
 RUN apk add --no-cache tini git curl
-COPY --from=builder /usr/local/lib/node_modules /usr/local/lib/node_modules
 
-ENTRYPOINT ["/sbin/tini", "--", "/usr/local/lib/node_modules/@subql/node/bin/run"]
-CMD ["-f","/app"]
+COPY --from=builder /app/package.json ./
+COPY --from=builder /app/yarn.lock ./
+COPY --from=builder /app/node_modules ./node_modules
+COPY --from=builder /app/dist ./dist
+COPY --from=builder /app/bin ./bin
```
stwiname, line 21: Ideally we can use `.npmignore` to determine what to copy. (we might need to make some updates to `.npmignore`)

DeveloperInProgress: modified the dockerfile to pack the node package with yarn before copying to docker, this way we only get files that will be in the npm package

.github/workflows/node-docker.yml
```diff
@@ -79,31 +101,48 @@ jobs:
...
+        run: yarn build
+
       ## node
       - name: Get updated node version
         id: get-node-version
         run: |
           sh .github/workflows/scripts/nodeVersion.sh
 
+      - name: Set execute permissions for the script
+        run: chmod +x .github/workflows/scripts/replace-versions.sh
```
stwiname, line 116: I think if you run this locally then commit then this doesn't need to be done in CI

packages/node/Dockerfile
```diff
@@ -1,14 +1,23 @@
...
 FROM node:18 as builder
-ARG RELEASE_VERSION
-ENTRYPOINT ["subql-node"]
-RUN npm i -g --unsafe-perm @subql/node@${RELEASE_VERSION}
 
+WORKDIR /app
+COPY ./packages/node ./
+
+RUN npm install -g --force yarn@latest
+RUN yarn pack -o /app.tgz
```
stwiname, line 8: Nice!

packages/node/Dockerfile
```diff
@@ -1,14 +1,23 @@
...
+RUN apk add --no-cache tini git curl tar
+
+COPY --from=builder /app/*.tgz /app.tgz
+RUN tar -xzvf /app.tgz --strip 1 && rm /app.tgz
+
+RUN yarn install
 
-ENTRYPOINT ["/sbin/tini", "--", "/usr/local/lib/node_modules/@subql/node/bin/run"]
-CMD ["-f","/app"]
+ENTRYPOINT ["/sbin/tini", "--", "node", "bin/run"]
```
stwiname, line 22: Im not sure why `"node",` is needed. The shebang in bin/run should cover that

stwiname (approved):

jiqiang90 (approved):

PR #2131, opened 2023-10-30
Improve reindex query

jiqiang90:
# Description

Fixes # (issue)

## Type of change

## Checklist

- [ ] I have tested locally


packages/node-core/src/indexer/poi/poi.service.ts
```diff
@@ -159,10 +160,67 @@ export class PoiService implements OnApplicationShutdown {
...
+async function batchDeletePoi(
+  model: PoiRepo,
+  transaction: Transaction,
+  targetBlockHeight: number,
+  batchSize = 10
+): Promise<void> {
+  // const offset = 0;
+  let completed = false;
+  // eslint-disable-next-line no-constant-condition
+  while (!completed) {
```
stwiname, line 194: Why not use `sqlIterator` there? It seems expensive to load before removing.

This maybe useful. Static destroy returns the number of items destroyed https://sequelize.org/api/v6/class/src/model.js~model#static-method-destroy

I see the same is happening with other data.

packages/node-core/src/indexer/poi/poi.service.ts
```diff
@@ -159,10 +160,67 @@ export class PoiService implements OnApplicationShutdown {
...
     this.storeCache.metadata.bulkRemove(['lastCreatedPoiHeight']);
   }
 }
+
+// REMOVE 10,000 record per batch
+async function batchDeletePoi(
+  model: PoiRepo,
+  transaction: Transaction,
+  targetBlockHeight: number,
+  batchSize = 10
```
stwiname, line 189: Comment says 10,000 but this is 10. Does it need to be changed to 10,000?

packages/node-core/src/indexer/store.service.ts
```diff
@@ -907,3 +886,74 @@ group by
...
+        limit: batchSize,
+        attributes: {include: ['_id']},
+        offset, // We need to apply offset, because after update the records, the record could still with in range, avoid endless query here.
+        where: {
+          __block_range: {
+            [Op.contains]: targetBlockHeight,
+          },
+        },
+      });
+      const recordsToDelete = await model.findAll({
```
stwiname, line 914: These could be run in parallel

packages/node-core/src/db/db.module.ts
```diff
@@ -66,6 +66,7 @@ const buildSequelizeOptions = (nodeConfig: NodeConfig, option: DbOption): Sequel
     dialect: 'postgres',
     ssl: nodeConfig.isPostgresSecureConnection,
     dialectOptions: {
+      options: {requestTimeout: nodeConfig.timeout * 1000}, // configure timeout is in seconds, requestTimeout is in milliseconds
```
stwiname, line 69: I don't know if its a good idea to apply a broad timeout for all queries. This could break certain operations such as rewind.

It also has other side effects as the timeout config is applied to a lot of other things. Increasing the timeout for a query might cause other things to timeout that shouldn't timeout and vice versa

packages/node-core/src/indexer/poi/poi.service.ts
jiqiang90: sqlIterator apply sql to whole table, it is useful for sql like `create index xxx` for all records.  

In batchDeletePoi, we only need to remove records with certain block ranges. Compare to sqlIteraor, even we add `where`  conditions to excute the block range in `sql`, it will still iterate the complete table. I think this is the better approach.

stwiname (approved):

PR #2132, opened 2023-10-30
Fix modulo block ahead of finalized

jiqiang90:
# Description

Fixes # (issue)

## Type of change

## Checklist

- [ ] I have tested locally

### DIFF
--- a/packages/node-core/src/indexer/fetch.service.ts
+++ b/packages/node-core/src/indexer/fetch.service.ts
@@ -303,7 +303,7 @@ export abstract class BaseFetchService<
             } else {
               const maxBlockSize = Math.min(batchBlocks.length, this.blockDispatcher.freeSize);
               const enqueueBlocks = batchBlocks.slice(0, maxBlockSize);
-              await this.enqueueBlocks(enqueueBlocks);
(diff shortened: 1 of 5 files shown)


packages/node-core/src/indexer/fetch.service.ts
```diff
@@ -313,14 +318,22 @@ export abstract class BaseFetchService<
...
     }
   }
 
   private async enqueueBlocks(enqueuingBlocks: number[]): Promise<void> {
+    // We check enqueuingBlocks length rather than cleanedBatchBlocks
+    // Because when bypass blocks, cleanedBatchBlocks can be [], but latestBufferHeight could be valid to update metadata
+    // See comments in blockDispatcher.enqueueBlocks method
+    if (!enqueuingBlocks.length) {
+      logger.info(`No blocks to enqueue at the moment.`);
+      await delay(10);
```
stwiname, line 334: I think we should remove this delay here, we should only have one place to throttle and this is a really large delay. If a chain has a block time of a couple of seconds this will add a big delay

stwiname (commented):
Can you please add a test for this

stwiname (approved):

PR #2151, opened 2023-11-09
Fix/near unavailable blocks

bz888:
# Description

Unable to resolve troubles with `instanceof isBlockUnavailableError`
When the `error` is caught on `worker.service` it is behaving as intended, retaining its instance as BlockUnavailableError. However, when that `error` is then thrown once again, for `worker.dispatcher` to catch and process, it no longer retains its instance as `BlockUnavailableError` it is just Error now. 

Fixes #https://github.com/subquery/subql-near/issues/81

## Type of change

## Checklist

- [x] I have tested locally


packages/node-core/src/indexer/worker/worker.service.ts
```diff
@@ -87,7 +88,9 @@ export abstract class BaseWorkerService<
 
       return await this.processFetchedBlock(block, await this.projectService.getDataSources(height));
     } catch (e: any) {
-      logger.error(e, `Failed to index block ${height}: ${e.stack}`);
+      if (!isBlockUnavailableError(e)) {
+        logger.error(e, `Failed to index block ${height}: ${e.stack}`);
+      }
       throw e;
```
stwiname, line 94: Should this still throw?

stwiname (approved):

PR #2179, opened 2023-11-23
Add/schema migration

### LINKED ISSUE #1289: [$3,500 in kSQT] Schema Migrations
### Description

Currently when users wish to change their project schema they need to reindex their project. This can be very slow. In order to improve this we should support schema migrations. These migrations are automated based on changes in project graphql schemas.

The reason for migrations being automated is driven by 2 main reasons:
- It limits security issues by having a narrow scope for migrations. Allowing raw SQL would be a large security issue.
- It limits really heavy migrations such as modifying large amounts of data.

### Requirements

- Migrations should allow the following features:
   - Add/remove colums
   - Add/remove indexes and multi-column indexes
   - Add/remove tables 
   - Changing a field type should drop the column and create a new one
- Migrations should only run on project upgrades
- Migrations should be reversed when rewinding and a project is downgraded
- A way to test migrations buy running through all the projects upgrades on an empty db

bz888:
# Description

Fixes #1289

## Type of change

## Checklist

- [ ] I have tested locally


packages/node-core/src/configure/ProjectUpgrade.service.ts
```diff
@@ -3,23 +3,28 @@
...
+    onProjectUpgrade?: OnProjectUpgradeCallback<P>
+  ) => Promise<number | undefined>;
   /**
    * This should only be called from within a worker thread as they dont have access to the store
    * */
   initWorker(currentHeight: number, onProjectUpgrade?: OnProjectUpgradeCallback<P>): void;
   updateIndexedDeployments: (id: string, blockHeight: number) => Promise<void>;
   readonly currentHeight: number;
-  setCurrentHeight: (newHeight: number) => Promise<void>;
+  setCurrentHeight: (newHeight: number, schema: string) => Promise<void>;
```
stwiname, line 27: It doesn't make sense for schema to be an argument here, it would never change.

packages/node-core/src/configure/ProjectUpgrade.service.ts
```diff
@@ -107,16 +111,18 @@ export class ProjectUpgradeSevice<P extends ISubqueryProject = ISubqueryProject>
   }
 
   async init(
-    metadata: CacheMetadataModel,
+    storeCache: StoreCacheService,
+    schemaMigrationService: SchemaMigrationService,
```
stwiname, line 115: I think now it would make more sense to construct this class in here rather than passing it as an argument. StoreCache should have the relevant information

packages/node-core/src/configure/ProjectUpgrade.service.ts
```diff
@@ -149,24 +155,37 @@ export class ProjectUpgradeSevice<P extends ISubqueryProject = ISubqueryProject>
...
+    const project = this.#currentProject;
     this.#currentProject = newProject;
 
     if (hasChanged) {
       if (isMainThread) {
         try {
-          // Use the project start height here for when resuming indexing at an arbitrary height
+          assert(this.#storeCache, 'CacheMetadataModel is undefined');
+          if (!this.#storeCache?.unfinalizedBlocks) {
+            assert(this.migrationService, 'Migration service is undefined');
```
stwiname, line 174: Can just assert here that init has been called

packages/node-core/src/configure/ProjectUpgrade.service.ts
```diff
@@ -149,24 +155,37 @@ export class ProjectUpgradeSevice<P extends ISubqueryProject = ISubqueryProject>
...
+              this.#storeCache?.indexCountLimit
+            );
+          }
+
           await this.updateIndexedDeployments(newProject.id, startHeight);
         } catch (e: any) {
-          logger.error(e, 'Failed to update deployment metadata');
-          process.exit(1);
+          logger.error(e, 'Failed to update deployment');
+          throw new Error(e);
```
stwiname, line 188:
```suggestion
          throw e;
```

packages/node-core/src/configure/SchemaMigration.service.ts
```diff
@@ -0,0 +1,500 @@
...
+
+  return {graphQLType: currentTypeNode.kind, dataType: name};
+}
+
+// just a wrapper for naming confusion
+function formatColumnName(columnName: string): string {
+  return Utils.underscoredIf(columnName, true);
+}
+
+const sequelizeToPostgresTypeMap = {
```
stwiname, line 81: Use https://github.com/subquery/subql/tree/main/packages/utils/src/types/supported

packages/node-core/src/configure/SchemaMigration.service.ts
```diff
@@ -0,0 +1,500 @@
...
+          removedIndexes,
+        };
+      }
+    }
+  });
+}
+
+function indexesEqual(index1: GraphQLEntityIndex, index2: GraphQLEntityIndex): boolean {
+  return (
+    index1.fields.join(',') === index2.fields.join(',') &&
```
stwiname, line 248: What if they have a different order?

packages/node/src/configure/SchemaMigration.service.test.ts
stwiname, line 1: This should live in node core

packages/node-core/src/indexer/storeCache/storeCache.service.ts
```diff
@@ -51,6 +55,13 @@ export class StoreCacheService extends BaseCacheService {
     this.poiRepo = poi;
   }
 
+  get unfinalizedBlocks(): boolean {
+    return this._unfinalizedBlocks;
+  }
+  get indexCountLimit(): number {
+    return this._indexCountLimit;
+  }
```
stwiname, line 63: These don't belong here, they don't relate specificly to store cache and are just a proxy for node config

packages/node-core/src/indexer/worker/worker.service.ts
```diff
@@ -36,12 +36,14 @@ export abstract class BaseWorkerService<
   protected abstract toBlockResponse(block: B): R;
   protected abstract processFetchedBlock(block: B, dataSources: DS[]): Promise<ProcessBlockResponse>;
 
+  private nodeConfig: NodeConfig;
```
stwiname, line 39: Can be reverted

packages/node-core/src/yargs.ts
```diff
@@ -230,6 +229,12 @@ export function yargsBuilder<
                 describe: 'Number of worker threads to use for fetching and processing blocks. Disabled by default.',
                 type: 'number',
               },
+              'allow-schema-migration': {
+                demandOption: false,
+                default: false,
```
stwiname, line 234:
```suggestion
```
The default is set in node config

packages/node/src/indexer/fetch.module.ts
```diff
@@ -16,6 +16,7 @@ import {
   IProjectUpgradeService,
   PoiSyncService,
   InMemoryCacheService,
+  SchemaMigrationService,
```
stwiname, line 19: Unused

packages/utils/src/types/index.ts
```diff
@@ -4,3 +4,5 @@
 export * from './generalTypes';
 
 export * from './TypeClass';
+
+export * from './supported';
```
stwiname, line 8: Did you forget to commit this file?

packages/node-core/test/schemas/schemaDiff.json
```diff
@@ -0,0 +1,350 @@
+{
```
stwiname, line 1: Do you think this is holding too much information?

We're redefining a schema here with all the details on the models and fields. I think a better, simpler option is to just have this data type have the minimal information to changes in schemas then use the new graphql schema as the source of truth with all the details like nullable etc.

packages/node-core/src/configure/configure.module.ts
```diff
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: GPL-3.0
 
 import {handleCreateSubqueryProjectError, LocalReader, makeTempDir, ReaderFactory} from '@subql/common';
+import {SchemaMigrationService} from '@subql/node-core/configure/SchemaMigration.service';
```
stwiname, line 5: Should be a relative import

packages/node-core/src/configure/ProjectUpgrade.service.ts
```diff
@@ -119,18 +121,21 @@ export class ProjectUpgradeSevice<P extends ISubqueryProject = ISubqueryProject>
...
 
     const lastProjectChange = this.validateIndexedData(indexedDeployments);
 
+    this.#currentHeight = currentHeight;
+    this.#currentProject = this.getProject(this.#currentHeight);
+
     if (lastProjectChange) {
       this.#currentHeight = lastProjectChange;
       this.#currentProject = this.getProject(this.#currentHeight);
     }
```
stwiname, line 141: This looks like it can be simplified

packages/node-core/src/indexer/project.service.ts
```diff
@@ -370,8 +367,10 @@ export abstract class BaseProjectService<
   }
 
   private async initUpgradeService(): Promise<number | undefined> {
```
stwiname, line 366: For clarity rather than using `this._startHeight` we should use an argument here. This would change from a runtime error to a compiler error and be caught earlier.

packages/node-core/src/configure/ProjectUpgrade.service.ts
```diff
@@ -238,14 +279,29 @@ export class ProjectUpgradeSevice<P extends ISubqueryProject = ISubqueryProject>
       throw new Error('No valid projects found, this could be due to the startHeight.');
     }
 
+    isRewindable = this.rewindableCheck(projects);
```
stwiname, line 282:
```suggestion
    const isRewindable = this.rewindableCheck(projects);
```

packages/node-core/src/indexer/storeCache/storeCache.service.ts
```diff
@@ -68,20 +68,28 @@ export class StoreCacheService extends BaseCacheService {
...
-        this.getNextStoreOperationIndex.bind(this),
-        () => this.flushCache(true),
-        this._useCockroachDb
-      );
+      this.cachedModels[entity] = this.createModel(entity);
     }
     return this.cachedModels[entity] as unknown as ICachedModel<T>;
   }
 
+  createModel(entityName: string): CachedModel<any> {
```
stwiname, line 76: Can be made private

packages/node-core/src/utils/migration.ts
```diff
@@ -0,0 +1,268 @@
...
+
+export class Migration {
+  private readonly transaction: Transaction;
+  private sequelizeModels: Set<ModelStatic<any>> = new Set();
+  private rawQueries: string[] = [];
+  constructor(
+    private sequelize: Sequelize,
+    private schemaName: string,
+    private config: NodeConfig,
+    private _tx: Transaction,
```
stwiname, line 90: I think this can be removed here and `run` can create a transaction

packages/node-core/src/utils/migration.ts
```diff
@@ -0,0 +1,268 @@
...
+
+      await this.transaction.commit();
+    } catch (e) {
+      await this.transaction.rollback();
+      throw e;
+    }
+
+    await Promise.all([...this.sequelizeModels].map((m) => m.sync()));
+
+    return [...this.sequelizeModels];
```
stwiname, line 145: Do these models need to be copied into a new array?

packages/node-core/src/configure/SchemaMigration.service.ts
```diff
@@ -343,6 +261,7 @@ export class SchemaMigrationService {
     dbSchema: string,
     blockHeight: number,
     _flushCache: (flushAll?: boolean) => Promise<void>,
+    updateCacheModels: (model: ModelStatic<Model<any, any>>[]) => void,
```
stwiname, line 264: `updateCacheModels` -> `updatedModels`. This can be generic, the migration service shouldn't be concerned with how these new models are going to be used.

packages/node-core/src/configure/SchemaMigration.service.ts
```diff
@@ -343,6 +261,7 @@ export class SchemaMigrationService {
     dbSchema: string,
     blockHeight: number,
     _flushCache: (flushAll?: boolean) => Promise<void>,
```
stwiname, line 263: `flushCache`, can remove the underscore

packages/node-core/src/configure/SchemaMigration.service.ts
```diff
@@ -0,0 +1,346 @@
...
+    compareRelations(currentData.relations, nextData.relations, changes);
+    compareModels(currentData.models, nextData.models, changes);
+
+    return changes;
+  }
+
+  async run(
+    currentSchema: GraphQLSchema,
+    nextSchema: GraphQLSchema,
+    dbSchema: string,
```
stwiname, line 261: Can be moved to the constructor

packages/node-core/src/configure/SchemaMigration.service.ts
```diff
@@ -0,0 +1,346 @@
...
+    blockHeight: number,
+    _flushCache: (flushAll?: boolean) => Promise<void>,
+    updateCacheModels: (model: ModelStatic<Model<any, any>>[]) => void,
+    config: NodeConfig
```
stwiname, line 265: Can be moved to the constructor

packages/node-core/src/configure/SchemaMigration.service.ts
```diff
@@ -0,0 +1,346 @@
...
+    index1.unique === index2.unique &&
+    index1.using === index2.using
+  );
+}
+
+function hasChanged(changes: SchemaChangesType): boolean {
+  return Object.values(changes).some((change) =>
+    Array.isArray(change) ? change.length > 0 : Object.keys(change).length > 0
+  );
+}
```
stwiname, line 217: I think the functions above here can be moved into another file

packages/node-core/src/utils/migration.ts
stwiname, line 1: I think this file should live next to SchemaMigrationService (in a sub dir) as its only relevant to that and can better control the visibility to make sure its not accessible elsewhere

packages/node-core/src/indexer/storeCache/storeCache.service.ts
```diff
@@ -50,6 +50,13 @@ export class StoreCacheService extends BaseCacheService {
     this.metadataRepo = meta;
     this.poiRepo = poi;
   }
+  get config() {
+    return this._config;
+  }
+
+  get sequelize() {
+    return this._sequelize;
+  }
```
stwiname, line 59: Are these used anywhere?

stwiname (approved):

PR #2187, opened 2023-11-28
update dictionary registry json path

bz888:
Update dictionary registry path in accordance with: https://github.com/subquery/templates/pull/27


packages/common/src/constants.ts
```diff
@@ -22,7 +22,7 @@ export enum NETWORK_FAMILY {
   substrate = 'Substrate',
   cosmos = 'Cosmos',
   algorand = 'Algorand',
-  ethereum = 'Ethereum',
+  ethereum = 'evm',
```
stwiname, line 25: Are there further implications for this? 

Was it a bug and does it relate to anythign else?

stwiname (approved):

PR #2190, opened 2023-11-29
remove redundant endpoint suggestions

bz888:
# Description

As `/network-endpoints` is no longer being used on hosted Services, this should be deprecated on the CLI as well.

## Type of change

## Checklist

- [x] I have tested locally

### DIFF
--- a/packages/cli/src/controller/deploy-controller.spec.ts
+++ b/packages/cli/src/controller/deploy-controller.spec.ts
@@ -10,7 +10,6 @@ import {
   deleteDeployment,
   deploymentStatus,
   ipfsCID_validate,
-  networkEndpoints,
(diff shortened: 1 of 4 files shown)


packages/cli/src/controller/deploy-controller.spec.ts
```diff
@@ -53,13 +52,13 @@ async function deployTestProject(
     url
   );
 
-  const endpoint = await networkEndpoints(url);
+  const endpoint = 'wss://polkadot.api.onfinality.io/public-ws';
```
stwiname, line 55: I don't think we need to default to polkadot. Chances are slim that's the network being used

stwiname (approved):

PR #2221, opened 2024-01-16
add support for csv export

### LINKED ISSUE #2210: Output to CSV
As well as outputting data to Postgres there should be an option to output to CSV as well. This will allow other DBs to import data.

Requirements:
* Data should be output to CSV as well as to Postgres
* Option can be enabled via a CLI flag for a directory path
* Each entity should have its own file
* It should only append data, no deletions or updates, but entities with the same id can be entered multiple times
* Initially all data should be stored in a single file for an entity but later on we could support:
    * File rotation
    * Splitting files by block range

bz888:
# Description

Fixes #2210

Docs #https://github.com/subquery/documentation/pull/468

## Checklist

- [x] I have tested locally

### DIFF
--- a/packages/node-core/CHANGELOG.md
+++ b/packages/node-core/CHANGELOG.md
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+- Support CSV output (#2221)
 
 ## [7.0.8] - 2024-01-10
 ### Fixed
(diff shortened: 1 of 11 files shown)


jiqiang90:
Some test outcome, we can improve on

1. csv should be override by default.
image
2. `2024-01-17T00:12:28.273Z <CLI-export> ERROR Failed to export accountk  error: relation "Polkadot-starter.accountks" does not exist`
image
When export an entity that not exist but csv still been created, which should not. 
3. `2024-01-17T00:15:06.977Z <CLI-export> ERROR Failed to export account  Error: ENOENT: no such file or directory, open 'exportPath/Polkadot-starter-_metadata.csv'` if I give a export path that not exist, it should be able to create it by default. 
4. ` export-csv --outPath=./ --entities=account,_metadata` should not throw 
image

jiqiang90:
We also need a performance test with large db set, maybe this can be done on host service after prerelease

bz888:
Closing this as it is different to the desired scope

PR #2244, opened 2024-02-05
Update table creation on schema migration

bz888:
# Description
Schema migraiton should be an autonomous action that should be in one transaction
Hence, we can not use sequelize.sync(), as if migration were to fail, table creation would still be executed 

Tests on
- Cyclic schema
- Historical (complex schema, multiple relationals)
- Non-historical (complex schema, multiple relationals)
- QueryGenerators

## Type of change

## Checklist

- [x] I have tested locally

### DIFF
--- a/packages/node-core/src/indexer/store.service.ts
+++ b/packages/node-core/src/indexer/store.service.ts
@@ -45,15 +45,13 @@ import {
   createUniqueIndexQuery,
   dropNotifyFunction,
   dropNotifyTrigger,
-  generateCreateIndexStatement,
-  generateCreateTableStatement,
+  generateOrderedStatements,
   getExistedIndexesQuery,
   getFkConstraint,
   getTriggers,
   modelsTypeToModelAttributes,
   SmartTags,
   smartTags,
-  sortModels,
   syncEnums,
   updateIndexesName,
 } from '../utils';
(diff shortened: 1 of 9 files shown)


packages/node-core/src/indexer/store.service.ts
```diff
@@ -352,7 +358,44 @@ export class StoreService {
...
+      sortedModels.reverse().forEach((model: ModelStatic<any>) => {
+        const tableQuery = generateCreateTableStatement(model, schema);
+        mainQueries.push(tableQuery);
+
+        if (model.options.indexes) {
+          const indexQuery = generateCreateIndexStatement(model.options.indexes, schema, model.tableName);
+          mainQueries.push(...indexQuery);
+        }
+      });
+    }
```
stwiname, line 384: I think this can be moved into a function

packages/node-core/src/utils/sync-helper.spec.ts
```diff
@@ -0,0 +1,283 @@
...
+          onUpdate: 'CASCADE',
+        },
+      } as any;
+    });
+
+    const v = addForeignKeyStatement(mockModel);
+    expect(v[0]).toBe(
+      `ALTER TABLE "test"."test-table"
+      ADD FOREIGN KEY (transfer_id_id) 
+      REFERENCES "test"."transfers" (id) ON DELETE NO ACTION ON UPDATE CASCADE;`
```
stwiname, line 231: Id should be escaped
```suggestion
      REFERENCES "test"."transfers" ("id") ON DELETE NO ACTION ON UPDATE CASCADE;`
```

packages/node-core/src/utils/sync-helper.ts
```diff
@@ -395,3 +401,122 @@ export function addRelationToMap(
...
+
+  const primaryKeyDefinition = `, PRIMARY KEY (${primaryKeyColumns.join(', ')})`;
+
+  const tableQuery = `
+    CREATE TABLE IF NOT EXISTS "${schema}"."${tableName}" (
+      ${columnDefinitions.join(',\n      ')}${primaryKeyDefinition}
+    );
+  `;
+
+  return tableQuery.concat(`\n${comments.join('\n')}`).trim();
```
stwiname, line 443: I think it would be better to return an array of queries

packages/node-core/src/indexer/store.service.ts
stwiname: Shouldn't migrations use this too?

bz888:
Closing this pr in favour of #2251

Issue #2249, opened 2024-02-10
Support Enum creation and removal for schema migrations


bz888:
### Prerequisites

* [x] Are you running the latest version(s)?
* [x] Have you searched the [documentation for your issue](https://academy.subquery.network/)?
* [x] Did you search for an existing issue or pull request?

### Description
Currently the Schema Migration feature does not support enum introduction or removal.

stwiname:
Fixed in https://github.com/subquery/subql/pull/2251

PR #2251, opened 2024-02-12
Refactor `store.service` Add support for relations, enums, subscription on Schema Migration

bz888:
# Description

Fixes and Updates
- Support Enums
- Support Subscription
- Support Relations
- Cyclic schema
- Historical (complex schema, multiple relationals)
- Non-historical (complex schema, multiple relationals)
- QueryGenerators

Refactor store.serivce to use generate queries instead of `sequelize.sync()`

Fixes:
#2249
#2241 

Documentation:
https://github.com/subquery/documentation/pull/474

## Type of change

## Checklist

- [x] I have tested locally


packages/node-core/src/configure/migration-service/SchemaMigration.service.ts
```diff
@@ -63,62 +69,121 @@ export class SchemaMigrationService {
...
-          await migrationAction.createTable(model);
+      if (sortedAddedModels.length) {
+        for (const model of sortedAddedModels) {
+          await migrationAction.createTable(model, this.withoutForeignKeys);
         }
       }
 
-      if (Object.keys(modifiedModels).length) {
+      if (Object.keys(sortedModifiedModels).length) {
         const entities = Object.keys(modifiedModels);
```
stwiname, line 187: Mix of sorted and unsorted here?

packages/node-core/src/configure/migration-service/migration.ts
```diff
@@ -18,60 +61,115 @@ import {
...
+const logger = getLogger('db-manager');
+
 export class Migration {
   private sequelizeModels: ModelStatic<any>[] = [];
-  private rawQueries: string[] = [];
+  private mainQueries: string[] = [];
   private readonly historical: boolean;
+  private extraQueries: string[] = [];
+  private foreignKeyMap: Map<string, Map<string, SmartTags>> = new Map<string, Map<string, SmartTags>>();
+  private useSubscription: boolean;
```
stwiname, line 74: This may have been a bug previously, but if this is based on a flag, do they get removed if you enable then disable?

packages/node-core/src/configure/migration-service/migration.ts
```diff
@@ -18,60 +61,115 @@ import {
...
+  private enumTypeMap: Map<string, string> = new Map<string, string>();
+  private removedIndexes: RemovedIndexes = {};
 
   constructor(
```
stwiname, line 78:
```suggestion
  private constructor(
```

packages/node-core/src/configure/migration-service/migration.ts
```diff
@@ -18,60 +61,115 @@ import {
...
+type RemovedIndexes = Record<string, IndexesOptions[]>;
+
+const logger = getLogger('db-manager');
+
 export class Migration {
   private sequelizeModels: ModelStatic<any>[] = [];
-  private rawQueries: string[] = [];
+  private mainQueries: string[] = [];
   private readonly historical: boolean;
+  private extraQueries: string[] = [];
```
stwiname, line 72: Comments explaining the reason for 2 sets would be good
```suggestion
  private mainQueries: string[] = [];
  private extraQueries: string[] = [];
  private readonly historical: boolean;

```

packages/node-core/src/configure/migration-service/migration.ts
```diff
@@ -18,60 +61,115 @@ import {
...
   ): Promise<Migration> {
-    const modelsRelationsEnums = getAllEntitiesRelations(graphQLSchema);
-    const enumTypeMap = new Map<string, string>();
-    for (const e of modelsRelationsEnums.enums) {
-      await syncEnums(sequelize, SUPPORT_DB.postgres, e, schemaName, enumTypeMap, logger);
+    const migration = new Migration(sequelize, storeService, schemaName, config, dbType);
+    await migration.init(currentSchema);
+    return migration;
+  }
+  async init(currentSchema: GraphQLSchema | null): Promise<void> {
```
stwiname, line 105: Merge this with create, that way it cant accidentally be called from elsewhere.

packages/node-core/src/configure/migration-service/migration.ts
```diff
@@ -18,60 +61,115 @@ import {
...
+    const CurrentModelsRelationsEnums = getAllEntitiesRelations(currentSchema);
```
stwiname, line 106:
```suggestion
    const currentModelsRelationsEnums = getAllEntitiesRelations(currentSchema);
```

packages/node-core/src/configure/migration-service/migration.ts
```diff
@@ -18,60 +61,115 @@ import {
...
+
+    // Should load all keys and enums of currentSchema, as nextSchema will be added from the migration execution
+    this.loadExistingForeignKeys(CurrentModelsRelationsEnums.relations);
+    await this.loadExisitingEnums(CurrentModelsRelationsEnums.enums);
+
+    if (this.useSubscription) {
+      this.extraQueries.push(createSendNotificationTriggerFunction(this.schemaName));
     }
```
stwiname, line 114: I think rather than these mutating Migration, they should return values and then be passed in as constructor args. It adds further safety that they are called in the correct place and have no side-effects

packages/node-core/src/configure/migration-service/migration.ts
```diff
@@ -93,46 +191,80 @@ export class Migration {
...
+      if (this.dbType !== SUPPORT_DB.cockRoach) {
+        this.extraQueries.push(dropNotifyTrigger(this.schemaName, sequelizeModel.tableName));
+      }
+    }
+
+    if (!this.useSubscription && this.dbType !== SUPPORT_DB.cockRoach) {
+      this.extraQueries.push(dropNotifyFunction(this.schemaName));
+    }
+
+    this.addModelToSequelizeCache(sequelizeModel);
```
stwiname, line 254: Previously this would have been run every startup, now it will only be run on migrations

stwiname (commented):
One concern i have is that this will break local development, what happens with schema changes when not dealing with projects? The behaviour should be the same as before.

packages/node-core/src/configure/migration-service/SchemaMigration.service.ts
```diff
@@ -63,84 +70,113 @@ export class SchemaMigrationService {
...
+        for (const index of modelValue.addedIndexes) {
+          migrationAction.createIndex(modelValue.model, index);
         }
       }
 
+      if (addedRelations.length) {
+        for (const relationModel of addedRelations) {
+          migrationAction.createRelation(relationModel);
+        }
+      }
```
stwiname, line 172:
```suggestion
        for (const relationModel of addedRelations) {
          migrationAction.createRelation(relationModel);
        }
```

packages/node-core/src/configure/migration-service/migration-helpers.ts
```diff
@@ -1,36 +1,42 @@
...
+  modifiedModels: ModifiedModels;
 
   addedRelations: GraphQLRelationsType[];
   removedRelations: GraphQLRelationsType[];
 
   addedEnums: GraphQLEnumsType[];
   removedEnums: GraphQLEnumsType[];
-  allEnums: GraphQLEnumsType[];
+  modifiedEnums: GraphQLEnumsType[];
+  // allEnums: GraphQLEnumsType[];
```
stwiname, line 39: Can go?

packages/node-core/src/configure/SchemaMigration.service.spec.ts
stwiname, line 1: This should be in the same directory as the service

packages/node-core/src/configure/migration-service/migration-helpers.ts
```diff
@@ -182,27 +199,89 @@ export function schemaChangesLoggerMessage(schemaChanges: SchemaChangesType): st
...
+      addRelationToMap(relation, foreignKeyMap, model, relatedModel);
+    });
+  }
+  return foreignKeyMap;
+}
 
-    // Adding enums
-    if (schemaChanges.addedEnums.length) {
-      logMessage += `Added Enums: ${formatModels(schemaChanges.addedEnums)}\n`;
+export async function loadExistingEnums(
```
stwiname, line 257: I think this probably belongs in sync-helpers

packages/node-core/src/configure/migration-service/migration.ts
```diff
@@ -145,47 +280,231 @@ export class Migration {
...
+      this.enumTypeMap.delete(e.name);
+    }
+  }
+
+  // Sequelize model will generate follow query to create hash indexes
+  // Example SQL:  CREATE INDEX "accounts_person_id" ON "polkadot-starter"."accounts" USING hash ("person_id")
+  // This will be rejected from cockroach db due to syntax error
+  // To avoid this we need to create index manually and add to extraQueries in order to create index in db
+  private beforeHandleCockroachIndex(
+    schema: string,
```
stwiname, line 471: Cant `this.schema` be used instead?

packages/node-core/src/configure/migration-service/migration.ts
```diff
@@ -145,47 +280,231 @@ export class Migration {
...
+    // UPDATE: this comment got syntax error with cockroach db, disable it for now. Waiting to be fixed.
+    // See https://github.com/cockroachdb/cockroach/issues/44135
+    const enumTypeName = enumNameToHash(e.name);
+    let type = `"${this.schemaName}"."${enumTypeName}"`;
+    let [results] = await this.sequelize.query(
+      `SELECT pg_enum.enumlabel as enum_value
+         FROM pg_type t JOIN pg_enum ON pg_enum.enumtypid = t.oid JOIN pg_catalog.pg_namespace n ON n.oid = t.typnamespace
+         WHERE t.typname = ? AND n.nspname = ? order by enumsortorder;`,
+      {replacements: [enumTypeName, this.schemaName]}
+    );
```
stwiname, line 418: This is a very similar query to in `getEnumDeprecated` we should try and reuse the same parts of the query

packages/node-core/src/configure/migration-service/migration.ts
```diff
@@ -145,47 +280,231 @@ export class Migration {
...
+    const tableName = modelToTableName(relation.from);
+    const fkConstraint = getFkConstraint(tableName, relation.foreignKey);
+
+    this.mainQueries.unshift(dropRelationStatement(this.schemaName, tableName, fkConstraint));
+
+    // TODO remove from sequelize model
+    // Should update sequelize model cache
+  }
+
+  async createEnum(e: GraphQLEnumsType): Promise<void> {
```
stwiname, line 403: I think this function should be reworked to be non-async so that it matches the other functions. 

We can move the async stuff to the create function.

packages/node-core/src/configure/migration-service/migration.ts
```diff
@@ -145,47 +280,231 @@ export class Migration {
...
+    let type = `"${this.schemaName}"."${enumTypeName}"`;
+    let [results] = await this.sequelize.query(
+      `SELECT pg_enum.enumlabel as enum_value
+         FROM pg_type t JOIN pg_enum ON pg_enum.enumtypid = t.oid JOIN pg_catalog.pg_namespace n ON n.oid = t.typnamespace
+         WHERE t.typname = ? AND n.nspname = ? order by enumsortorder;`,
+      {replacements: [enumTypeName, this.schemaName]}
+    );
+
+    const enumTypeNameDeprecated = `${this.schemaName}_enum_${enumNameToHash(e.name)}`;
+    const resultsDeprecated = await getEnumDeprecated(this.sequelize, enumTypeNameDeprecated);
```
stwiname, line 421: This could be run in parallel with the above query

packages/node-core/src/configure/migration-service/migration.ts
```diff
@@ -145,47 +280,231 @@ export class Migration {
...
+    if (resultsDeprecated.length !== 0) {
+      results = resultsDeprecated;
```
stwiname, line 423: If there is a migration then there could be a combination of enums in results and resultsDeprecated

packages/node-core/src/configure/migration-service/migration.ts
```diff
@@ -145,47 +280,231 @@ export class Migration {
...
+        `@enum\\n@enumName ${e.name}${e.description ? `\\n ${e.description}` : ''}`
+      );
+
+      queries.push(commentOnEnumStatement(type, comment));
+    }
+    this.mainQueries.unshift(...queries);
+    this.enumTypeMap.set(e.name, type);
+  }
+
+  dropEnum(e: GraphQLEnumsType): void {
```
stwiname, line 457: What happens if this is a deprecated enum?

packages/node-core/src/configure/migration-service/migration.ts
```diff
@@ -2,76 +2,177 @@
...
+  Utils,
+} from '@subql/x-sequelize';
 import {GraphQLSchema} from 'graphql';
-import Pino from 'pino';
+import {isEqual} from 'lodash';
 import {StoreService} from '../../indexer';
-import {formatAttributes, formatColumnName, generateHashedIndexName, modelToTableName, syncEnums} from '../../utils';
-import {getColumnOption, modelsTypeToModelAttributes} from '../../utils/graphql';
+import {getLogger} from '../../logger';
 import {
```
stwiname, line 28: Based on the massive number of imports 
`import * as utils from '../../utils`

stwiname, line 28: The naming of these functions is inconsistent, some end in `Query` others in `Statement` and some have nothing.

packages/node-core/src/configure/migration-service/migration.ts
```diff
@@ -2,76 +2,177 @@
...
 
-  constructor(
+  private constructor(
     private sequelize: Sequelize,
     private storeService: StoreService,
     private schemaName: string,
     private config: NodeConfig,
-    private enumTypeMap: Map<string, string>
+    private dbType: SUPPORT_DB,
+    // TODO refactor load map funcs
```
stwiname, line 98: Still a todo?

packages/node-core/src/configure/migration-service/migration.ts
```diff
@@ -2,76 +2,177 @@
...
+    private initForeignKeyMap: Map<string, Map<string, SmartTags>>,
+    private initEnumTypeMap: Map<string, string>,
+    private existingIndexes: {indexname: string}[]
   ) {
     this.historical = !config.disableHistorical;
+    this.useSubscription = config.subscription;
+
+    // TODO test if this is working, enable, then disable
```
stwiname, line 106: Still a todo?

packages/node-core/src/configure/migration-service/migration.ts
```diff
@@ -2,76 +2,177 @@
...
+      schemaName,
+      config,
+      dbType,
+      foreignKeyMap,
+      enumTypeMap,
+      indexesResult
+    );
+
+    if (migration.useSubscription) {
+      migration.extraQueries.push(createSendNotificationTriggerFunction(schemaName));
```
stwiname, line 144: This could be in the constructor

packages/node-core/src/configure/migration-service/migration.ts
stwiname:
```suggestion
  private readonly useSubscription: boolean;
```

packages/node-core/src/configure/migration-service/migration.ts
```diff
@@ -2,76 +2,177 @@
...
+  private foreignKeyMap: Map<string, Map<string, SmartTags>>;
+  private enumTypeMap: Map<string, string>;
+  // TODO i think this can be removed
+  private removedIndexes: RemovedIndexes = {};
 
-  constructor(
+  private constructor(
     private sequelize: Sequelize,
     private storeService: StoreService,
     private schemaName: string,
```
stwiname, line 95:
```suggestion
    private readonly schemaName: string,
```

packages/node-core/src/configure/migration-service/migration.ts
```diff
@@ -2,76 +2,177 @@
...
     private config: NodeConfig,
-    private enumTypeMap: Map<string, string>
+    private dbType: SUPPORT_DB,
```
stwiname, line 97:
```suggestion
    private readonly dbType: SUPPORT_DB,
```

stwiname:
There still seems to be quite a few TODOs scattered about

packages/node-core/src/db/index.ts
```diff
@@ -2,3 +2,5 @@
 // SPDX-License-Identifier: GPL-3.0
 
 export * from './db.module';
+export * from './migration-service';
+export * from './sync-helper';
```
stwiname, line 6: I think if we move sequelizeHelper to this directory then we can remove this export

packages/node-core/src/indexer/store.service.ts
```diff
@@ -187,8 +160,9 @@ export class StoreService {
     this._modelsRelations = modelsRelations;
 
     try {
-      await this.syncSchema(schema, this.config.subscription);
+      await this.syncSchema(schema);
     } catch (e: any) {
+      console.trace(e);
```
stwiname, line 165: Can go

packages/node-core/src/db/migration-service/migration.ts
```diff
@@ -0,0 +1,474 @@
...
+   */
+  private mainQueries: string[] = [];
+  private extraQueries: string[] = [];
+  private readonly historical: boolean;
+  private readonly useSubscription: boolean;
+  private foreignKeyMap: Map<string, Map<string, syncHelper.SmartTags>> = new Map<
+    string,
+    Map<string, syncHelper.SmartTags>
+  >();
+  private enumTypeMap: Map<string, {enumValues: string[]; name?: string; type?: string}>;
```
stwiname, line 55: Make `{enumValues: string[]; name?: string; type?: string}` into a type. Its used multiple places

packages/node-core/src/utils/graphql.ts
```diff
@@ -15,24 +15,31 @@ import {
...
 
   const type = field.isEnum
-    ? `${enums.get(field.type)}${field.isArray ? '[]' : ''}`
+    ? `${enumType}${field.isArray ? '[]' : ''}`
     : field.isArray
     ? getTypeByScalarName('Json')?.sequelizeType
     : getTypeByScalarName(field.type)?.sequelizeType;
 
-  if (type === undefined) {
+  if (type === undefined || type === 'undefined') {
```
stwiname, line 42: Whats going on here? Why would it be `'undefined'`?

stwiname (commented):
There still seem to be a couple of TODOs in places

stwiname (approved):

PR #2255, opened 2024-02-18
add profiler to project init

bz888:
Adding a profilter to project.init


packages/node/src/indexer/project.service.ts
```diff
@@ -63,6 +64,11 @@ export class ProjectService extends BaseProjectService<
     );
   }
 
+  @profiler()
```
stwiname, line 67: Why not add profiler to the super class?

stwiname (approved):

PR #2267, opened 2024-02-23
[release] 20240223

bz888:
release for :
- node-core
- node
- cli
- common


packages/common/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@subql/common",
-  "version": "3.4.0",
+  "version": "3.4.1",
```
stwiname, line 3: Common-substrate will also need to be released with this

stwiname (approved):

PR #2275, opened 2024-02-27
Remove scale-batch-size flag that had no functionality

stwiname:
# Description
The `scale-batch-size` flag had no effect or change the actual batch size so any related code can be removed.

Fixes # (issue)

## Checklist

- [x] I have tested locally


packages/node-core/src/indexer/fetch.service.ts
```diff
@@ -28,7 +28,6 @@ export abstract class BaseFetchService<
   private _latestBestHeight?: number;
   private _latestFinalizedHeight?: number;
   private isShutdown = false;
-  private batchSizeScale = 1;
```
stwiname, line 31: This value is never used anywhere and why the change is not useful

PR #2283, opened 2024-03-04
Fix/hot schema reload, missing metadata keys

bz888:
# Description

Fixes # (issue)

## Type of change

## Checklist

- [x] I have tested locally

### DIFF
--- a/packages/node-core/src/db/sync-helper.ts
+++ b/packages/node-core/src/db/sync-helper.ts
@@ -72,12 +72,18 @@ function escapedName(...args: string[]): string {
   return args.map((a) => `"${a}"`).join('.');
 }
 
-function commentOn(type: 'CONSTRAINT' | 'TABLE' | 'COLUMN' | 'FUNCTION', entity: string, comment: string): string {
-  return `COMMENT ON ${type} ${entity} IS E'${comment}'`;
+function commentOn(
+  type: 'CONSTRAINT' | 'TABLE' | 'COLUMN' | 'FUNCTION',
+  entity: string,
+  comment: string,
+  constraint?: string
+): string {
+  const constraintPart = constraint ? `${constraint} ON ` : '';
+  return `COMMENT ON ${type} ${constraintPart}${entity} IS E'${comment}';`;
 }
 
 export function commentConstraintQuery(schema: string, table: string, constraint: string, comment: string): string {
-  return commentOn('CONSTRAINT', escapedName(schema, table), comment);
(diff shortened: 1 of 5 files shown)


packages/node-core/src/indexer/storeCache/cacheMetadata.ts
```diff
@@ -11,6 +11,7 @@ import {ICachedModelControl} from './types';
 
 type MetadataKey = keyof MetadataKeys;
 const incrementKeys: MetadataKey[] = ['processedBlockCount', 'schemaMigrationCount'];
+type IncrementalMetadataKey = typeof incrementKeys[number];
```
bz888, line 14: Is this the correct way to do typing here ?

packages/node-core/src/indexer/storeCache/cacheMetadata.test.ts
```diff
@@ -0,0 +1,67 @@
...
+  afterAll(async () => {
+    await sequelize.close();
+  });
+
+  it('Ensure increment keys are created on _metadata table', async () => {
+    schema = '"metadata-test-1"';
+    await sequelize.createSchema(schema, {});
+    await sequelize.query(
+      `CREATE TABLE IF NOT EXISTS ${schema}."_metadata" ("key" VARCHAR(255) , "value" JSONB, "createdAt" TIMESTAMP WITH TIME ZONE NOT NULL, "updatedAt" TIMESTAMP WITH TIME ZONE NOT NULL, PRIMARY KEY ("key"));`
+    );
```
stwiname, line 40: I think here it would make sense to use `MetadataFactory`. If that we're to change we could tell if it breaks tests

stwiname (approved):

packages/node-core/src/db/sync-helper.ts
```diff
@@ -72,12 +72,18 @@ function escapedName(...args: string[]): string {
   return args.map((a) => `"${a}"`).join('.');
 }
 
-function commentOn(type: 'CONSTRAINT' | 'TABLE' | 'COLUMN' | 'FUNCTION', entity: string, comment: string): string {
-  return `COMMENT ON ${type} ${entity} IS E'${comment}'`;
+function commentOn(
+  type: 'CONSTRAINT' | 'TABLE' | 'COLUMN' | 'FUNCTION',
+  entity: string,
+  comment: string,
+  constraint?: string
```
stwiname, line 79: I don't think this should live here. Keep the changes to `commentConstraintQuery`

PR #2302, opened 2024-03-15
Dynamic ds meta optimisation

stwiname:
# Description
### Problem
With a large number of dynamic ds every time a new one is created it will rewrite the whole set of params to the db. This results in a slow query which has to transfer a large amount of data. 

### Solution
To have a specific query to append only new datasource parameters to the metadata.

## Type of change

## Checklist

- [ ] I have tested locally

### DIFF
--- a/packages/node-core/src/indexer/project.service.ts
+++ b/packages/node-core/src/indexer/project.service.ts
@@ -109,8 +109,9 @@ export abstract class BaseProjectService<
 
       // Init metadata before rest of schema so we can determine the correct project version to create the schema
       await this.storeService.initCoreTables(this._schema);
-      await this.dynamicDsService.init(this.storeService.storeCache.metadata);
       await this.ensureMetadata();
+      // DynamicDsService is dependent on metadata so we need to ensure it exists first
+      await this.dynamicDsService.init(this.storeService.storeCache.metadata);
 
       /**
        * WARNING: The order of the following steps is very important.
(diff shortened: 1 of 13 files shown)


packages/node-core/src/meta/meta.service.ts
```diff
@@ -29,8 +29,8 @@ export abstract class BaseMetaService {
   private lastProcessedTimestamp?: number;
   private processedBlockCount?: number;
 
-  constructor(private storeCacheService: StoreCacheService, private config: NodeConfig) {
-    UPDATE_HEIGHT_INTERVAL = config.storeFlushInterval * 1000;
+  constructor(private storeCacheService: StoreCacheService, config: NodeConfig) {
+    // TODO update UPDATE_HEIGHT_INTERVAL should be configureable based on config.storeFlushInterval * 1000 but need to use SchedulerRegistry for that
```
stwiname, line 33: I've left this as a TODO so we don't create a breaking change and can make a patch release. 
Im not sure how important it is to be configurable, and it probably makes sense to remove this entirely and have fetch service be responsible for this with its intervals for finalized/latest

packages/node-core/src/indexer/project.service.ts
```diff
@@ -198,7 +200,7 @@ export abstract class BaseProjectService<
 
     this.eventEmitter.emit(IndexerEvent.NetworkMetadata, this.apiService.networkMeta);
 
-    const keys = [
+    const keys: (keyof MetadataKeys)[] = [
```
stwiname, line 203: I think `ensureMetadata` should probably live in cacheMetadata and be called from here instead. Its out of scope of this PR

jiqiang90 (approved):

PR #2322, opened 2024-03-22
Fix dictionary issues

stwiname:
# Description
- Fixes exports from tests leading to tests not being able to be run independently
- Fixes updating v2 metadata end block on queries
  - Wrong data type
  - Wrong casing on type 
- Fixes wrong error message on v2 query failure

### DIFF
--- a/packages/node-core/src/indexer/dictionary/v2/types.ts
+++ b/packages/node-core/src/indexer/dictionary/v2/types.ts
@@ -11,7 +11,7 @@ export interface FieldSelector {
 
 export interface RawDictionaryResponseData<RFB> {
   blocks: RFB[]; //raw fat blocks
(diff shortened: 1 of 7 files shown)


packages/node-core/src/indexer/dictionary/v2/types.ts
```diff
@@ -11,7 +11,7 @@ export interface FieldSelector {
 
 export interface RawDictionaryResponseData<RFB> {
   blocks: RFB[]; //raw fat blocks
-  blockRange: number[];
+  blockRange: [string, string]; // Hex encoded numbers
```
stwiname, line 14: This will need to be fixed in takoyaki

jiqiang90 (approved):

PR #2331, opened 2024-03-27
Fix fetch miss check ds endHeight with modulo number

jiqiang90:
# Description

Fixes # (issue)

## Type of change

## Checklist

- [ ] I have tested locally

### DIFF
--- a/packages/node-core/CHANGELOG.md
+++ b/packages/node-core/CHANGELOG.md
@@ -5,6 +5,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Fixed
+- Fix unable to find index in `modelIndexedFields` due to a special case entity name (2327)
 - Fix multi-chain has `block-range` issue due to migartion handle historical status incorrectly
 
 ## [Unreleased]
--- a/packages/node-core/CHANGELOG.md
+++ b/packages/node-core/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - Fix unable to find index in `modelIndexedFields` due to a special case entity name (2327)
 - Fix multi-chain has `block-range` issue due to migartion handle historical status incorrectly
+- Fix in fetch service when dictionary scopedDictionaryEntries failed keeps retrying with infinite loop
 - Fix modulo block didn't apply correctly with multiple dataSources (#2331)
 
 ## [Unreleased]
--- a/packages/node-core/CHANGELOG.md
+++ b/packages/node-core/CHANGELOG.md
@@ -4,18 +4,30 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
 ### Fixed
-- Fix unable to find index in `modelIndexedFields` due to a special case entity name (2327)
-- Fix multi-chain has `block-range` issue due to migartion handle historical status incorrectly
-- Fix in fetch service when dictionary scopedDictionaryEntries failed keeps retrying with infinite loop
 - Fix modulo block didn't apply correctly with multiple dataSources (#2331)
 
-## [Unreleased]
+## [8.0.0] - 2024-03-28
 ### Added
 - Add service to support for dictionary v2 (#2257)
+- Support for ordering with store `getBy*` methods (#2325)
 
 ### Removed
 - `@subql/apollo-links` for resolving dictionary endpoints and matching `dictionary-resolver` flag (#2305)
+
+### Fixed
+- Various issues with store cache (#2325)
+    - `getByFields` ignoring mutations from the current block
+    - Order and offset issues when using store `getByFields`, `getByField` and `getOnByField`
+    - Being able to mutate data in the store cache without calling methods
+    - Matching cache data with `getByFields`, `getByField` and `getOneByField`
+
+## [7.5.1] - 2024-03-26
+### Fixed
+- Fix unable to find index in `modelIndexedFields` due to a special case entity name (2327)
+- Fix multi-chain has `block-range` issue due to migartion handle historical status incorrectly
 
 ## [7.5.0] - 2024-03-20
 ### Changed
@@ -625,7 +637,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 - Move blockchain agnostic code from `node` to `node-core` package. (#1222)
 
-[Unreleased]: https://github.com/subquery/subql/compare/node-core/7.5.0...HEAD
+[Unreleased]: https://github.com/subquery/subql/compare/node-core/8.0.0...HEAD
+[8.0.0]: https://github.com/subquery/subql/compare/node-core/7.5.1...node-core/8.0.0
+[7.5.1]: https://github.com/subquery/subql/compare/node-core/7.5.0...node-core/7.5.1
 [7.5.0]: https://github.com/subquery/subql/compare/node-core/7.4.3...node-core/7.5.0
 [7.4.3]: https://github.com/subquery/subql/compare/node-core/7.4.2...node-core/7.4.3
 [7.4.2]: https://github.com/subquery/subql/compare/node-core/7.4.1...node-core/7.4.2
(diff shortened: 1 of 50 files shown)


packages/node/src/indexer/fetch.service.ts
```diff
@@ -83,12 +83,20 @@ export class FetchService extends BaseFetchService<
...
+   * @param endHeight
+   * @protected
+   */
+  protected getModulos(startHeight: number, endHeight: number): number[] {
+    const modulos = getModulos(
       this.projectService.getAllDataSources(),
       isCustomDs,
       SubstrateHandlerKind.Block,
     );
+    return modulos.filter((n) => n >= startHeight && n <= endHeight);
```
stwiname, line 99: We should move this filter to node core, then we can avoid a breaking change and also reduce duplicate code

stwiname (commented):
Needs a changelog

packages/node-core/src/indexer/fetch.service.ts
```diff
@@ -36,7 +36,8 @@ export abstract class BaseFetchService<DS extends BaseDataSource, B extends IBlo
 
   // The rough interval at which new blocks are produced
   protected abstract getChainInterval(): Promise<number>;
-  protected abstract getModulos(): number[];
+  // This return all modulo numbers in the filters of all dataSources
+  protected abstract getAllModuloNumbers(): number[];
```
stwiname, line 40: This is still a breaking change. The name should remain the same

packages/node-core/src/indexer/fetch.service.ts
```diff
@@ -194,11 +201,17 @@ export abstract class BaseFetchService<DS extends BaseDataSource, B extends IBlo
...
    * @param startBlockHeight
    * @param endBlockHeight is either FinalizedHeight or BestHeight, ensure ModuloBlocks not greater than this number
+   * @param moduloNumbers valid modulo number from handler filters
    */
-  getEnqueuedModuloBlocks(startBlockHeight: number, endBlockHeight: number): (IBlock<FB> | number)[] {
+  private getEnqueuedModuloBlocks(
+    startBlockHeight: number,
+    endBlockHeight: number,
+    moduloNumbers: number[]
+  ): (IBlock<FB> | number)[] {
```
stwiname, line 210:
```suggestion
  ): number[] {
```

packages/node-core/src/indexer/fetch.service.ts
```diff
@@ -194,11 +201,17 @@ export abstract class BaseFetchService<DS extends BaseDataSource, B extends IBlo
    *
    * @param startBlockHeight
    * @param endBlockHeight is either FinalizedHeight or BestHeight, ensure ModuloBlocks not greater than this number
+   * @param moduloNumbers valid modulo number from handler filters
    */
-  getEnqueuedModuloBlocks(startBlockHeight: number, endBlockHeight: number): (IBlock<FB> | number)[] {
+  private getEnqueuedModuloBlocks(
```
stwiname, line 206: Why not merge this with `getModulorBlocks` and move a lot of the logic from here https://github.com/subquery/subql/pull/2331/files#diff-04fd5e8c6ab146427c7f38891863dab7467a38756570af71f587358719cd9c71R251-R265 into it too

packages/node-core/src/indexer/fetch.service.ts
```diff
@@ -278,25 +258,65 @@ export abstract class BaseFetchService<DS extends BaseDataSource, B extends IBlo
...
+    // If there are modulos hanlders only, then number of moduloNumbers should be match number of with handlers
+    const moduloNumbers = this.getModulos(relevantDS);
+    const handlers = [...relevantDS.map((ds) => ds.mapping.handlers)].flat();
+    return !!handlers.length && moduloNumbers.length === handlers.length;
+  }
+
+  private getRelevantDsDetails(startBlockHeight: number): {endHeight: number | undefined; value: DS[]} {
+    const details = this.projectService.getDataSourcesMap().getDetails(startBlockHeight);
+    assert(details, `Datasources not found for height ${startBlockHeight}`);
+    return {endHeight: details?.endHeight, value: details?.value};
```
stwiname, line 306:
```suggestion
    return {endHeight: details.endHeight, value: details.value};
```

packages/node-core/src/indexer/fetch.service.ts
```diff
@@ -278,25 +258,65 @@ export abstract class BaseFetchService<DS extends BaseDataSource, B extends IBlo
...
   }
 
+  // get all modulo numbers with a specific block ranges
+  private getModuloBlocks(startHeight: number, endHeight: number): number[] {
+    // no modulos in the filters been found
+    if (this.getModulos(this.projectService.getAllDataSources()).length < 0) return [];
+    // Find relevant ds
+    const {endHeight: rangeEndHeight, value: relevantDS} = this.getRelevantDsDetails(startHeight);
+    // Min of current ds endHeight
+    const minDsEndHeight = Math.min(...relevantDS.map((d) => d?.endBlock ?? Number.MAX_SAFE_INTEGER));
```
stwiname, line 268: I think this is redundant, doesn't `rangeEndHeight` always return the same or less value?

packages/node-core/src/indexer/fetch.service.ts
```diff
@@ -278,25 +258,65 @@ export abstract class BaseFetchService<DS extends BaseDataSource, B extends IBlo
     }
   }
 
+  // get all modulo numbers with a specific block ranges
+  private getModuloBlocks(startHeight: number, endHeight: number): number[] {
+    // no modulos in the filters been found
+    if (this.getModulos(this.projectService.getAllDataSources()).length < 0) return [];
```
stwiname, line 264: Is this check needed? It seems like a place to easily introduce bugs

packages/node-core/src/indexer/fetch.service.spec.ts
```diff
@@ -417,6 +439,47 @@ describe('Fetch Service', () => {
     expect(enqueueBlocksSpy).toHaveBeenLastCalledWith([2, 3, 4, 5, 6, 7, 8, 9, 10, 11], 11);
   });
 
+  it('when enqueueu ds endHeight less than modulo height, should not include any modulo', async () => {
```
stwiname, line 442:
```suggestion
  it('when enqueue ds endHeight less than modulo height, should not include any modulo', async () => {
```

stwiname (commented):
This looks much better

packages/node-core/src/indexer/fetch.service.ts
jiqiang90: Yes, looks like a mistake here. I thought `rangeEndHeight` would be next dataSource startHeight -1.

packages/node-core/CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- Fix modulo block didn't apply correctly with multiple dataSources (#2331)
```
stwiname, line 10:
```suggestion
- Fix modulo filter not applying correctly with multiple dataSources (#2331)
```

packages/node-core/src/indexer/fetch.service.spec.ts
```diff
@@ -349,7 +378,7 @@ describe('Fetch Service', () => {
 
   it('enqueues modulo blocks WITHOUT dictionary', async () => {
     // Set modulos to every 3rd block. We only have 1 data source
-    fetchService.modulos = [3];
+    (fetchService as any).getModulos = () => [3];
```
stwiname, line 381: Because we have a testing implementation of FetchService, we should add functionality to that rather than doing `any` casts

packages/node-core/src/indexer/fetch.service.spec.ts
```diff
@@ -372,8 +401,81 @@ describe('Fetch Service', () => {
...
+    dataSources[0].mapping.handlers.push({
+      kind: 'mock/Handler',
+      handler: 'mockFunction',
+      filter: {},
+    });
+    // 2 handlers, only 1 has modulo filter
+    expect((fetchService as any).useModuloHandlersOnly(dataSources)).toBeFalsy();
+  });
+
+  it('can exclude modulo blocks', () => {
```
stwiname, line 434: Theres too much use of `any` in these tests. It makes them very fragile and to me is a sign the code is not easy to work with. 

We have a TestFetchService which can be extended to make setting test data easier, but calling functions with any and validating the result is poor practice and makes it very hard to make any changes to fetch service because we have no type checking and private functions shouldn't need to be concerned with any tests.

packages/node/CHANGELOG.md
```diff
@@ -6,10 +6,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
```
stwiname, line 9: The fix isn't on node. Changelog should be something like `Changed: update modulos to match changes with node core`

stwiname (approved):

PR #2358, opened 2024-04-16
cli migrate

jiqiang90:
# Description

Fixes # (issue)

## Type of change

## Checklist

- [ ] I have tested locally


packages/cli/src/commands/migrate/index.ts
```diff
@@ -0,0 +1,51 @@
...
+  DEFAULT_SUBQL_SCHEMA,
+} from '../../constants';
+import {migrateManifest, migrateSchema, prepareProject} from '../../controller/migrate';
+import {getNetworkFamily} from '../../utils';
+
+export default class Migrate extends Command {
+  static description = 'Schema subgraph project to subquery project';
+
+  static flags = {
+    file: Flags.string({char: 'f', description: 'specify subgraph file/directory path'}),
```
stwiname, line 21: This should be able to be a git url. We should download to a temp dir if it is

packages/cli/src/commands/migrate/index.ts
```diff
@@ -0,0 +1,51 @@
...
+    if (!lstatSync(subgraphDir).isDirectory() || lstatSync(subqlDir).isDirectory()) {
+      this.error('migrate from Subgraph failed, input and output path need to be directory');
+    }
+
+    const subgraphManifestPath = path.join(subgraphDir, DEFAULT_SUBGRAPH_MANIFEST);
+    const subgraphSchemaPath = path.join(subgraphDir, DEFAULT_SUBGRAPH_SCHEMA);
+
+    const subqlManifestPath = path.join(subqlDir, DEFAULT_SUBQL_MANIFEST);
+    const subqlSchemaPath = path.join(subqlDir, DEFAULT_SUBQL_SCHEMA);
+    cli.action.start('Preparing migrate project');
```
stwiname, line 40:
```suggestion
    cli.action.start('Preparing to migrate project');
```

packages/cli/src/commands/migrate/index.ts
```diff
@@ -0,0 +1,51 @@
...
+    try {
+      await prepareProject(networkFamily, subqlDir);
+      await migrateManifest(networkFamily, subgraphManifestPath, subqlManifestPath);
+      await migrateSchema(networkFamily, subqlSchemaPath, subgraphSchemaPath);
+      // await migrateMapping(network, subgraphManifestPath,subqlManifestPath)
+    } catch (e) {
+      this.error(`Migrate project failed: ${e}`);
```
stwiname, line 47: Do you think this should clean up any files that were created?

packages/cli/src/commands/migrate/manifest.ts
stwiname, line 1: We don't need commands for each file, just the index one is fine

packages/cli/src/controller/migrate/manifest/ethereum.ts
```diff
@@ -0,0 +1,74 @@
...
+    kind: EthereumDatasourceKind.Runtime,
+    assets: new Map(ds.mapping.abis.map((a) => [a.name, {file: a.file}])),
+    mapping: {
+      file: DEFAULT_HANDLER_BUILD_PATH,
+      handlers: [
+        ...((ds.mapping.blockHandlers
+          ? ds.mapping.blockHandlers.map((h) => {
+              return {
+                handler: h.handler,
+              };
```
stwiname, line 32: I this missing `kind: EthereumHandlerKind.Block`?

packages/cli/src/controller/migrate/manifest/migrate-manifest.controller.ts
```diff
@@ -0,0 +1,117 @@
...
+export function subgraphTemplateToSubqlTemplate(
+  network: NETWORK_FAMILY,
+  subgraphTemplates: SubgraphTemplate[]
+): TemplateKind[] {
+  const convertFunction = networkTemplateConverters[network as NETWORK_FAMILY];
+  return subgraphTemplates.map((t) => convertFunction(t));
+}
+
+export function subgraphDsToSubqlDs(network: NETWORK_FAMILY, subgraphDs: SubgraphDataSource[]): DatasourceKind[] {
+  const convertFunction = networkDsConverters[network];
```
stwiname, line 87: Needs to check if `convertFunction` is undefined and throw. You should look at enabling stricter TS settings like we have in node core. 

I imagine its too much work to get every build issue fixed but enabling it locally then seeing if there are errors in the new files would be a good start

packages/cli/src/types.ts
```diff
@@ -199,3 +197,82 @@ export interface GenerateDeploymentChainInterface {
   flags: DeploymentFlagsInterface;
 }
 
+export interface SubgraphTemplateSource {
```
stwiname, line 200: I think these can live in a separate file, lets keep subgraph types separate to ours where possible

packages/cli/src/controller/migrate/manifest/migrate-manifest.controller.ts
```diff
@@ -0,0 +1,117 @@
...
+  try {
+    await renderTemplate(PROJECT_TEMPLATE_PATH, projectPath, {
+      props: {
+        importTypes: {
+          network: 'ethereum',
+          projectClass: 'EthereumProject',
+          projectDatasourceKind: 'EthereumDatasourceKind',
+          projectHandlerKind: 'EthereumHandlerKind',
+        },
+        projectJson: project,
```
stwiname, line 32: What happens with enums here? We want the manifest to be TS not JSON

packages/cli/src/controller/migrate/abis/migrate-abis.controller.ts
```diff
@@ -0,0 +1,40 @@
...
+async function copyAbiFilesToTargetPathAsync(abiFiles: string[], targetPath: string): Promise<void> {
+  try {
+    for (const file of abiFiles) {
+      const fileName = path.basename(file);
+      const targetFilePath = path.join(targetPath, fileName);
+      await fs.promises.copyFile(file, targetFilePath);
+    }
+    console.log('ABI files copied successfully.');
+  } catch (error) {
+    console.error('Error copying ABI files:', error);
```
stwiname, line 27: Should this throw and fail?

packages/cli/src/controller/migrate/abis/migrate-abis.controller.ts
```diff
@@ -0,0 +1,40 @@
...
+import fs from 'fs';
+import path from 'path';
+import {uniqWith} from 'lodash';
+import {DEFAULT_ABI_DIR} from '../../generate-controller';
+import {SubgraphDataSource, SubgraphProject} from '../types';
+
+function extractAllAbiFiles(dataSources: SubgraphDataSource[]): string[] {
+  return uniqWith(
+    dataSources.flatMap((dataSource) =>
+      dataSource.mapping.abis.map((abi) => abi.file).filter((file) => file !== null && file !== undefined)
```
stwiname, line 13:
```suggestion
      dataSource.mapping.abis.map((abi) => abi.file).filter(Boolean)
```

packages/cli/src/controller/migrate/abis/migrate-abis.controller.ts
```diff
@@ -0,0 +1,40 @@
...
+    )
+  );
+}
+
+async function copyAbiFilesToTargetPathAsync(abiFiles: string[], targetPath: string): Promise<void> {
+  try {
+    for (const file of abiFiles) {
```
stwiname, line 20: This could use Promise.all

packages/cli/src/controller/migrate/constants.ts
```diff
@@ -0,0 +1,66 @@
...
+      path: 'Ethereum/ethereum-starter',
+      description: '',
+      remote: 'https://github.com/subquery/ethereum-subql-starter',
+    },
+  },
+};
+
+// @ts-ignore
+export const graphNetworkNameChainId: Record<NETWORK_FAMILY, Record<string, string>> = {
+  [NETWORK_FAMILY.ethereum]: {mainnet: '1'},
```
stwiname, line 58: Mainnet just means any ethereum network i thought. Do we need to prompt the user?

packages/cli/src/controller/migrate/constants.ts
```diff
@@ -0,0 +1,66 @@
...
+      name: 'eth-starter',
+      path: 'Ethereum/ethereum-starter',
+      description: '',
+      remote: 'https://github.com/subquery/ethereum-subql-starter',
+    },
+  },
+};
+
+// @ts-ignore
+export const graphNetworkNameChainId: Record<NETWORK_FAMILY, Record<string, string>> = {
```
stwiname, line 57:
```suggestion
export const graphNetworkNameChainId: Partial<Record<NETWORK_FAMILY, Record<string, string>>> = {
```

packages/cli/src/controller/migrate/manifest/ethereum.ts
```diff
@@ -0,0 +1,96 @@
...
+        break;
+      case EthereumHandlerKind.Event:
+        h.migrateHandlerType = 'EthereumHandlerKind.Event';
+        break;
+      case EthereumHandlerKind.Call:
+        h.migrateHandlerType = 'EthereumHandlerKind.Call';
+        break;
+      default:
+        break;
+    }
```
stwiname, line 79: Would this work? Im pretty sure there is a way to get an enum value
```ts
h.migrateHandlerType = `EthereumHandlerKind.${h.kind.toString()}`
```

packages/cli/src/controller/migrate/manifest/ethereum.ts
```diff
@@ -0,0 +1,96 @@
...
+      file: DEFAULT_HANDLER_BUILD_PATH,
+      handlers: [
+        ...((ds.mapping.blockHandlers
+          ? ds.mapping.blockHandlers.map((h) => {
+              return {
+                kind: EthereumHandlerKind.Block,
+                handler: h.handler,
+              };
+            })
+          : []) as EthereumBlockHandler[]),
```
stwiname, line 35: I think this is more readable
```suggestion
        ...(ds.mapping.blockHandlers ?? []).map((h) => {
              return {
                kind: EthereumHandlerKind.Block,
                handler: h.handler,
              };
            })) as EthereumBlockHandler[]),
```

packages/cli/src/controller/migrate/manifest/ethereum.ts
```diff
@@ -0,0 +1,96 @@
...
+        h.migrateHandlerType = 'EthereumHandlerKind.Call';
+        break;
+      default:
+        break;
+    }
+  });
+}
+
+export function convertEthereumDs(ds: SubgraphDataSource): MigrateDatasourceKind<EthereumDs> {
+  const subqlDs = baseDsConversion(ds) as MigrateDatasourceKind<EthereumDs>;
```
stwiname, line 84: It would be good to remove these casts, could `baseDsConversion` be made generic?

packages/cli/src/controller/migrate/manifest/migrate-manifest.controller.ts
```diff
@@ -0,0 +1,121 @@
...
+import {findRunnerByNetworkFamily, prepareDirPath, renderTemplate} from '../../../utils';
+import {TemplateKind} from '../../codegen-controller';
+import {graphToSubqlNetworkFamily, networkDsConverters, networkTemplateConverters} from '../constants';
+import {getChainIdByNetworkName} from '../migrate-controller';
+import {MigrateDatasourceKind, SubgraphDataSource, SubgraphProject, SubgraphTemplate, ChainInfo} from '../types';
+
+const PROJECT_TEMPLATE_PATH = path.resolve(__dirname, '../../../template/project.ts.ejs');
+
+export function readSubgraphManifest(inputPath: string): SubgraphProject {
+  if (!fs.existsSync(inputPath)) {
```
stwiname, line 19: I think rather than check this, we can just handle the `ENOENT` error when calling `readFileSync`

packages/cli/src/controller/migrate/manifest/migrate-manifest.controller.ts
```diff
@@ -0,0 +1,121 @@
...
+  const subgraphManifest: SubgraphProject = YAML.parse(fs.readFileSync(inputPath, 'utf8'));
+  return subgraphManifest;
+}
+
+export function extractNetworkFromManifest(subgraphProject: SubgraphProject): ChainInfo {
+  const firstDsKind = subgraphProject.dataSources[0].kind;
+  const firstDsNetwork = subgraphProject.dataSources[0].network;
+  if (!firstDsKind || !firstDsNetwork) {
+    throw new Error(`Subgraph dataSource kind or network not been found`);
+  }
```
stwiname, line 31: What happens if any ds > 0 is for a different network?

packages/cli/src/controller/migrate/manifest/migrate-manifest.controller.ts
```diff
@@ -0,0 +1,121 @@
...
+          projectHandlerKind: 'EthereumHandlerKind',
+        },
+        projectJson: project,
+      },
+      helper: {
+        upperFirst,
+      },
+    });
+  } catch (e) {
+    throw new Error(`When render project manifest having problems.`);
```
stwiname, line 61:
```suggestion
    throw new Error(`Failed to create project manifest.`);
```

packages/cli/src/controller/migrate/manifest/migrate-manifest.controller.ts
```diff
@@ -0,0 +1,121 @@
...
+          projectDatasourceKind: 'EthereumDatasourceKind',
+          projectHandlerKind: 'EthereumHandlerKind',
+        },
+        projectJson: project,
+      },
+      helper: {
+        upperFirst,
+      },
+    });
+  } catch (e) {
```
stwiname, line 60: We need a way we can expose this underlying error. If a user faces an error here, we have no way to debug it without making a build. 

We could set a log level or include the same error

packages/cli/src/controller/migrate/manifest/migrate-manifest.controller.ts
```diff
@@ -0,0 +1,121 @@
...
+    specVersion: '1.0.0',
+    runner: {
+      node: {version: '^', name: findRunnerByNetworkFamily(chainInfo.networkFamily)},
+      query: {version: '^', name: '@subql/query'},
+    },
+    version: subgraphManifest.specVersion,
+    dataSources: subgraphDsToSubqlDs(chainInfo.networkFamily, subgraphManifest.dataSources),
+    description: subgraphManifest.description,
+    schema: subgraphManifest.schema,
+    repository: subgraphManifest.repository,
```
stwiname, line 94: This would be different now right?

packages/cli/src/controller/migrate/manifest/migrate-manifest.controller.ts
```diff
@@ -0,0 +1,121 @@
...
+  };
+}
+
+export function subgraphTemplateToSubqlTemplate(
+  network: NETWORK_FAMILY,
+  subgraphTemplates: SubgraphTemplate[]
+): TemplateKind[] {
+  const convertFunction = networkTemplateConverters[network as NETWORK_FAMILY];
+  if (!convertFunction) {
+    throw new Error(`When migration, unable find convert template method for ${network} `);
```
stwiname, line 107:
```suggestion
    throw new Error(`${network} does not support migration of templates.`);
```

packages/cli/src/controller/migrate/manifest/migrate-manifest.controller.ts
```diff
@@ -0,0 +1,121 @@
...
+  return subgraphTemplates.map((t) => convertFunction(t));
+}
+
+export function subgraphDsToSubqlDs(
+  network: NETWORK_FAMILY,
+  subgraphDs: SubgraphDataSource[]
+): MigrateDatasourceKind[] {
+  const convertFunction = networkDsConverters[network];
+  if (!convertFunction) {
+    throw new Error(`When migration, unable find convert dataSource method for ${network} `);
```
stwiname, line 118: Is this the first place the user would face an error if they tried an unsupported network?
```suggestion
    throw new Error(`${network} is not supported with migrations`);
```

packages/cli/src/controller/migrate/constants.ts
```diff
@@ -0,0 +1,66 @@
+// Copyright 2020-2024 SubQuery Pte Ltd authors & contributors
+// SPDX-License-Identifier: GPL-3.0
+
+import {NETWORK_FAMILY} from '@subql/common';
+import {convertEthereumDs, convertEthereumTemplate} from './manifest/ethereum';
+import {DsConvertFunction, TemplateConvertFunction, NetworkExampleProject} from './types';
+
+// @ts-ignore
+export const networkDsConverters: Record<NETWORK_FAMILY, DsConvertFunction> = {
```
stwiname, line 9: I think its worth combining all these `*Converters` into one record, with the keys being a complete object. 

That way when implementing a new network family there's enforcement that everything is implemented and its easier to throw an error if a network is unsupported

stwiname (commented):
This is looking really good. 

I think we need some more error handling and making sure we don't drop things that the graph might support that we dont. Its worth checking over the graph docs to see the specs of their manifest.

I can already spot some things that are missing. We don't necessarily need to support them but we should throw errors if we come across them.
* Grafting
* `features`
* `dataSource.endBlocks`
* `dataSources.context`

packages/cli/src/controller/migrate/abis/migrate-abis.controller.ts
jiqiang90: I think we should allow this pass but log the error, so in this way cli will not remove the migrated project directory,  this allow user to manually fix/copy over the abi files.

packages/cli/src/controller/migrate/constants.ts
jiqiang90: No it refer to a specific network https://thegraph.com/docs/en/developing/supported-networks/
example from poap: 
https://github.com/poap-xyz/poap-subgraph/blob/aa640881ffd6ca5f50930ad32b78006f51e4a626/subgraph.template.yaml#L9
https://github.com/poap-xyz/poap-subgraph/blob/aa640881ffd6ca5f50930ad32b78006f51e4a626/config/goerli.json#L2

jiqiang90: But good we found https://thegraph.com/docs/_next/data/IlewALDCf3LFk3rL28g-h/en/developing/supported-networks.json

packages/cli/src/controller/migrate/manifest/migrate-manifest.controller.ts
jiqiang90: I don't think this legal in their manifest, I can add an assertion here too, we don't support this either

packages/cli/src/controller/migrate/manifest/migrate-manifest.controller.ts
```diff
@@ -0,0 +1,144 @@
...
+}
+
+export function subgraphTemplateToSubqlTemplate(
+  network: NETWORK_FAMILY,
+  subgraphTemplates: SubgraphTemplate[]
+): TemplateKind[] {
+  const convertFunction = networkConverters[network as NETWORK_FAMILY]?.templateConverter;
+  if (!convertFunction) {
+    throw new Error(`${network} does not support migration of templates.`);
+  }
```
stwiname, line 131: Checking the network exists can be moved to `graphManifestToSubqlManifest`. We don't need to check multiple times and `templateConverter` should be a requirement for each network family

packages/cli/src/controller/migrate/migrate-controller.ts
```diff
@@ -0,0 +1,81 @@
...
+import {ChainInfo, SubgraphProject} from './types';
+
+/**
+ *
+ * @param input is github link or ssh, or bitbucket link
+ * return undefined if link not match any
+ */
+const githubSshRegex = /^git@github\.com:(?<domain>[^/]+)\/(?<repository>[^/]+)\.git$/;
+const githubLinkRegex = /^https:\/\/github\.com\/(?<domain>[^/]+)\/(?<repository>[^/]+)(?:\/tree\/(?<branch>[^/]+))?/;
+const bitbucketSshRegex = /^git@bitbucket\.org:(?<domain>[^/]+)\/(?<repository>[^/]+)\.git$/;
```
stwiname, line 18: I think the ssh regex can be combined. We don't care about the domain

packages/cli/src/controller/migrate/schema/migrate-schema.controller.ts
```diff
@@ -0,0 +1,12 @@
...
+// SPDX-License-Identifier: GPL-3.0
+
+/**
+ *  Migrate a subgraph project schema to subquery schema file
+ * @param network network family
+ * @param inputPath file path to subgraph.yaml
+ * @param outputPath file path to project.ts
+ */
+export async function migrateSchema(inputPath: string, outputPath: string) {
+  //TODO
```
stwiname, line 11: Maybe here for now, we could log instructions as to what to do. Can have a link to some docs etc.

packages/cli/src/controller/migrate/manifest/ethereum.ts
```diff
@@ -0,0 +1,82 @@
...
+  subqlDs.startBlock = ds.source.startBlock;
+  subqlDs.endBlock = ds.source.endBlock;
+  subqlDs.options = {abi: ds.source.abi, address: ds.source.address};
+  return subqlDs;
+}
+
+export function convertEthereumTemplate(ds: SubgraphTemplate): MigrateDatasourceKind<EthTemplate> {
+  const subqlTemplate = baseDsConversion<EthTemplate>(ds);
+  subqlTemplate.options = {abi: ds.source.abi};
+  subqlTemplate.name = 'eth-template-name';
```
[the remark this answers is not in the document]
jiqiang90: my mistake, their template do have a name

packages/cli/src/controller/migrate/abis/migrate-abis.controller.ts
```diff
@@ -0,0 +1,38 @@
...
+    console.error('Error copying ABI files:', error);
+  }
+}
+
+export async function migrateAbis(
+  subgraphManifest: SubgraphProject,
+  subgraphDir: string,
+  subqlDir: string
+): Promise<void> {
+  const abiPaths = extractAllAbiFiles(subgraphManifest.dataSources);
```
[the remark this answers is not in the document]
jiqiang90: Good point, but I think the difficulty here is we can not determine the root directory that contains all abis files in their project, as there can be multiple directories. 
So far we support move all these abi files indicated in manifest first, I think we can log a message to notify users to manually move other abis as they needed.

packages/cli/src/commands/migrate.ts
```diff
@@ -0,0 +1,114 @@
...
+        await git().clone(link, subgraphDir, branch ? ['-b', branch, '--single-branch'] : ['--single-branch']);
+      }
+      console.log(
+        `* Pull subgraph project from git: ${link}, branch: ${branch ?? 'default branch'}${
+          gitSubDirectory ? `, subdirectory:${gitSubDirectory}` : '.'
+        }`
+      );
+    } else if (direMatch) {
+      if (gitSubDirectory) {
+        this.error(`Git sub directory is only work with git path, not local directory `);
```
stwiname, line 75:
```suggestion
        this.error(`Git sub directory only works with git path, not local directories.`);
```

packages/cli/src/controller/migrate/manifest/ethereum.ts
```diff
@@ -0,0 +1,82 @@
...
+              return {
+                kind: EthereumHandlerKind.Call,
+                migrateHandlerType: 'EthereumHandlerKind.Call',
+                handler: h.handler,
+                filter: {
+                  f: h.function,
+                },
+              };
+            })
+          : []),
```
stwiname, line 62:
```suggestion
        ...(ds.mapping.eventHandlers ?? []).map((h) => {
              return {
                kind: EthereumHandlerKind.Event,
                migrateHandlerType: 'EthereumHandlerKind.Event',
                handler: h.handler,
                filter: {
                  topics: [h.event],
                },
              };
            }),
        ...(ds.mapping.callHandlers ?? []).map((h) => {
              return {
                kind: EthereumHandlerKind.Call,
                migrateHandlerType: 'EthereumHandlerKind.Call',
                handler: h.handler,
                filter: {
                  f: h.function,
                },
              };
            }),
```

packages/cli/src/controller/migrate/manifest/migrate-manifest.controller.ts
```diff
@@ -0,0 +1,150 @@
...
+    throw new Error(`Subgraph dataSource kind or network not been found`);
+  }
+  if (!subgraphDsNetworks.every((network) => network === subgraphDsNetworks[0])) {
+    throw new Error(`All network values in subgraph Networks should be the same. Got ${subgraphDsNetworks}`);
+  }
+  const firstDsKind = subgraphDsKinds[0];
+  const firstDsNetwork = subgraphDsNetworks[0];
+  const networkFamily = graphToSubqlNetworkFamily[firstDsKind];
+  if (!networkFamily) {
+    throw new Error(`Corresponding subquery network is not find with subgraph data source kind ${firstDsKind}`);
```
stwiname, line 65:
```suggestion
    throw new Error(`Corresponding SubQuery network is not found with subgraph data source kind ${firstDsKind}`);
```

packages/cli/src/controller/migrate/types.ts
```diff
@@ -0,0 +1,85 @@
...
+
+type MigrateMappingType<T extends DatasourceKind | TemplateKind> = {
+  handlers: (T['mapping']['handlers'][number] & {migrateHandlerType: string})[];
+};
+export type MigrateDatasourceKind<T extends DatasourceKind | TemplateKind = DatasourceKind> = T & {
+  migrateDatasourceType: string;
+  mapping: T['mapping'] & MigrateMappingType<T>;
+};
+
+// TODO, currently use DatasourceKind, which migrate network supported,should be a new type include all network dataSources
```
stwiname, line 83: Is this still a todo?

stwiname (approved):

PR #2402, opened 2024-05-14
Monitor service

jiqiang90:
# Description

TODO:
- [x] add service to other place ,and write data
- [x] ideally this should add from main threat, but we need handle workers with this service
- [ ] part of Admin api 

Fixes # (issue)

## Type of change

## Checklist

- [ ] I have tested locally

### DIFF
--- a/packages/node-core/src/indexer/monitor.service.ts
+++ b/packages/node-core/src/indexer/monitor.service.ts
@@ -1,0 +1,460 @@
+// Copyright 2020-2024 SubQuery Pte Ltd authors & contributors
+// SPDX-License-Identifier: GPL-3.0
+
+import * as fs from 'fs';
+import path from 'path';
+import * as readline from 'readline';
+import {Injectable} from '@nestjs/common';
+import {NodeConfig} from '../configure';
+import {getLogger} from '../logger';
+import {setMonitorService} from '../process';
+
+const UNIT_MB = 1024 * 1024;
+
+const logger = getLogger('Monitor');
+
+enum FileLocation {
+  'A' = 'fileA.txt',
+  'B' = 'fileB.txt',
+}
+
+interface IndexEntry {
+  blockHeight: number;
+  startLine: number;
+  file: keyof typeof FileLocation; // 'A' or 'B' for fileAPath and fileBPath respectively
+  forked?: boolean;
+}
+
+interface IndexBlockEntry extends IndexEntry {
+  endLine: number;
+}
+
+interface FileStat {
+  fileSize: number;
+  fileEndLine: number;
+}
+
+export interface MonitorServiceInterface {
+  write(blockData: string): void;
+  createBlockFork(blockHeight: number): void;
+  createBlockStart(blockHeight: number): void;
+}
+
+@Injectable()
+export class MonitorService implements MonitorServiceInterface {
+  private readonly outputPath: string;
+  private readonly monitorFileSize: number;
+  private readonly indexPath: string;
+  private _currentFile: keyof typeof FileLocation | undefined;
+  private currentIndexHeight: number | undefined; // We need to keep this to update index when switch file, so we know current processing height
+  private _cachedFileStats: Record<keyof typeof FileLocation, FileStat> | undefined;
+
+  constructor(protected config: NodeConfig) {
+    this.outputPath = config.monitorOutDir;
+    this.monitorFileSize = config.monitorFileSize * UNIT_MB;
+    this.indexPath = path.join(this.outputPath, `${this.config.dbSchema}-index.csv`);
+    this.init();
+    setMonitorService(this);
+  }
+
+  static forceClean(monitorDir: string): void {
+    if (fs.existsSync(monitorDir)) {
+      fs.rmSync(monitorDir, {recursive: true, force: true});
+      logger.info('force cleaned monitor service files');
+    }
+  }
+
+  // maybe a good idea to expose this, if error didn't be caught api could still use reset
+  /**
+   * Reset all files and memory values
+   */
+  resetAll(): void {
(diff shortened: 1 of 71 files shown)


packages/node-core/src/indexer/monitor.service.ts
```diff
@@ -0,0 +1,382 @@
...
+    this.createOrResetFile(this.indexPath);
+    this._currentFile = this.getFilePath('A');
+    this._lastLine = 0;
+    this.currentFileSize = 0;
+  }
+
+  /**
+   * Init service, also validate
+   */
+  init(): void {
```
stwiname, line 66: Why not move this into the constructor?

packages/node-core/src/yargs.ts
```diff
@@ -233,6 +233,15 @@ export function yargsBuilder<
                   'Path to a directory for CSV output.  If this is not provided then data will not be output to CSV',
                 type: 'string',
               },
+              'monitor-out-dir': {
+                describe:
+                  'Path to a directory for monitor service output.  If this is not provided then data will store to default folder',
+                type: 'string',
+              },
+              'monitor-file-size': {
```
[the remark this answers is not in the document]
jiqiang90: updated, change to `MB`

packages/node-core/src/indexer/monitor.service.ts
```diff
@@ -0,0 +1,428 @@
...
+        input: fileStream,
+        crlfDelay: Infinity,
+      });
+
+      let currentLine = 0;
+
+      rl.on('line', (line) => {
+        if (indexEntry.endLine === undefined) {
+          throw new Error(`end line in indexEntry is expect to be defined`);
+        }
```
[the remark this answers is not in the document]
jiqiang90: i think this can be removed, endLine is always defined now

packages/node-core/src/utils/string.ts
```diff
@@ -0,0 +1,13 @@
...
+// SPDX-License-Identifier: GPL-3.0
+
+export function handledStringify(obj: any): string {
+  try {
+    return JSON.stringify(obj);
+  } catch (error) {
+    if (error instanceof Error) {
+      return error.stack || error.message;
+    }
+    return 'Unknown error when Stringify';
```
stwiname, line 11: Can we give a hint as to where this could have gone wrong? It will make things harder to debug otherwise

stwiname (commented):
A couple of suggestions:
* Make the montior service an optional dependency, this will clean up some tests
* Another option to simplify integration, would be to use event emitter to write, to the monitor service, that way we don't need to inject the service everywhere

packages/node-core/src/indexer/worker/worker.ts
```diff
@@ -195,6 +199,7 @@ export async function createIndexerWorker<
     {
       ...cacheHostFunctions(cache),
       ...storeHostFunctions(store),
+      ...(monitorService ? monitorHostFunctions(monitorService) : {}),
```
stwiname, line 202: It looks like WorkerMonitorService needs to handle with being an empty object

packages/node-core/src/subcommands/forceClean.service.ts
```diff
@@ -66,8 +68,12 @@ export class ForceCleanService {
           }
         );
       }
-
       logger.info('force cleaned schema and tables');
+      const monitorDir = this.nodeConfig.monitorOutDir ?? DEFAULT_MONITOR_STORE_PATH;
+      if (fs.existsSync(monitorDir)) {
+        fs.rmSync(monitorDir, {recursive: true, force: true});
+        logger.info('force cleaned monitor service files');
+      }
```
stwiname, line 76: I think this could be a static method on the monitor service

stwiname:
I think POI being enabled needs to set the size to non-zero. But we should determine the size based on the network project

packages/node-core/src/indexer/poi/poiSync.service.spec.ts
```diff
@@ -347,3 +348,73 @@ describe('Poi Service sync', () => {
...
+    const syncPoiJobSpy = jest.spyOn(poiSyncService as any, 'syncPoiJob');
+
+    // before start sync, latestSyncedPoi is 2
+    expect(poiSyncService.latestSyncedPoi.id).toBe(2);
+
+    // Start sync in a separate promise to simulate async behavior
+    const syncPromise = poiSyncService.syncPoi();
+
+    // We simulate let sync job keep going for a while
+    await delay(6);
```
stwiname, line 404: Could we reduce these delays so the test runs faster?

packages/node-core/src/utils/project.ts
```diff
@@ -24,12 +25,10 @@ export async function getValidPort(argvPort: number): Promise<number> {
...
-    logger.error(
-      `Unable to find available port (tried ports in range (${port}..${
-        port + 10
-      })). Try setting a free port manually by setting the --port flag`
-    );
-    process.exit(1);
+    const errMsg = `Unable to find available port (tried ports in range (${port}..${
+      port + 10
+    })). Try setting a free port manually by setting the --port flag`;
+    logger.error(errMsg);
```
stwiname, line 31: Should this use `exitWithError`?

tsconfig.json
```diff
@@ -1,7 +1,8 @@
 {
   "compilerOptions": {
     "module": "commonjs",
-    "target": "es2021",
+    "target": "ES2022",
+    "lib": ["ES2017", "ES2020", "ES2021","ES2022"],
```
stwiname, line 5: Do we need the earlier ones?

Can this just be `"lib": ["ES2022"],`?

stwiname (approved):

PR #2418, opened 2024-05-29
Update  block failed message 

### LINKED ISSUE #2414: Improve error message for pokladot failing to decode block
## Description

Currently when a block cannot be decoded the error message just says that it failed to index block without any indication how to solve the issue.

The error should be updated for substrate with the following message:
```
Failed to index block ${blockNum}. This is because the block cannot be decoded. To solve this you can either:
* Skip the block
* Update the chain types. You can test this by viewing the block with https://polkadot.js.org/apps/
For further information please read the docs: https://academy.subquery.network/
```

yoozo:
# Description
Update block failed message.

Fixes #2414 (issue)

## Type of change

## Checklist

- [ ] I have tested locally

### DIFF
--- a/packages/node/src/utils/substrate.test.ts
+++ b/packages/node/src/utils/substrate.test.ts
@@ -7,6 +7,7 @@ import {
   fetchBlocksArray,
   fetchBlocksBatches,
   filterExtrinsic,
+  getBlockByHeight,
 } from './substrate';
 
 const ENDPOINT_POLKADOT = 'wss://rpc.polkadot.io';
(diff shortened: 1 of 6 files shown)


stwiname (commented):
The updated error message should be for a specific error not all errors failing to index blocks. 

It needs to handle a specific error type here https://github.com/subquery/subql/blob/main/packages/node/src/utils/substrate.ts#L257

jiqiang90:
It will also face decode error at https://github.com/subquery/subql/blob/3e1d2ee7ea1ddce0c48b71d9d77b329121edb433/packages/node/src/utils/substrate.ts#L332

yoozo:
The comments have all been fixed.

stwiname (commented):
Looking much better, it needs to catch a specific type of error though, not just all errors from these methods.

@jiqiang90 can you please help finding a reproducible case if needed

jiqiang90:
> The updated error message should be for a specific error not all errors failing to index blocks.
> 
> It needs to handle a specific error type here https://github.com/subquery/subql/blob/main/packages/node/src/utils/substrate.ts#L257

I think from our previous experience, only these two calls when fetch block and get event will throw decode error, I'm happy with this change. 

@yoozo we can try this project , with change start height to 43249 and remove chainTypes. https://github.com/subquery/subql-starter/blob/main/Karura/karura-starter/project.ts

Will be nice we and include a unit test here too

jiqiang90 (approved):

packages/node/src/utils/substrate.test.ts
```diff
@@ -65,4 +66,22 @@ describe('substrate utils', () => {
       filterExtrinsic(block.extrinsics[2], { isSigned: false }),
     ).toBeFalsy();
   });
+
+  it('decode fail message', async () => {
+    const provider = new WsProvider(ENDPOINT_KARURA);
+    const api = await ApiPromise.create({ provider });
+
+    await expect(getBlockByHeight(api, 86614)).rejects.toThrow(
+      /Unable to decode|failed decoding|unknown type/,
```
stwiname, line 75: We want to test that it throws an error with our message and not the underlying error that we wrap

yoozo: I think both are needed.

stwiname (approved):

Issue #2387, opened 2024-05-05
[$300 in SQT] Query Service: Order by null order


stwiname:
When we order by ANY nullable field with DESC order, nulls are always on top.
That’s related to the [default postgres ordering params](https://www.postgresql.org/docs/current/queries-order.html), with NULLS FIRST set for DESC ordering.

Postgraphile has an untested example of a plugin for this https://www.graphile.org/postgraphile/make-add-pg-table-order-by-plugin/#makeaddpgtableorderbyplugin

Requirements:
* Backwards compatibility, any ordering queries should still work with any changes.
* Specifying the null order should be achieved by providing a new argument. eg `orderBy: [ENTITY_FIELD_ASC, NULL_OPTION] ` if it doesn't break compatibility otherwise `orderByNull: NULL_OPTION`

Faizalhidayat:
module.exports = function PgOrderNullsLastPlugin(builder) {
  builder.hook("GraphQLObjectType:fields", (fields, build, context) => {
    const {
      extend,
      pgIntrospectionResultsByKind: introspectionResultsByKind,
    } = build;
    const {
      scope: { isPgRowType, pgIntrospection: table },
    } = context;

    if (!isPgRowType || !table || table.kind !== "class") {
      return fields;
    }

    const newFields = introspectionResultsByKind.attribute
      .filter((attr) => attr.classId === table.id)
      .reduce((memo, attr) => {
        const ascFieldName = `${attr.name}_ASC_NULLS_LAST`;
        const descFieldName = `${attr.name}_DESC_NULLS_LAST`;

        return extend(memo, {
          [ascFieldName]: {
            description: `Orders by ${attr.name} in ascending order, nulls last.`,
            type: "Order",
          },
          [descFieldName]: {
            description: `Orders by ${attr.name} in descending order, nulls last.`,
            type: "Order",
          },
        });
      }, {});

    return extend(fields, newFields);
  });
};



This plugin adds two new ordering options for each field: FIELDNAME_ASC_NULLS_LAST and FIELDNAME_DESC_NULLS_LAST. These options will order by the field in ascending or descending order, respectively, with nulls last.

stwiname:
@Faizalhidayat please make a pull request, we don't accept comments with example code.

abhishek818:
@stwiname Is the reward/bounty still available?

stwiname:
> @stwiname Is the reward/bounty still available?

It sure is

PR #2421, opened 2024-06-02
Fix: Issue #2387 - Null Order Handling Improvement

Faizalhidayat:
### Summary:
This PR addresses the issue where null values are incorrectly ordered at the top when sorting by nullable fields in descending order. The proposed solution allows users to specify the null order while maintaining backward compatibility.

### Changes Made:
- Added a new argument for specifying null order in queries.
- Updated the relevant database queries to handle null ordering correctly.

### Testing:
- Tested the changes with various scenarios, including null values.
- Verified that existing queries continue to work as expected.

### Screenshots:
N/A (No visual changes)

### DIFF
--- a/packages/query/src/graphql/plugins/OrderByNullsPlugin.ts
+++ b/packages/query/src/graphql/plugins/OrderByNullsPlugin.ts
@@ -1,0 +1,36 @@
+import { makeAddPgTableOrderByPlugin, orderByAscDesc } from 'graphile-utils';
+import { Build, Plugin } from 'postgraphile';
+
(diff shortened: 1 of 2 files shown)


stwiname (changes requested):
Hi @Faizalhidayat 

This PR doesn't do as it says. The plugin you have created doesn't reflect on the DB schema, it uses hard coded values for schema, table and order. 

No projects are going to have a table called `MyTable` or properties `my_nullable_field` on that table.

I suggest you test your code works by using one of the starter proejcts and providing an example query we can use to test.

Please also use the PR template that is provided rather than your own format

PR #2430, opened 2024-06-06
db size for admin api

jiqiang90:
# Description

Fixes # (issue)

## Type of change

## Checklist

- [ ] I have tested locally

### DIFF
--- a/packages/node-core/src/db/sync-helper.ts
+++ b/packages/node-core/src/db/sync-helper.ts
@@ -276,15 +276,29 @@ export function getExistedIndexesQuery(schema: string): string {
   return `SELECT indexname FROM pg_indexes WHERE schemaname = '${schema}'`;
 }
 
-export async function getDbSizeQuery(sequelize: Sequelize, schema: string): Promise<number> {
-  const [size] = (await sequelize.query(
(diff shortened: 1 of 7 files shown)


packages/node-core/src/db/sync-helper.ts
```diff
@@ -276,6 +276,17 @@ export function getExistedIndexesQuery(schema: string): string {
   return `SELECT indexname FROM pg_indexes WHERE schemaname = '${schema}'`;
 }
 
+export async function getDbSizeQuery(sequelize: Sequelize, schema: string): Promise<number> {
```
stwiname, line 279: I think we can have a query that does everything in one and updates the metadata table if its not been update in a certain time

packages/query/src/graphql/plugins/GetMetadataPlugin.ts
```diff
@@ -33,6 +33,7 @@ const METADATA_TYPES = {
   deployments: 'string',
   lastCreatedPoiHeight: 'number',
   latestSyncedPoiHeight: 'number',
+  dbSize: 'number',
```
stwiname, line 36: I think this might be string because of bignumber type

stwiname (approved):

PR #2445, opened 2024-06-13
Strict utils types

yoozo:
# Description

Fixes # (issue)

enable ts strict mode in utils package

## Type of change

## Checklist

- [X] I have tested locally

### DIFF
--- a/packages/utils/src/graphql/entities.ts
+++ b/packages/utils/src/graphql/entities.ts
@@ -23,7 +23,7 @@ import {
 } from 'graphql';
 import {findDuplicateStringArray} from '../array';
 import {Logger} from '../logger';
-import {getTypeByScalarName} from '../types';
+import {TypeNames, getTypeByScalarName} from '../types';
 import {DirectiveName} from './constant';
 import {buildSchemaFromFile} from './schema';
 import {
@@ -84,7 +84,7 @@ export function getAllEntitiesRelations(_schema: GraphQLSchema | string | null):
   for (const entity of entities) {
     const newModel: GraphQLModelsType = {
       name: entity.name,
-      description: entity.description || undefined,
+      description: entity.description ?? undefined,
       fields: [],
       indexes: [],
     };
@@ -105,7 +105,7 @@ export function getAllEntitiesRelations(_schema: GraphQLSchema | string | null):
       else if (enums.has(typeString)) {
         newModel.fields.push({
           type: typeString,
-          description: field.description || undefined,
+          description: field.description ?? undefined,
           isEnum: true,
           isArray: isListType(isNonNullType(field.type) ? getNullableType(field.type) : field.type),
           nullable: !isNonNullType(field.type),
@@ -281,7 +281,7 @@ function packEntityField(
   return {
     name: isForeignKey ? `${field.name}Id` : field.name,
     type: isForeignKey ? FieldScalar.String : typeString,
-    description: field.description || undefined,
+    description: field.description ?? undefined,
     isArray: isListType(isNonNullType(field.type) ? getNullableType(field.type) : field.type),
     nullable: !isNonNullType(field.type),
     isEnum: false,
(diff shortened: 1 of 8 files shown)


packages/utils/src/types/TypeClass.ts
```diff
@@ -6,10 +6,22 @@ import {DataTypes} from '@subql/x-sequelize';
 type ValueOf<T> = T[keyof T];
 export type SequelizeTypes = string | ValueOf<typeof DataTypes>;
 
-export class TypeClass {
+export interface TypeToHashData {
```
stwiname, line 9: It would be nice to tie this to the type definitions. https://github.com/subquery/subql/tree/main/packages/utils/src/types/supported

If we add a new type in supported, e.g. BigFloat then this type no longer matches, there should be a compiler error but i don't think that would be the case

packages/utils/src/graphql/entities.ts
```diff
@@ -80,10 +80,11 @@ export function getAllEntitiesRelations(_schema: GraphQLSchema | string | null):
   const modelRelations = {models: [], relations: [], enums: [...enums.values()]} as GraphQLModelsRelationsEnums;
   const derivedFrom = schema.getDirective('derivedFrom');
   const indexDirective = schema.getDirective('index');
+  assert(derivedFrom && indexDirective, 'derivedFrom and index directives are required');
   for (const entity of entities) {
     const newModel: GraphQLModelsType = {
       name: entity.name,
-      description: entity.description,
+      description: entity.description || undefined,
```
stwiname, line 87: if description is "" or 0 it will be converted to undefined, its better to use `??`. This applies to the other cases as well

```suggestion
      description: entity.description ?? undefined,
```

packages/utils/src/graphql/entities.ts
```diff
@@ -311,15 +316,12 @@ export function setJsonObjectType(
     //check if field is also json
     const typeString = extractType(field.type);
     const isJsonType = jsonObjects.map((json) => json.name).includes(typeString);
+    const jsonObject = jsonObjects.find((object) => object.name === typeString);
+    assert(jsonObject, `Json object ${typeString} not found`);
```
stwiname, line 320: This is a change in behaviour, the `jsonInterface` property can now never be undefined. 

If it is required, then we can remove the `isJsonType` otherwise we should remove the assertion

packages/utils/src/types/TypeClass.ts
```diff
@@ -24,7 +24,7 @@ export class TypeClass {
     assert(this._sequelizeType !== undefined, `Type ${this.name} associated sequelize type is not supported`);
     return this._sequelizeType;
   }
-  hashCode(data: any): Uint8Array {
+  hashCode(data: D | any): Uint8Array {
```
stwiname, line 27: Can `| any` be removed?

packages/utils/src/types/generalTypes.spec.ts
```diff
@@ -9,12 +9,12 @@ import {wrappedNumToU8a} from './u8aUtils';
...
   it('can get json type', () => {
     const typeClass = getTypeByScalarName('Json');
-    expect(typeClass.name).toBe('Json');
+    expect(typeClass?.name).toBe('Json');
   });
 
   it('get sequelize date type', () => {
-    const typeClass = getTypeByScalarName('Date');
-    expect(typeClass.sequelizeType).toBe('timestamp');
+    const typeClass = getTypeByScalarName('DateObj');
```
stwiname, line 16: This test didn't fail before, whats the reason for this change?

packages/utils/src/graphql/entities.ts
yoozo: fix

stwiname (commented):
Code looks good, nice work. This is just missing a change log and can you please fill out the PR template

yoozo:
Modified

stwiname (approved):

PR #2447, opened 2024-06-13
enable ts strict in query package 

yoozo:
# Description

Fixes # (issue)

enable ts strict mode in query package

## Type of change

## Checklist

- [X] I have tested locally

### DIFF
--- a/packages/query/src/configure/configure.module.ts
+++ b/packages/query/src/configure/configure.module.ts
@@ -14,8 +14,8 @@ async function ensurePool(poolConfig: PoolConfig): Promise<Pool> {
   const pgPool = new Pool(poolConfig);
   try {
     await pgPool.connect();
-  } catch (e) {
-    if (JSON.stringify((e as Error).message).includes(CONNECTION_SSL_ERROR_REGEX)) {
+  } catch (e: any) {
+    if (JSON.stringify(e.message).includes(CONNECTION_SSL_ERROR_REGEX)) {
       poolConfig.ssl = undefined;
       return ensurePool(poolConfig);
     }
@@ -49,8 +49,8 @@ export class ConfigureModule {
           }
 
           return sslConfig;
-        } catch (e) {
-          getLogger('db config').error(e as Error);
+        } catch (e: any) {
+          getLogger('db config').error(e);
           throw e;
         }
       }
--- a/packages/query/src/configure/configure.module.ts
+++ b/packages/query/src/configure/configure.module.ts
@@ -15,7 +15,7 @@ async function ensurePool(poolConfig: PoolConfig): Promise<Pool> {
   try {
     await pgPool.connect();
   } catch (e) {
-    if (JSON.stringify(e.message).includes(CONNECTION_SSL_ERROR_REGEX)) {
+    if (JSON.stringify((e as Error).message).includes(CONNECTION_SSL_ERROR_REGEX)) {
       poolConfig.ssl = undefined;
       return ensurePool(poolConfig);
     }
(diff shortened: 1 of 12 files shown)


packages/query/src/configure/configure.module.ts
```diff
@@ -15,7 +15,7 @@ async function ensurePool(poolConfig: PoolConfig): Promise<Pool> {
   try {
     await pgPool.connect();
   } catch (e) {
-    if (JSON.stringify(e.message).includes(CONNECTION_SSL_ERROR_REGEX)) {
+    if (JSON.stringify((e as Error).message).includes(CONNECTION_SSL_ERROR_REGEX)) {
```
stwiname, line 18: The pattern we have been using is `} catch (e: any) {`, I think it would be good to retain that pattern

packages/query/src/configure/x-postgraphile/debugClient.ts
```diff
@@ -100,14 +100,17 @@ export function debugPgClient(pgClient: PoolClient, logger: Pino.Logger): PoolCl
...
-          let res: string;
-          res = `\n SQL query: ${query} `;
-          if (values && values.length !== 0) {
-            res = res.concat(` \n Values: ${JSON.stringify(values)}`);
-          }
-          logger.info(res);
-        });
+        // pgClient._explainResults = pgClient._explainResults || [];
+        if (pgClient._explainResults) {
+          pgClient._explainResults.forEach(({query, values}: {query: string; values?: any[]}) => {
```
stwiname, line 105: Could just use optional chaining here

packages/query/src/graphql/graphql.module.ts
```diff
@@ -41,8 +42,8 @@ const SCHEMA_RETRY_NUMBER = 5;
   providers: [ProjectService],
 })
 export class GraphqlModule implements OnModuleInit, OnModuleDestroy {
-  private apolloServer: ApolloServer;
-  private dbType: SUPPORT_DB;
+  private apolloServer!: ApolloServer;
+  private dbType!: SUPPORT_DB;
```
stwiname, line 46: Rather than adding `!` can you please use private getters/setters that have an assertion.

Here's an example https://github.com/subquery/subql/blob/main/packages/node-core/src/indexer/store.service.ts#L76-L78

packages/query/src/graphql/graphql.module.ts
```diff
@@ -133,9 +132,12 @@ export class GraphqlModule implements OnModuleInit, OnModuleDestroy {
       const pluginHook = makePluginHook([PgPubSub]);
       // Must be called manually to init PgPubSub since we're using Apollo Server and not postgraphile
       options = pluginHook('postgraphile:options', options, {pgPool: this.pgPool});
-      options.replaceAllPlugins.push(PgSubscriptionPlugin);
+      options.replaceAllPlugins = options.replaceAllPlugins || [];
```
stwiname, line 135: A newish feature is [??=](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_assignment)

```suggestion
      options.replaceAllPlugins ??= [];
```

packages/query/tsconfig.json
```diff
@@ -6,7 +6,8 @@
     "tsBuildInfoFile": "dist/.tsbuildinfo",
     "rootDir": "src",
     "outDir": "./dist",
-    "lib": ["dom"]
+    "lib": ["dom"],
```
stwiname, line 9: Can you please try removing this line and see if it produces any errors. The target for this isn't dom

stwiname (commented):
Looks good, same as the other PR, needs a changelog and PR template filled out

yoozo:
Modified

PR #2448, opened 2024-06-13
enable ts strict in common package 

yoozo:
# Description

Fixes # (issue)

Enable ts strict mode in common package

## Type of change

## Checklist

- [X] I have tested locally

### DIFF
--- a/packages/common/src/multichain/models.ts
+++ b/packages/common/src/multichain/models.ts
@@ -8,10 +8,10 @@ import {RunnerQueryBaseModel} from '../project';
 
 export class MultichainProjectManifestModel implements MultichainProjectManifest {
   @Equals('1.0.0')
-  specVersion: string;
+  specVersion!: string;
   @IsString({each: true})
-  projects: string[];
+  projects!: string[];
   @IsObject()
   @Type(() => RunnerQueryBaseModel)
-  query: QuerySpec;
+  query!: QuerySpec;
 }
--- a/packages/common/src/project/readers/github-reader.ts
+++ b/packages/common/src/project/readers/github-reader.ts
@@ -22,26 +22,23 @@ export class GithubReader implements Reader {
     return undefined;
   }
 
-  async getPkg(): Promise<IPackageJson | undefined> {
-    return this.getFile('package.json') as Promise<unknown> as Promise<IPackageJson | undefined>;
+  async getPkg(): Promise<IPackageJson> {
+    return this.getFile('package.json') as Promise<unknown> as Promise<IPackageJson>;
   }
 
-  async getProjectSchema(): Promise<unknown | undefined> {
+  async getProjectSchema(): Promise<unknown> {
     // Github reader not support ts manifest.
     const projectYaml = await this.getFile(DEFAULT_MANIFEST);
-    if (projectYaml === undefined) {
-      throw new Error('Fetch project from github got undefined');
-    }
     return yaml.load(projectYaml);
   }
 
-  async getFile(fileName: string): Promise<string | undefined> {
+  async getFile(fileName: string): Promise<string> {
     try {
       const branch = await this.getDefaultBranch();
       const {data} = await this.api.get(path.join(branch, fileName));
       return data;
     } catch (err) {
-      return undefined;
+      throw new Error('Fetch project from github got undefined');
     }
   }
 
--- a/packages/common/src/project/readers/github-reader.ts
+++ b/packages/common/src/project/readers/github-reader.ts
@@ -10,7 +10,7 @@ import {DEFAULT_MANIFEST} from '../utils';
 
 export class GithubReader implements Reader {
   private readonly api: AxiosInstance;
-  private defaultBranch!: string;
+  private defaultBranch?: string;
 
   constructor(private readonly key: string) {
     this.api = axios.create({
@@ -48,6 +48,6 @@ export class GithubReader implements Reader {
     }
     const {data} = await axios.get(`https://api.github.com/repos/${this.key}`);
     this.defaultBranch = data.default_branch;
-    return this.defaultBranch;
+    return this.defaultBranch as string;
   }
 }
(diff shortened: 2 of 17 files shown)


packages/common/src/project/readers/github-reader.ts
```diff
@@ -10,7 +10,7 @@ import {DEFAULT_MANIFEST} from '../utils';
 
 export class GithubReader implements Reader {
   private readonly api: AxiosInstance;
-  private defaultBranch: string;
+  private defaultBranch!: string;
```
stwiname, line 13: I think this can be optional

packages/common/src/multichain/models.ts
```diff
@@ -8,10 +8,10 @@ import {RunnerQueryBaseModel} from '../project';
...
   @Equals('1.0.0')
-  specVersion: string;
+  specVersion!: string;
   @IsString({each: true})
-  projects: string[];
+  projects!: string[];
   @IsObject()
   @Type(() => RunnerQueryBaseModel)
-  query: QuerySpec;
+  query!: QuerySpec;
```
stwiname, line 16: Can you please explain this, it doesn't make sense to me to need the `!`. Same applies to the other files with this change

packages/common/src/project/readers/local-reader.ts
```diff
@@ -9,14 +10,19 @@ import {IPackageJson} from 'package-json-type';
...
+  ) {}
 
   get root(): string {
     return path.resolve(this.projectPath);
   }
 
   async getPkg(): Promise<IPackageJson | undefined> {
-    return yaml.load(await this.getFile('package.json')) as IPackageJson;
+    const pkg = (await this.getFile('package.json')) || '';
+    assert(pkg, 'package.json not found');
```
stwiname, line 24: I think rather than assert here, we could update `getFile` to no longer return undefined and instead throw an error

packages/common/src/multichain/models.ts
yoozo: How should this be modified? The code only declares a class but doesn’t use it anywhere. Should I define a constructor for it?

packages/common/src/project/readers/local-reader.ts
yoozo:
```
  getProjectSchema(): Promise<unknown | undefined>;
  getPkg(): Promise<IPackageJson | undefined>;
  getFile(file: string): Promise<string | undefined>;
```
Should I modify Reader? Not returning undefined.

packages/common/src/multichain/models.ts
stwiname: This is used by class-transformer/class-validator so it will construct this class and validate it. It might be worth looking at their repos to see how others have solved this

packages/common/src/project/readers/local-reader.ts
stwiname: Yes i think so

packages/common/src/project/readers/github-reader.ts
yoozo: fix

packages/common/src/project/readers/local-reader.ts
yoozo: fix

packages/common/src/project/readers/github-reader.ts
```diff
@@ -22,34 +23,35 @@ export class GithubReader implements Reader {
...
 
-  async getFile(fileName: string): Promise<string | undefined> {
+  async getFile(fileName: string): Promise<string> {
     try {
       const branch = await this.getDefaultBranch();
       const {data} = await this.api.get(path.join(branch, fileName));
       return data;
     } catch (err) {
-      return undefined;
+      throw new Error('Fetch project from github got undefined');
```
stwiname, line 42:
```suggestion
      throw new Error('Failed to fetch file from github', { cause: err });
```

packages/common/src/project/readers/ipfs-reader.ts
```diff
@@ -28,28 +31,26 @@ export class IPFSReader implements Reader {
     return undefined;
   }
 
-  async getPkg(): Promise<IPackageJson | undefined> {
-    return Promise.resolve(undefined);
+  // eslint-disable-next-line @typescript-eslint/require-await
+  async getPkg(): Promise<IPackageJson> {
+    throw new Error('Project package.json is not found');
```
stwiname, line 36: This is a change in behaviour, we need to check that nowhere calls this
```suggestion
    throw new Error('IPFS Reader doesn't support reading package.json');
```

packages/common/src/project/readers/ipfs-reader.ts
```diff
@@ -28,28 +31,26 @@ export class IPFSReader implements Reader {
...
       if (this.cache[fileName] === undefined) {
         const resolvedFileName = fileName.replace('ipfs://', '');
         this.cache[fileName] = this.resultToBuffer(this.ipfs.cat(resolvedFileName));
       }
       return await this.cache[fileName];
     } catch (e) {
-      console.error(`Failed to fetch file from IPFS: ${fileName}`, e);
-      return undefined;
+      console.error(`Failed to fetch project from IPFS: ${fileName}`, e);
+      throw new Error(`Failed to fetch project from IPFS: ${fileName}`);
```
stwiname, line 53: We have started providing the cause to all errors where we can. It would be worth reading this https://medium.com/with-orus/the-5-commandments-of-clean-error-handling-in-typescript-93a9cbdf1af5

```suggestion
      throw new Error(`Failed to fetch project from IPFS: ${fileName}`, { cause: e });
```

packages/common/src/project/readers/reader.ts
```diff
@@ -1,6 +1,7 @@
 // Copyright 2020-2024 SubQuery Pte Ltd authors & contributors
 // SPDX-License-Identifier: GPL-3.0
 
+import assert from 'assert';
```
stwiname, line 4: Unused

stwiname (approved):

PR #2450, opened 2024-06-14
enable ts strict in node 

yoozo:
# Description

Fixes # (issue)
Enable ts strict in node

## Type of change

## Checklist

- [x] I have tested locally

### DIFF
--- a/packages/node-core/src/utils/project.ts
+++ b/packages/node-core/src/utils/project.ts
@@ -17,7 +17,7 @@ import {exitWithError} from '../process';
 
 const logger = getLogger('Project-Utils');
 
-export async function getValidPort(argvPort: number): Promise<number> {
+export async function getValidPort(argvPort: number | undefined): Promise<number> {
   const validate = (x: any) => {
     const p = parseInt(x);
     return isNaN(p) ? null : p;
--- a/packages/node-core/src/utils/project.ts
+++ b/packages/node-core/src/utils/project.ts
@@ -5,7 +5,14 @@ import fs from 'fs';
 import os from 'os';
 import path from 'path';
 import {DEFAULT_PORT, findAvailablePort, GithubReader, IPFSReader, LocalReader} from '@subql/common';
-import {BaseAssetsDataSource, BaseCustomDataSource, BaseDataSource, Reader, TemplateBase} from '@subql/types-core';
+import {
+  BaseAssetsDataSource,
+  BaseCustomDataSource,
+  BaseDataSource,
+  BaseTemplateDataSource,
+  Reader,
+  TemplateBase,
+} from '@subql/types-core';
 import {getAllEntitiesRelations} from '@subql/utils';
 import {QueryTypes, Sequelize} from '@subql/x-sequelize';
 import {stringToArray, getSchedule} from 'cron-converter';
@@ -278,20 +285,23 @@ export async function insertBlockFiltersCronSchedules<DS extends BaseDataSource 
   return dataSources;
 }
 
-export async function loadProjectTemplates<T extends BaseDataSource & TemplateBase>(
+export async function loadProjectTemplates<T extends BaseTemplateDataSource>(
   templates: T[] | undefined,
   root: string,
   reader: Reader,
-  isCustomDs: IsCustomDs<T, T & BaseCustomDataSource>
+  isCustomDs: IsCustomDs<T, Omit<T & BaseCustomDataSource, keyof TemplateBase>>
 ): Promise<T[]> {
   if (!templates || !templates.length) {
     return [];
   }
-  const dsTemplates = await updateDataSourcesV1_0_0(templates, reader, root, isCustomDs);
+
+  const templateIsCustomDs = (template: T): template is T & BaseCustomDataSource =>
+    isCustomDs(template) && 'name' in template;
+  const dsTemplates = await updateDataSourcesV1_0_0(templates, reader, root, templateIsCustomDs);
   return dsTemplates.map((ds, index) => ({
     ...ds,
     name: templates[index].name,
-  })) as T[];
+  }));
 }
 
 export function getStartHeight(dataSources: BaseDataSource[]): number {
--- a/packages/node-core/src/utils/project.ts
+++ b/packages/node-core/src/utils/project.ts
@@ -229,13 +229,13 @@ export async function initHotSchemaReload(schema: string, storeService: StoreSer
   await storeService.initHotSchemaReloadQueries(schema);
 }
 
-type IsRuntimeDs = (ds: BaseDataSource) => boolean;
+type IsRuntimeDs<DS> = (ds: DS) => ds is DS;
 
 // eslint-disable-next-line @typescript-eslint/require-await
 export async function insertBlockFiltersCronSchedules<DS extends BaseDataSource = BaseDataSource>(
   dataSources: DS[],
   getBlockTimestamp: (height: number) => Promise<Date>,
-  isRuntimeDs: IsRuntimeDs,
+  isRuntimeDs: IsRuntimeDs<DS>,
   blockHandlerKind: string
 ): Promise<DS[]> {
   dataSources = await Promise.all(
@@ -282,7 +282,7 @@ export async function loadProjectTemplates<T extends BaseDataSource & TemplateBa
   templates: T[] | undefined,
   root: string,
   reader: Reader,
-  isCustomDs: IsCustomDs<BaseDataSource, BaseCustomDataSource>
+  isCustomDs: IsCustomDs<T, T & BaseCustomDataSource>
 ): Promise<T[]> {
   if (!templates || !templates.length) {
     return [];
@@ -291,7 +291,7 @@ export async function loadProjectTemplates<T extends BaseDataSource & TemplateBa
   return dsTemplates.map((ds, index) => ({
     ...ds,
     name: templates[index].name,
-  })) as T[]; // How to get rid of cast here?
+  })) as T[];
 }
 
 export function getStartHeight(dataSources: BaseDataSource[]): number {
--- a/packages/node-core/src/utils/project.ts
+++ b/packages/node-core/src/utils/project.ts
@@ -17,7 +17,7 @@ import {exitWithError} from '../process';
 
 const logger = getLogger('Project-Utils');
 
-export async function getValidPort(argvPort: number | undefined): Promise<number> {
+export async function getValidPort(argvPort?: number): Promise<number> {
   const validate = (x: any) => {
     const p = parseInt(x);
     return isNaN(p) ? null : p;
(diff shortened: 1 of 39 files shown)


packages/node/src/indexer/worker/worker.service.ts
```diff
@@ -67,9 +68,10 @@ export class WorkerService extends BaseWorkerService<
     block: IBlock<BlockContent | LightBlockContent>,
     dataSources: SubstrateDatasource[],
   ): Promise<ProcessBlockResponse> {
-    const runtimeVersion = !isFullBlock(block.block)
-      ? undefined
-      : await this.workerRuntimeService.getRuntimeVersion(block.block.block);
+    assert(isFullBlock(block.block), 'Block should be Full block');
+    const runtimeVersion = await this.workerRuntimeService.getRuntimeVersion(
+      block.block.block,
+    );
```
stwiname, line 74: This is also a change in behaviour

packages/node/src/utils/project.ts
```diff
@@ -42,14 +43,16 @@ export async function getChainTypes(
...
-    return loadChainTypes(file, root);
+    return loadChainTypes(file, root) as ChainTypes;
   } else {
     // If it is stored in ipfs or other resources, we will use the corresponding reader to read the file
     // Because ipfs not provide extension of the file, it is difficult to determine its format
     // We will use yaml.load to try to load the script and parse them to supported chain types
     // if it failed, we will give it another another attempt, and assume the script written in js
     // we will download it to a temp folder, and load them within sandbox
     const res = await reader.getFile(file);
+    assert(res, `File ${file} not found`);
```
stwiname, line 54: I think `getFile` should be updated to throw rather than return undefined

packages/node/src/utils/substrate.ts
```diff
@@ -375,12 +377,18 @@ export async function fetchBlocksBatches(
...
-        ? overallSpecVer
-        : runtimeVersions[idx].specVersion.toNumber();
+    let parentSpecVersion: number | undefined;
+    if (overallSpecVer !== undefined) {
+      parentSpecVersion = overallSpecVer;
+    } else {
+      parentSpecVersion = runtimeVersions
+        ? runtimeVersions[idx].specVersion.toNumber()
+        : undefined;
+    }
```
stwiname, line 390: I think this could be reverted and optional chaining added.

```suggestion
const parentSpecVersion =
      overallSpecVer !== undefined
        ? overallSpecVer
        : runtimeVersions?.[idx].specVersion.toNumber();
```

packages/node/src/indexer/api.service.ts
```diff
@@ -154,7 +155,15 @@ export class ApiService
           chainTypes,
         }),
       //postConnectedHook
-      (connection: ApiPromiseConnection, endpoint: string, index: number) => {
+      (
+        connection: IApiConnectionSpecific<
+          ApiPromise,
+          ApiAt,
+          IBlock<BlockContent>[] | IBlock<LightBlockContent>[]
+        >,
```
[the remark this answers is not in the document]
yoozo: fix

packages/node/src/configure/SubqueryProject.ts
```diff
@@ -146,14 +146,15 @@ async function loadProjectFromManifestBase(
...
-  let schemaString: string;
+  let schemaString: string | undefined;
   try {
     schemaString = await reader.getFile(projectManifest.schema.file);
   } catch (e) {
     throw new Error(
       `unable to fetch the schema from ${projectManifest.schema.file}`,
     );
   }
+  assert(schemaString, 'Schema file is empty');
```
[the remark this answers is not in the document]
yoozo: Here, we need to merge the common PR first and then make adjustments.

packages/node-core/src/utils/project.ts
```diff
@@ -17,7 +17,7 @@ import {exitWithError} from '../process';
 
 const logger = getLogger('Project-Utils');
 
-export async function getValidPort(argvPort: number): Promise<number> {
+export async function getValidPort(argvPort: number | undefined): Promise<number> {
```
stwiname, line 20:
```suggestion
export async function getValidPort(argvPort?: number): Promise<number> {
```

packages/node/src/configure/SchemaMigration.service.test.ts
```diff
@@ -58,11 +58,11 @@ describe('SchemaMigration integration tests', () => {
     await apiService.init();
     await projectService.init(1);
 
-    const dbResults = await sequelize.query(
+    const dbResults: string[][] = await sequelize.query(
```
stwiname, line 61: Can we use generics wtih sequelize.query?

packages/node/src/configure/SubqueryProject.ts
```diff
@@ -72,7 +72,7 @@ export class SubqueryProject implements ISubqueryProject {
     this.#dataSources = await insertBlockFiltersCronSchedules(
       this.dataSources,
       getTimestamp,
-      isRuntimeDs,
+      isRuntimeDs, // Note: ts-strict change --- Not sure it should be modified
```
stwiname, line 75: Theres no change here, can we remove this comment?

packages/node/src/indexer/worker/worker.ts
```diff
@@ -33,7 +33,7 @@ import { WorkerService } from './worker.service';
 
 const logger = getLogger(`worker #${threadId}`);
 
-async function initWorker(startHeight: number): Promise<void> {
+async function initWorker(startHeight?: number): Promise<void> {
```
stwiname, line 36: How come this is now optional? I think its required

packages/node/src/utils/project.ts
```diff
@@ -43,14 +44,17 @@ export async function getChainTypes(
   // If the project is load from local, we will direct load them
   let raw: unknown;
   if (reader instanceof LocalReader) {
-    raw = loadChainTypes(file, root);
+    return loadChainTypes(file, root) as ChainTypes;
```
stwiname, line 47: Can you please rebase, there was a fix to this code recently

packages/node/tsconfig.json
```diff
@@ -4,7 +4,8 @@
     "sourceMap": true,
     "tsBuildInfoFile": "dist/.tsbuildinfo",
     "rootDir": "src",
-    "outDir": "./dist"
+    "outDir": "./dist",
+    "strict": true
```
stwiname, line 8: Because this is the last package in the repo to enable strict, we can set it on the base tsconfig and remove the explicit option on each of the packages

packages/node/src/indexer/dictionary/substrateDictionary.service.ts
```diff
@@ -101,3 +104,26 @@ export class SubstrateDictionaryService extends DictionaryService<
     return dict.getSpecVersions();
   }
 }
+
+function convertGetDsProcessor(
```
stwiname, line 108: This is a functional change and should have a test.

I think we will end up implementing this over for each SDK which isn't ideal. 

Could this not be fixed by improvements to the dictionary types? E.g. doing something like `typeof DsProcessorService["getDsProcessor"]`

packages/node/src/indexer/dictionary/v1/substrateDictionaryV1.ts
```diff
@@ -62,7 +65,7 @@ function getBaseHandlerKind<
   ds: SubstrateDataSource,
   handler: SubstrateHandler,
   getDsProcessor: (ds: SubstrateDatasource) => P,
-): SubstrateHandlerKind {
+): SubstrateHandlerKind | undefined {
```
stwiname, line 68: I think this could throw instead of being undefined. Then if we added another type we would be notified about the failure rather than silently not handling code.

Alternatively the switch that consumes the result of this function could also throw

packages/node/src/indexer/runtime/runtime.service.spec.ts
```diff
@@ -41,21 +42,22 @@ const getApiService = (): ApiService =>
...
-            const { id } = specVersions.find(
+            const specVersion = specVersions.find(
               (v) => v.start <= blockNum && (!v.end || v.end >= blockNum),
             );
+            assert(specVersion, `Spec version not found for block ${blockNum}`);
 
             return Promise.resolve({
               specVersion: {
-                toNumber: () => parseInt(id),
+                toNumber: () => parseInt(specVersion.id),
```
stwiname, line 52: Not related to the PR but this can lead to issues if not provided. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt#radix

```suggestion
                toNumber: () => parseInt(specVersion.id, 10),
```

packages/types/src/interfaces.ts
```diff
@@ -8,7 +8,7 @@ import {IEvent} from '@polkadot/types/types';
 
 export interface SubstrateBlock extends SignedBlock {
   // parent block's spec version, can be used to decide the correct metadata that should be used for this block.
-  specVersion: number;
+  specVersion?: number;
```
stwiname, line 11: I think this is required. @jiqiang90 can you confirm

packages/node/src/indexer/store.service.test.ts
```diff
@@ -77,8 +77,8 @@ describe('Store service integration test', () => {
         replacements: { schema: schemaName },
       },
     );
-    const expectedTables = result.map(
-      (t: { table_name: string }) => t.table_name,
+    const expectedTables = (result as { table_name: string }[]).map(
```
stwiname, line 80: I think its better to use generics on `sequelize.query`

packages/node/src/indexer/store.service.test.ts
```diff
@@ -278,7 +278,7 @@ WHERE
 
     tempDir = (projectService as any).project.root;
 
-    const result = await sequelize.query(
+    const result: { enum_type: string }[] = await sequelize.query(
```
stwiname, line 281: Same here for generics

packages/node/src/utils/project.ts
```diff
@@ -51,6 +52,7 @@ export async function getChainTypes(
     // if it failed, we will give it another attempt, and assume the script written in js
     // we will download it to a temp folder, and load them within sandbox
     const res = await reader.getFile(file);
+    let raw: unknown;
```
stwiname, line 55: Raw is already defined on line 45

packages/node/src/indexer/store.service.test.ts
```diff
@@ -64,7 +64,7 @@ describe('Store service integration test', () => {
 
     tempDir = (projectService as any).project.root;
 
-    const [result] = await sequelize.query(
+    const result = await sequelize.query<{ table_name: string }>(
```
stwiname, line 67: Is this right? I thought it would be 
```suggestion
    const [result] = await sequelize.query<[{ table_name: string }]>(
```

packages/node/src/utils/project.ts
```diff
@@ -1,6 +1,7 @@
 // Copyright 2020-2024 SubQuery Pte Ltd authors & contributors
 // SPDX-License-Identifier: GPL-3.0
 
+import assert from 'assert';
```
stwiname, line 4: Unused

stwiname (approved):

PR #2451, opened 2024-06-16
fix missing validate chainType logic

jiqiang90:
# Description
Fix not validating chainTypes when loading project from local files

Fixes # (issue)

## Type of change

## Checklist

- [ ] I have tested locally


stwiname (approved):

packages/node/src/utils/project.ts
```diff
@@ -42,7 +42,8 @@ export async function getChainTypes(
 ): Promise<ChainTypes> {
   // If the project is load from local, we will direct load them
   if (reader instanceof LocalReader) {
-    return loadChainTypes(file, root);
+    const raw = loadChainTypes(file, root);
+    return parseChainTypes(raw);
```
stwiname, line 46: It seems every path for returning does `return parseChainTypes`. Reworking this function so there is a single return would reduce chances of the issue happening again

PR #2452, opened 2024-06-16
Wait new block when last process height is same as last chain height

jiqiang90:
# Description

Fixes # (issue)

## Type of change

## Checklist

- [ ] I have tested locally

### DIFF
--- a/packages/node-core/src/indexer/fetch.service.ts
+++ b/packages/node-core/src/indexer/fetch.service.ts
@@ -110,7 +110,7 @@ export abstract class BaseFetchService<DS extends BaseDataSource, B extends IBlo
       // Alternative, we can expose async function getLastProcessedHeight() to ensure accuracy.
       if (startHeight - 1 === this.latestHeight()) {
         logger.warn(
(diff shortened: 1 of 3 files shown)


packages/node-core/src/indexer/fetch.service.ts
```diff
@@ -102,10 +102,21 @@ export abstract class BaseFetchService<DS extends BaseDataSource, B extends IBlo
...
     await Promise.all([this.getFinalizedBlockHead(), this.getBestBlockHead()]);
 
+    this.projectService.startHeight;
+
     if (startHeight > this.latestHeight()) {
-      throw new Error(
-        `The startBlock of dataSources in your project manifest (${startHeight}) is higher than the current chain height (${this.latestHeight()}). Please adjust your startBlock to be less that the current chain height.`
-      );
+      // This is at init stage, lastProcessedHeight should be always - 1 from the startHeight in this case
+      // this is reverse calculated from projectService.nextProcessHeight()
```
stwiname, line 109: It seems like the source of startHeight needs to be fixed rather than adding special cases here

stwiname (approved):

Issue #1963, opened 2023-08-21
Skip dictionary when difference between lastProcessedHeight and latestHeight is small


ianhe8x:
### Prerequisites

* [ ] Are you running the latest version(s)?
* [ ] Have you searched the [documentation for your issue](https://academy.subquery.network/)?
* [ ] Did you search for an existing issue or pull request?

### Description
We want to reduce unnecessary dict queries

### Details

Skip dictionary when difference between lastProcessedHeight and latestHeight is small. (mandatory value or use batch size)

**Local Environment:** [You can get this information from executing `subql version`.]
**Query Version:** [What is the version of the query service?]
**Indexer Version:** [What is the version of the indexer service?]
**Network Details:**
  * [Network]
  * [Block height, a block height where the issue is happening]
  * [Dictionary endpoint, if used]

### Steps to Reproduce

1. [First Step]
2. [Second Step]
3. [and so on...]

**Example project:** [A link to a minimal example that can reproduce the issue]

**Expected behavior:** [What you expected to happen]

**Actual behavior:** [What actually happened]

### Any other information

Is there any other information you would like to add?

stwiname:
This has been updated to match the last block, but we can be even looser and stop using the dictionary if its within a certain number of blocks

PR #2454, opened 2024-06-17
Skip dictionary if within batch size of latest height

stwiname:
# Description
Skips using the dictionary if project nearly fully indexed and within a batch size of the latest height

Fixes https://github.com/subquery/subql/issues/1963

## Type of change

## Checklist

- [x] I have tested locally

### DIFF
--- a/packages/node-core/src/indexer/fetch.service.spec.ts
+++ b/packages/node-core/src/indexer/fetch.service.spec.ts
@@ -655,8 +655,8 @@ describe('Fetch Service', () => {
   });
 
   it('throws if the start block is greater than the chain latest height', async () => {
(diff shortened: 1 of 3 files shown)


packages/node-core/src/indexer/fetch.service.spec.ts
```diff
@@ -655,8 +655,8 @@ describe('Fetch Service', () => {
   });
 
   it('throws if the start block is greater than the chain latest height', async () => {
```
jiqiang90, line 657:
```suggestion
  it('throws if the start block is greater than the chain latest height, and last processed height is not same as chain latest height', async () => {
```

jiqiang90 (approved):

PR #2465, opened 2024-06-21
Strict cli

yoozo:
# Description

Fixes # (issue)
Enable ts strict in cli

## Type of change

## Checklist

- [x] I have tested locally

### DIFF
--- a/packages/cli/src/commands/deployment/deploy.ts
+++ b/packages/cli/src/commands/deployment/deploy.ts
@@ -90,6 +90,7 @@ export default class Deploy extends Command {
       }
     }
     if (!flags.queryVersion) {
+      assert(validator.manifestRunner, 'Please set manifestRunner in your project');
       try {
         const queryVersions = await imageVersions(
           validator.manifestRunner.query.name,
--- a/packages/cli/src/commands/deployment/deploy.ts
+++ b/packages/cli/src/commands/deployment/deploy.ts
@@ -98,7 +98,13 @@ export default class Deploy extends Command {
           ROOT_API_URL_PROD
         );
         if (!flags.useDefaults) {
-          const response = await promptWithDefaultValues(inquirer, 'Enter query version', null, queryVersions, true);
+          const response = await promptWithDefaultValues(
+            inquirer,
+            'Enter query version',
+            undefined,
+            queryVersions,
+            true
+          );
           flags.queryVersion = response;
         } else {
           flags.queryVersion = queryVersions[0];
--- a/packages/cli/src/commands/deployment/deploy.ts
+++ b/packages/cli/src/commands/deployment/deploy.ts
@@ -63,12 +63,10 @@ export default class Deploy extends Command {
       } else {
         flags.dict = validateDictEndpoint;
       }
-
-      assert(flags.dict, 'Dictionary is required');
     }
 
-    assert(validator.manifestRunner, 'Please set manifestRunner in your project');
     if (!flags.indexerVersion) {
+      assert(validator.manifestRunner, 'Please set manifestRunner in your project');
       try {
         const indexerVersions = await imageVersions(
           validator.manifestRunner.node.name,
--- a/packages/cli/src/commands/deployment/deploy.ts
+++ b/packages/cli/src/commands/deployment/deploy.ts
@@ -63,6 +63,8 @@ export default class Deploy extends Command {
       } else {
         flags.dict = validateDictEndpoint;
       }
+
+      assert(flags.dict, 'Dictionary is required');
     }
 
     assert(validator.manifestRunner, 'Please set manifestRunner in your project');
--- a/packages/cli/src/commands/deployment/deploy.ts
+++ b/packages/cli/src/commands/deployment/deploy.ts
@@ -1,6 +1,7 @@
 // Copyright 2020-2024 SubQuery Pte Ltd authors & contributors
 // SPDX-License-Identifier: GPL-3.0
 
+import assert from 'assert';
 import {Command, Flags} from '@oclif/core';
 import chalk from 'chalk';
 import cli from 'cli-ux';
@@ -51,18 +52,20 @@ export default class Deploy extends Command {
         throw new Error(chalk.red('Please ensure a valid is passed using --endpoint flag'));
       }
 
-      flags.endpoint = await promptWithDefaultValues(cli, 'Enter endpoint', undefined, null, true);
+      flags.endpoint = await promptWithDefaultValues(cli, 'Enter endpoint', undefined, undefined, true);
     }
 
     if (!flags.dict) {
+      assert(validator.chainId, 'Please set chainId in your project');
       const validateDictEndpoint = processEndpoints(await dictionaryEndpoints(ROOT_API_URL_PROD), validator.chainId);
       if (!flags.useDefaults && !validateDictEndpoint) {
-        flags.dict = await promptWithDefaultValues(cli, 'Enter dictionary', validateDictEndpoint, null, false);
+        flags.dict = await promptWithDefaultValues(cli, 'Enter dictionary', validateDictEndpoint, undefined, false);
       } else {
         flags.dict = validateDictEndpoint;
       }
     }
 
+    assert(validator.manifestRunner, 'Please set manifestRunner in your project');
     if (!flags.indexerVersion) {
       try {
         const indexerVersions = await imageVersions(
@@ -75,7 +78,7 @@ export default class Deploy extends Command {
           flags.indexerVersion = await promptWithDefaultValues(
             inquirer,
             'Enter indexer version',
-            null,
+            undefined,
             indexerVersions,
             true
           );
--- a/packages/cli/src/commands/init.ts
+++ b/packages/cli/src/commands/init.ts
@@ -1,6 +1,7 @@
 // Copyright 2020-2024 SubQuery Pte Ltd authors & contributors
 // SPDX-License-Identifier: GPL-3.0
 
+import assert from 'assert';
 import fs from 'fs';
 import path from 'path';
 import {URL} from 'url';
@@ -43,12 +44,12 @@ function filterInput(arr: string[]) {
 
 async function promptValidRemoteAndBranch(): Promise<string[]> {
   let isValid = false;
-  let remote: string;
+  let remote: string | undefined;
   while (!isValid) {
     try {
-      remote = await cli.prompt('Custom template git remote', {
+      remote = (await cli.prompt('Custom template git remote', {
         required: true,
-      });
+      })) as string;
       new URL(remote);
       isValid = true;
     } catch (e) {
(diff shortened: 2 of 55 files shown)


packages/cli/src/commands/init.ts
```diff
@@ -78,11 +79,11 @@ export default class Init extends Command {
...
-  private projectPath: string; //path on GitHub
-  private project: ProjectSpecBase;
-  private location: string;
-  private networkFamily: NETWORK_FAMILY;
-  private network: string;
+  private projectPath!: string; //path on GitHub
+  private project!: ProjectSpecBase;
+  private location!: string;
+  private networkFamily!: NETWORK_FAMILY;
+  private network!: string;
```
stwiname, line 86: I don't think we need some of these private variables. 

`networkFamily` and `network` are only used within `run`
`location` can be passed as an argument to `cloneCustomRepo`
I think similar changes could be made for `project` and `projectPath` too

packages/cli/src/commands/migrate.ts
```diff
@@ -94,9 +96,9 @@ export default class Migrate extends Command {
       await migrateManifest(chainInfo, subgraphManifest, subqlManifestPath);
       // render package.json
       await preparePackage(subqlDir, {
-        name: subgraphManifest.name,
+        name: subgraphManifest.name || '',
```
stwiname, line 99:
```suggestion
        name: subgraphManifest.name ?? '',
```

packages/cli/src/commands/migrate.ts
```diff
@@ -94,9 +96,9 @@ export default class Migrate extends Command {
...
         description: subgraphManifest.description,
-        author: subgraphManifest.author,
+        author: subgraphManifest.author || '',
```
stwiname, line 101:
```suggestion
        author: subgraphManifest.author ?? '',
```

packages/cli/src/controller/build-controller.ts
```diff
@@ -74,15 +74,15 @@ export async function runWebpack(
 
   await new Promise((resolve, reject) => {
     webpack(config).run((error, stats) => {
-      if (error) {
+      if (error || !stats) {
```
stwiname, line 77: I think this needs to be separated. If !stats then error could also be undefined so it would reject with undefined.

packages/cli/src/controller/build-controller.ts
```diff
@@ -74,15 +74,15 @@ export async function runWebpack(
...
         reject(error);
         return;
       }
 
       if (stats.hasErrors()) {
         const info = stats.toJson();
 
-        reject(info.errors[0].message);
+        reject(info.errors?.map((e) => e.message).join('\n'));
```
stwiname, line 85:
```suggestion
        reject(info.errors?.map((e) => e.message).join('\n') ?? 'Unknown error');
```

packages/cli/src/controller/deploy-controller.spec.ts
```diff
@@ -39,6 +41,7 @@ async function deployTestProject(
   testAuth: string,
   url: string
 ): Promise<DeploymentDataType> {
+  assert(validator.manifestRunner, 'Please set manifestRunner in your project');
```
stwiname, line 44: I think in tests we can have non-null assertions, everything is in our control

packages/cli/src/controller/deploy-controller.ts
```diff
@@ -57,7 +58,7 @@ export async function createDeployment(
     );
     return res.data.deployment;
   } catch (e) {
-    errorHandle(e, 'Error deploying to hosted service:');
+    throw errorHandle(e, 'Error deploying to hosted service:');
```
stwiname, line 61: `errorHandle` says the return type is error but it just throws. Do you think we could update it so it returns errors rather than throwing?

packages/cli/src/controller/deploy-controller.ts
```diff
@@ -152,12 +155,17 @@ export async function updateDeployment(
...
 }
 
-export async function ipfsCID_validate(cid: string, authToken: string, url: string): Promise<ValidateDataType> {
+export async function ipfsCID_validate(
+  cid: string | undefined,
+  authToken: string,
+  url: string
+): Promise<ValidateDataType> {
   try {
+    assert;
```
stwiname, line 168: Should be asserting the cid?

packages/cli/src/controller/generate-controller.spec.ts
```diff
@@ -287,7 +287,7 @@ describe('CLI codegen:generate', () => {
   });
   it('filter out existing methods, input address === undefined && ds address === "", should filter', () => {
     const ds = mockDsFn();
-    ds[0].options.address = '';
+    ds[0].options = Object.assign(ds[0].options ?? {}, {address: ''});
```
stwiname, line 290: Again, i think in tests its easier to use non-null assertions
```suggestion
    ds[0].options!.address = '';
```

packages/cli/src/createProject.fixtures.ts
```diff
@@ -40,7 +41,10 @@ async function getExampleProject(networkFamily: string, network: string): Promis
     code: string;
     networks: {code: string; examples: ExampleProjectInterface[]}[];
   }[];
-  return templates.find((t) => t.code === networkFamily)?.networks.find((n) => n.code === network)?.examples[0];
+  const template = templates.find((t) => t.code === networkFamily)?.networks.find((n) => n.code === network)
+    ?.examples[0];
+  assert(template, 'Failed to get template');
```
stwiname, line 46: I think this function can return undefined, its a change in behaviour i don't think is needed

packages/cli/src/constants.ts
```diff
@@ -21,10 +22,9 @@ export const FUNCTION_REG = /function\s*:\s*['"]([^'"]+)['"]/;
 export const CHAIN_ID_REG = /chainId:\s*(\[[^\]]+\]|['"`][^'"`]+['"`])/;
 export const CAPTURE_CHAIN_ID_REG = /chainId:\s*("([^"]*)"|(?<!")(\d+))/;
 
-export const ACCESS_TOKEN_PATH = path.resolve(
-  process.env[process.platform === 'win32' ? 'USERPROFILE' : 'HOME'],
-  '.subql/SUBQL_ACCESS_TOKEN'
-);
+const rootPath = process.env[process.platform === 'win32' ? 'USERPROFILE' : 'HOME'];
+assert(rootPath, 'Cannot determine root path');
```
stwiname, line 26:
```suggestion
assert(rootPath, 'Cannot determine root path, please create an issue and include your OS. https://github.com/subquery/subql/issues/new');
```

packages/cli/src/controller/add-chain-controller.ts
```diff
@@ -241,8 +242,10 @@ export async function updateDockerCompose(projectDir: string, chainManifestPath:
   let subqlNodeService = getSubqlNodeService(dockerCompose);
   if (subqlNodeService) {
     // If the service already exists, update its configuration
-    subqlNodeService.command = subqlNodeService.command.filter((cmd) => !cmd.startsWith('-f='));
+    subqlNodeService.command = subqlNodeService.command?.filter((cmd) => !cmd.startsWith('-f=')) ?? [];
     subqlNodeService.command.push(`-f=app/${path.basename(chainManifestPath)}`);
+
+    assert(subqlNodeService.healthcheck, 'healthcheck is required');
```
stwiname, line 248: I don't think this needs to throw, we can just do nothing in this case

packages/cli/src/controller/generate-controller.spec.ts
yoozo: If changed this way, ESlint checks will report an error.

packages/cli/src/createProject.fixtures.ts
yoozo: However, cloneProjectTemplate() requires exampleProject to be defined.

packages/cli/src/commands/multi-chain/deploy.ts
```diff
@@ -114,8 +116,10 @@ export default class MultiChainDeploy extends Command {
         throw new Error(chalk.bgRedBright('Invalid IPFS CID'));
       }
 
+      assert(validator.chainId, 'Please set chainId in your project');
       if (!indexerVersions[validator.chainId]) {
         try {
+          assert(validator.manifestRunner, 'Please set manifestRunner in your project');
```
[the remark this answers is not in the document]
yoozo: Yes, I think we should deduplicate it. However, there is a fixed version that needs to be released now. We can start this work after that version is released

packages/cli/src/commands/deployment/deploy.ts
```diff
@@ -51,18 +52,22 @@ export default class Deploy extends Command {
...
-        flags.dict = await promptWithDefaultValues(cli, 'Enter dictionary', validateDictEndpoint, null, false);
+        flags.dict = await promptWithDefaultValues(cli, 'Enter dictionary', validateDictEndpoint, undefined, false);
       } else {
         flags.dict = validateDictEndpoint;
       }
+
+      assert(flags.dict, 'Dictionary is required');
     }
 
+    assert(validator.manifestRunner, 'Please set manifestRunner in your project');
```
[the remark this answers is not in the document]
yoozo: The manifestRunner in the validator type is currently defined as optional.

packages/cli/src/commands/init.ts
```diff
@@ -78,29 +79,26 @@ export default class Init extends Command {
       description: 'Give the starter project name',
     }),
   };
-  private projectPath: string; //path on GitHub
-  private project: ProjectSpecBase;
-  private location: string;
-  private networkFamily: NETWORK_FAMILY;
-  private network: string;
+  // private projectPath!: string; //path on GitHub
```
stwiname, line 82: Can be removed

packages/cli/src/modulars/moduleLoader.ts
```diff
@@ -39,11 +42,23 @@ export function loadDependency<N extends NETWORK_FAMILY>(network: N): ModuleCach
...
+    moduleCache[network]?.isRuntimeDs === undefined
   ) {
     throw new Error(`${packageName} is not compatible, please make sure package update to latest version`);
   }
   return moduleCache[network];
 }
+
+export function loadEthModule(): EthereumNetworkModule {
+  const ethModule = loadDependency(NETWORK_FAMILY.ethereum);
+  assert(ethModule, 'Eth module got undefined, expected to be loaded from local or global');
```
stwiname, line 56: This assertion should be redundant, `loadDependency` should never return undefined

packages/cli/tsconfig.json
```diff
@@ -8,7 +8,8 @@
     "rootDir": "src",
     "tsBuildInfoFile": "lib/.tsbuildinfo",
     "outDir": "lib",
-    "noImplicitAny": true
+    "noImplicitAny": true,
+    "strict": true
```
stwiname, line 12: This can now be set on the root tsconfig and removed from all the sub pacakges

stwiname (approved):

PR #2481, opened 2024-07-09
Fix admin api db size

jiqiang90:
# Description

Fixed issue for admin api, when dbSize not been set in the metadata, fetch and update dbSize will return undefined issue.

## Type of change

## Checklist

- [x] I have tested locally

### DIFF
--- a/packages/node-core/src/db/sync-helper.test.ts
+++ b/packages/node-core/src/db/sync-helper.test.ts
@@ -12,8 +12,12 @@ const nodeConfig = new NodeConfig({subquery: 'packages/node-core/test/v1.0.0', s
 
 describe('sync helper test', () => {
   let app: INestApplication;
+  let sequelize: Sequelize;
+  let schema: string;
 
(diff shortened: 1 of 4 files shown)


packages/node-core/src/db/sync-helper.test.ts
```diff
@@ -0,0 +1,48 @@
...
+        value JSONB,
+        "createdAt" TIMESTAMP WITH TIME ZONE NOT NULL,
+        "updatedAt" TIMESTAMP WITH TIME ZONE NOT NULL
+    )`);
+
+    const dbSize = await getDbSizeAndUpdateMetadata(sequelize, schema);
+    expect(dbSize).not.toBeUndefined();
+
+    await sequelize.dropSchema(schema, {});
+    await sequelize.close();
```
stwiname, line 46: I think this needs to be in afterEach, if anything above fails then it wont be cleaned up currently

stwiname (approved):

PR #2495, opened 2024-07-22
Poi fix and change 20240716

jiqiang90:
# Description

### Changed
- Breaking change: Require indexing environment timezone set to UTC, avoid inconsistent result from cache and database 
### Fixed
- Fix handle bigint type in jsonb array, both get and set method will generate consistent result 

## Type of change

## Checklist

- [ ] I have tested locally


packages/node-core/src/indexer/StoreOperations.test.ts
```diff
@@ -0,0 +1,201 @@
...
+      1
+    );
+    await flush(2);
+
+    const rawDb = (await cacheModel.model.findOne({where: {id: '0x01'}}))?.toJSON();
+
+    const rawCache = await (cacheModel as any).getCache.get('0x01');
+
+    const entityDb = EntityClass.create<EraEntity>('EraEntity', rawDb, {} as Store);
+    if (entityDb) {
```
stwiname, line 172: Wont this always be true?

packages/node-core/src/indexer/project.service.ts
```diff
@@ -157,6 +159,13 @@ export abstract class BaseProjectService<
     await this.dsProcessorService.validateProjectCustomDatasources(await this.getDataSources());
   }
 
+  private ensureTimezone(): void {
+    const timezone = process.env.TZ;
+    if (!timezone || timezone.toLowerCase() !== 'utc') {
+      throw new Error('Environment Timezone is not set to UTC. This may cause issues with indexing or proof of index');
```
stwiname, line 165: This should explain to the user how to set their timezone.

packages/node-core/src/utils/graphql.ts
```diff
@@ -112,5 +114,70 @@ export function getColumnOption(
...
+
+type OperationType = 'Get' | 'Set';
+
+/***
+ * Process nested json set/get, output is same as input value type
+ * @param field
+ * @param value
+ * @param operationType
+ */
+function processNestedJson(
```
stwiname, line 154: I don't think this is backwards compatible. Currently bigInts will have the `n` suffix (or hex for the network project) which doesnt work.

A problem i see with this is that there is no requirement that value has the same structure as the fields. An alternative would be to use the [reviver on JSON.parse](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse#using_the_reviver_parameter) to convert BigInts using the `n` suffix.

The downside to this is that the user might want a string with a digits followed by an `n` would be converted to a BigInt when it shouldn't. But it would also be backwards compatible with the special case used by the network project because it wouldn't convert hex. 

I think this function could also be further generalised. What happens if we want to do the same with Dates? I think the same issue would exist

packages/node-core/src/utils/graphql.ts
```diff
@@ -112,5 +114,70 @@ export function getColumnOption(
...
+): any {
+  // bigIntFields and nestJsonFields from this level in the entity/json
+  const bigIntFields = field.jsonInterface?.fields.filter((f) => f.type === 'BigInt');
+  const nestJsonFields = field.jsonInterface?.fields.filter((f) => f.jsonInterface);
+
+  const processBigIntFields = (value: any) => {
+    if (bigIntFields && bigIntFields.length) {
+      for (const bigIntField of bigIntFields) {
+        value[bigIntField.name] =
+          operationType === 'Get' ? BigInt(value[bigIntField.name]) : value[bigIntField.name].toString();
```
stwiname, line 167: This doesn't handle `value[bigIntField.name]` being undefined/null.

packages/node-core/src/utils/graphql.ts
```diff
@@ -112,5 +114,70 @@ export function getColumnOption(
...
+      }
+    }
+    return value;
+  };
+
+  if (nestJsonFields && nestJsonFields.length) {
```
stwiname, line 173:
```suggestion
  if (nestJsonFields) {
```

packages/node-core/src/utils/graphql.ts
```diff
@@ -112,5 +114,70 @@ export function getColumnOption(
...
+  field: GraphQLEntityField | GraphQLJsonFieldType,
+  value: any,
+  operationType: OperationType
+): any {
+  // bigIntFields and nestJsonFields from this level in the entity/json
+  const bigIntFields = field.jsonInterface?.fields.filter((f) => f.type === 'BigInt');
+  const nestJsonFields = field.jsonInterface?.fields.filter((f) => f.jsonInterface);
+
+  const processBigIntFields = (value: any) => {
+    if (bigIntFields && bigIntFields.length) {
```
stwiname, line 164:
```suggestion
    if (bigIntFields) {
```

packages/node-core/src/utils/graphql.ts
```diff
@@ -112,5 +114,70 @@ export function getColumnOption(
...
+        this.setDataValue(field.name, null);
+        return;
+      }
+      if (isArray(val)) {
+        const setValue = val.length === 0 ? [] : val.map((v) => processNestedJson(field, v, 'Set'));
+        this.setDataValue(field.name, setValue);
+      } else if (isObject(val)) {
+        this.setDataValue(field.name, processNestedJson(field, val, 'Set'));
+      } else {
+        throw new Error(`input for Json type is only support object or array, got type ${typeof val}`);
```
stwiname, line 139:
```suggestion
        throw new Error(`input for Json type only supports object or array, received type ${typeof val}`);
```

packages/node-core/src/indexer/StoreOperations.test.ts
jiqiang90: https://github.com/subquery/subql/blob/1cfae76b3534f6541f0c04a09416b79abafdd2ef/packages/node-core/src/indexer/store/entity.ts#L18

it should be true, but need this for return type assertion

packages/node-core/src/utils/graphql.ts
jiqiang90: bigIntFields can be empty []

packages/node-core/src/utils/graphql.ts
jiqiang90: nestJsonFields also can be []

packages/node-core/src/indexer/StoreOperations.test.ts
stwiname: I was getting confused with this not being undefined.https://github.com/subquery/subql/pull/2495/files#diff-46a6558964a1655301f62eff47b0ce02ce083d8500fbec26a8c446ce8f4eb423R28

I think there should be an expect here instead. If both entities are undefined then the operationStacks aren't updated and will always be the same, then this could pass when it shouldn't

packages/node-core/src/utils/graphql.ts
stwiname: If its empty then the `for` will not iterate anything

stwiname (approved):

PR #2505, opened 2024-07-29
Reload chainTypes

jiqiang90:
- Fixed reload chainTypes when project changed, 

## Type of change

## Checklist

- [ ] I have tested locally

### DIFF
--- a/packages/node-core/src/indexer/connectionPool.service.ts
+++ b/packages/node-core/src/indexer/connectionPool.service.ts
@@ -49,6 +49,7 @@ export class ConnectionPoolService<T extends IApiConnectionSpecific<any, any, an
 
   async onApplicationShutdown(): Promise<void> {
     await Promise.all(Object.values(this.allApi).map((api) => api?.apiDisconnect()));
+    this.allApi = {};
   }
 
   async addToConnections(api: T, endpoint: string): Promise<void> {
--- a/packages/node-core/src/indexer/connectionPool.service.ts
+++ b/packages/node-core/src/indexer/connectionPool.service.ts
@@ -251,4 +251,11 @@ export class ConnectionPoolService<T extends IApiConnectionSpecific<any, any, an
     this.lastCacheFlushTime = Date.now();
     await this.handleConnectionStateChange();
   }
+
+  async updateChainTypes(newChainTypes: unknown): Promise<void> {
+    logger.info(`Network chain types updated!`);
+    for (const endpoint in this.allApi) {
+      await this.allApi[endpoint].updateChainTypes?.(newChainTypes);
+    }
+  }
 }
--- a/packages/node-core/src/indexer/connectionPool.service.ts
+++ b/packages/node-core/src/indexer/connectionPool.service.ts
@@ -49,7 +49,6 @@ export class ConnectionPoolService<T extends IApiConnectionSpecific<any, any, an
 
   async onApplicationShutdown(): Promise<void> {
     await Promise.all(Object.values(this.allApi).map((api) => api?.apiDisconnect()));
-    this.allApi = {};
   }
 
   async addToConnections(api: T, endpoint: string): Promise<void> {
--- a/packages/node-core/src/indexer/connectionPool.service.ts
+++ b/packages/node-core/src/indexer/connectionPool.service.ts
@@ -252,9 +252,9 @@ export class ConnectionPoolService<T extends IApiConnectionSpecific<any, any, an
   }
 
   async updateChainTypes(newChainTypes: unknown): Promise<void> {
-    logger.info(`Network chain types updated!`);
     for (const endpoint in this.allApi) {
       await this.allApi[endpoint].updateChainTypes?.(newChainTypes);
     }
+    logger.info(`Network chain types updated!`);
   }
 }
--- a/packages/node/CHANGELOG.md
+++ b/packages/node/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Fixed
-- Fixed api didn't reload new deployment chainTypes when project upgraded (#2505)
+- Fixed api not reloading new deployment chainTypes when project upgrades (#2505)
 
 ## [5.0.1] - 2024-07-29
 ### Fixed
--- a/packages/node/CHANGELOG.md
+++ b/packages/node/CHANGELOG.md
@@ -6,6 +6,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- Fixed api didn't reload new deployment chainTypes when project upgraded (#2505)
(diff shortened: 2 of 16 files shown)


packages/node/src/indexer/x-provider/http.ts
```diff
@@ -54,7 +54,7 @@ export class HttpProvider implements ProviderInterface {
 
   readonly #stats: ProviderStats;
 
-  readonly #ctx: ReturnType<typeof context>;
+  readonly ctx: ReturnType<typeof context>;
```
stwiname, line 57: Whats the reason for this change? It removes the privacy of the variable

packages/node/src/indexer/project.service.ts
```diff
@@ -74,7 +74,8 @@ export class ProjectService extends BaseProjectService<
     return getTimestamp(block);
   }
 
-  protected onProjectChange(project: SubqueryProject): void | Promise<void> {
+  protected async onProjectChange(project: SubqueryProject): Promise<void> {
+    await this.apiService.reloadChainTypes(project.chainTypes);
     this.apiService.updateBlockFetching();
```
stwiname, line 79: Why not combine these into a single function?

packages/node-core/src/indexer/connectionPool.service.ts
```diff
@@ -49,6 +49,7 @@ export class ConnectionPoolService<T extends IApiConnectionSpecific<any, any, an
 
   async onApplicationShutdown(): Promise<void> {
     await Promise.all(Object.values(this.allApi).map((api) => api?.apiDisconnect()));
+    this.allApi = {};
```
stwiname, line 52: What's the reason for this change?

packages/node/src/indexer/api.service.ts
```diff
@@ -160,11 +160,24 @@ export class ApiService
...
+  }
+
+  async reloadChainTypes(chainTypes: unknown): Promise<void> {
+    // close any existing connections
+    await this.connectionPoolService.onApplicationShutdown();
+    logger.info(`Reload chainTypes`);
+    await this.load(chainTypes);
+  }
+
+  private async load(newChainTypes?: unknown): Promise<ApiService> {
```
stwiname, line 173: I think this could have a better function name.

packages/node/src/indexer/api.service.ts
```diff
@@ -160,11 +160,24 @@ export class ApiService
   }
 
   async init(): Promise<ApiService> {
+    return this.load();
+  }
+
+  async reloadChainTypes(chainTypes: unknown): Promise<void> {
```
stwiname, line 166: Do we have to recreate all the api instances? This seems quite inefficient, creating new polkadot api instances is slow.

stwiname (commented):
It would be good to separate the timezone changes to a separate PR to get a release out

packages/node/src/indexer/x-provider/http.ts
jiqiang90:
```
Cannot read private member #ctx from an object whose class did not declare it
TypeError: Cannot read private member #ctx from an object whose class did not declare it
    at Proxy.disconnect (/Users/jiatwork/ofn/subql/packages/node/src/indexer/x-provider/http.ts:122:16)
    at RpcCore.disconnect (/Users/jiatwork/ofn/subql/node_modules/@polkadot/rpc-core/cjs/bundle.js:103:30)

```

It throwing this error, I feel it should be safe to remove # here as it only accessed from polkadot

packages/node-core/src/indexer/connectionPool.service.ts
jiqiang90: Trying to reset all existing connection here, but I think need a better place

packages/node/src/indexer/project.service.ts
jiqiang90: `updateBlockFetching` also used in api service constructor, also make sense keep them independent.

packages/node/CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- Fixed api didn't reload new deployment chainTypes when project upgraded (#2505)
```
stwiname, line 10:
```suggestion
- Fixed api not reloading new deployment chainTypes when project upgrades (#2505)
```

packages/node/src/indexer/apiPromise.connection.ts
```diff
@@ -123,6 +131,14 @@ export class ApiPromiseConnection
     await this.unsafeApi.disconnect();
   }
 
+  async updateRegisteredChainTypes(chainTypes: RegisteredTypes): Promise<void> {
+    this.unsafeApi = await ApiPromiseConnection.createApiPromise(
```
stwiname, line 135: This doesn't reuse the connection. If we have a websocket we don't want to create a new connection

packages/node/src/indexer/api.service.ts
```diff
@@ -206,12 +206,16 @@ export class ApiService
...
   }
 
-  updateBlockFetching(): void {
-    const onlyEventHandlers = isOnlyEventHandlers(this.project);
+  async updateChainTypes(newChainTypes: unknown): Promise<void> {
+    const chainTypes = await updateChainTypesHasher(newChainTypes);
+    await this.connectionPoolService.updateChainTypes(chainTypes);
+  }
+
+  updateBlockFetching(project?: SubqueryProject): void {
```
stwiname, line 217: Theres no need to pass through the newChainTypes/project in these. `this.project` will be the updated project

packages/node/src/indexer/apiPromise.connection.ts
jiqiang90: image

New approach : Extract current ws/http provider from current apiPromise instance, and reconstructed a new apiPromise

packages/node/src/indexer/apiPromise.connection.ts
```diff
@@ -123,6 +124,16 @@ export class ApiPromiseConnection
     await this.unsafeApi.disconnect();
   }
 
+  async updateChainTypes(chainTypes: RegisteredTypes): Promise<void> {
+    // Typeof Decorate<'promise' | 'rxjs'>, but we need to access this private method
+    const currentApiOptions = (this.unsafeApi as any)._options as ApiOptions;
+    const apiOption = {
+      ...currentApiOptions,
+      ...chainTypes,
+    };
```
stwiname, line 133:
```suggestion
    const apiOption: ApiOptions = {
      ...currentApiOptions,
      chainTypes,
    };
```

stwiname (approved):

packages/node/src/indexer/apiPromise.connection.ts
jiqiang90: should be same line as https://github.com/subquery/subql/pull/2505/files#diff-c9f4c413e59e6f6047e10cd5abdc7e5aef91fd9d24a84732459c7d6acf395631R84

Issue #2520, opened 2024-08-07
The query app being killed due to idle TCP connections to postgresql db


mpastecki:
### Prerequisites

* [x] Are you running the latest version(s)?
* [x] Have you searched the [documentation for your issue](https://academy.subquery.network/)?
* [x] Did you search for an existing issue or pull request?

### Description

The query app being killed due to idle connections to postgresql db.

### Details

In our environment the tcp connections between services are killed if found to be inactive. To be more precise we use a Service Mesh with Envoy Proxies attached to each services. [Envoy proxy](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/tcp_proxy/v3/tcp_proxy.proto) by default kills connections that are inactive for 1 hour.

The query package uses a [pg.Pool](https://node-postgres.com/apis/pool) instance to manage database clients.

There are two pg.Clients explicitly used that cause us some issues:

https://github.com/subquery/subql/blob/e9f357d0ea1d6ffa41a3c807ec6f3ce755c4b737/packages/query/src/configure/configure.module.ts#L13-L24

Here, as I get it, it is to test whether the provided SSL configuration is valid, and clear it otherwise. The problem is that the client is not released once the check happens. So if there is any mechanism that checks for idle connections, the client will be killed eventually and that causes the subql/query app to be killed as there is no error listener attached to the client. Attaching an error listener to the pool is not enough to handle these kind of exceptions. One has to be attached to the client too.

The simplest fix would be to modify the code to release the client once it is no longer in use:

```ts
 async function ensurePool(poolConfig: PoolConfig): Promise<Pool> { 
   const pgPool = new Pool(poolConfig); 
   try { 
     const pgClient = await pgPool.connect();
   } catch (e: any) { 
     if (JSON.stringify(e.message).includes(CONNECTION_SSL_ERROR_REGEX)) { 
       poolConfig.ssl = undefined; 
       return ensurePool(poolConfig); 
     } 
   } finally {
     // Release the client back to the pool
     pgClient.release();
   }
   return pgPool; 
 } 
```

The other pg.Client is explicitly initialized here:

https://github.com/subquery/subql/blob/e9f357d0ea1d6ffa41a3c807ec6f3ce755c4b737/packages/query/src/graphql/graphql.module.ts#L163-L176

[...]

PR #2525, opened 2024-08-08
Update PostgreSQL connection pool handling

mpastecki:
# Description

This improves the handling of database pool connections that are explicitly used in the code to verify SSL settings and to set up the database schema-change listener. Previously, these two clients were either idle or waiting for the rare event of a schema change.

In environments where idle TCP connections are eventually killed, these connections were terminated, causing the entire app to crash due to an uncaught exception.

Additionally, an error handler has been attached to the schema change PostgreSQL client to listen for `error` events. This should improve visibility into any issues that arise. When an error occurs, it will be logged, and the application will be terminated to prevent potential data corruption.

Fixes ([#2520 ](https://github.com/subquery/subql/issues/2520))

## Type of change

## Checklist

- [x] I have tested locally

### DIFF
--- a/packages/query/src/graphql/graphql.module.ts
+++ b/packages/query/src/graphql/graphql.module.ts
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: GPL-3.0
 
 import assert from 'assert';
+import {setInterval} from 'timers';
 import PgPubSub from '@graphile/pg-pubsub';
 import {Module, OnModuleDestroy, OnModuleInit} from '@nestjs/common';
 import {HttpAdapterHost} from '@nestjs/core';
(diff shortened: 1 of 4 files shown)


packages/query/src/graphql/graphql.module.ts
```diff
@@ -127,6 +128,18 @@ export class GraphqlModule implements OnModuleInit, OnModuleDestroy {
     }
   }
 
+  private setupKeepAlive(pgClient) {
```
stwiname, line 131:
```suggestion
  private setupKeepAlive(pgClient: PoolClient) {
```
This will need an update import

packages/query/src/graphql/graphql.module.ts
```diff
@@ -127,6 +128,18 @@ export class GraphqlModule implements OnModuleInit, OnModuleDestroy {
...
+    setInterval(() => {
+      (async () => {
```
stwiname, line 133:
```suggestion
      void (async () => {
```

Just to stop the linter warning

stwiname (commented):
This looks good, just a couple of minor improvements

mpastecki:
Thanks, updated.

stwiname (approved):

PR #2526, opened 2024-08-09
add support for null ordering in 'order by' query (#2387)

### LINKED ISSUE #2387: [$300 in SQT] Query Service: Order by null order

abhishek818:
# Description
adds support for null ordering in 'order by' query

Fixes #2387 

## Type of change

## Checklist

- [ ] I have tested locally
- [X] Linked to any relevant issues
- [X] My code is up to date with the base branch

### DIFF
--- a/packages/cli/src/controller/codegen-controller.test.ts
+++ b/packages/cli/src/controller/codegen-controller.test.ts
@@ -4,7 +4,7 @@
 import fs from 'fs';
 import path from 'path';
 import {rimraf} from 'rimraf';
-import {codegen, generateModels} from './codegen-controller';
+import {codegen} from './codegen-controller';
 
 jest.setTimeout(30000);
 
@@ -95,9 +95,6 @@ describe('Codegen can generate schema', () => {
 
     const fooFile = await fs.promises.readFile(`${projectPath}/src/types/models/Foo.ts`, 'utf8');
 
-    const fooFile2 = await fs.promises.readFile(`${projectPath}/src/types/models/index.ts`, 'utf8');
-    console.log(fooFile2);
-    expect(fooFile2).toContain(` `);
     expect(fooFile).toContain(
       `import {
     Bar,
--- a/packages/query/CHANGELOG.md
+++ b/packages/query/CHANGELOG.md
@@ -18,6 +18,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 - Update dependencies (#2518)
 
+### Added
+- add support for null ordering in 'order by' query (#2387)
+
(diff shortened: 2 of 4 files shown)


abhishek818:
@stwiname @jiqiang90 hey are the test suites broken (some pass, others fail) ?
Was trying to add a test case in packages\query\src\graphql\graphql.test.ts file. Can you help with running the tests (any specific set of commands etc.)

abhishek818:
anyways up for a review..

stwiname:
> @stwiname @jiqiang90 hey are the test suites broken (some pass, others fail) ? Was trying to add a test case in **packages\query\src\graphql\graphql.test.ts** file. Can you help with running the tests (any specific set of commands etc.)

@abhishek818 because its an integration test it needs a DB to connect to. You can either run postgres yourself or `yarn test:docker` but you would probably want to change it to just run that test.

These changes look good but tests are still needed,

abhishek818:
Added the tests, Up for review..

`test-1  | PASS packages/query/src/graphql/graphql.test.ts (863 MB heap size)`

stwiname (approved):

stwiname (changes requested):
There is a failing test `GraphqlHistorical › to filter historical items when ordering`, this is a change in behaviour as if the nulls order is not specified then it should be default. 

As the [Postgres docs](https://www.postgresql.org/docs/current/queries-order.html#QUERIES-ORDER) say `ASC` and `DESC` change the default null behaviour. This should be retained if null order isn't specified.

abhishek818:
> There is a failing test `GraphqlHistorical › to filter historical items when ordering`, this is a change in behaviour as if the nulls order is not specified then it should be default.
> 
> As the [Postgres docs](https://www.postgresql.org/docs/current/queries-order.html#QUERIES-ORDER) say `ASC` and `DESC` change the default null behaviour. This should be retained if null order isn't specified.

Thanks for pointing this out. Have fixed this.
Up for a review now..

stwiname (approved):

Issue #1887, opened 2023-07-17
Allow query historical data by block range


jiqiang90:
### Prerequisites

* [ ] Are you running the latest version(s)?
* [ ] Have you searched the [documentation for your issue](https://academy.subquery.network/)?
* [ ] Did you search for an existing issue or pull request?

### Description

We currently allow pass a block height, and get any result less equal than this height 

![image](https://github.com/subquery/subql/assets/10431657/bb004d53-c1d3-4537-bbd4-6b10cde03e30)

We now want to add a filter for block height, so it can filter data with in a specific block range 

![image](https://github.com/subquery/subql/assets/10431657/783b903e-446a-41b3-8648-d27f984c1291)

### Details

These details can help to reproduce the environment the issue is occurring

**Local Environment:** [You can get this information from executing `subql version`.]
**Query Version:** [What is the version of the query service?]
**Indexer Version:** [What is the version of the indexer service?]
**Network Details:**
  * [Network]
  * [Block height, a block height where the issue is happening]
  * [Dictionary endpoint, if used]

### Steps to Reproduce

1. [First Step]
2. [Second Step]
3. [and so on...]

**Example project:** [A link to a minimal example that can reproduce the issue]

**Expected behavior:** [What you expected to happen]

**Actual behavior:** [What actually happened]

### Any other information

Is there any other information you would like to add?

stwiname:
This can be done after we update postgraphile version

abhishek818:
Will take this up!

stwiname:
> Will take this up!

@abhishek818 i have assigned you to this. Let us know if you have any questions

abhishek818:
> > Will take this up!
> 
> @abhishek818 i have assigned you to this. Let us know if you have any questions

Added a query in draft PR, pls have a look.

Gmin2:
hey @stwiname is this still open. like I see a draft PR but there is no contiuation of it!

and also is this is still part of developer guild !

PR #2540, opened 2024-08-19
query historical data by block range - initial changes (#1887)

abhishek818:
# Description

Fixes # (issue)

## Type of change

- [X] This change requires a documentation update

## Checklist

- [ ] I have tested locally
- [X] I have added tests relevant to my changes


abhishek818:
@stwiname 

Hey, can you give this PR changes a brief look, I will get time to work on this only during coming weekend AND I might get some queries about the requirements when nobody online will be around,
So just wanted to see if this is going in right direction?

Queries of below type is implemented currently (perhaps partially). 
```
query {
        items(blockRange: [0, 100]) {
```

Further we need below types also right ??
```
query {
        items(blockHeight: {
                                      lessThan/lessThanOrEqualTo/greaterThan/greaterThanOrEqualTo : <value>
                                      }
                 ) {
```

What further requirements/any sample responses/queries expected to cover are needed do tell, thanks..

stwiname:
Hey @abhishek818 

I think this is fine
```gql
query {
        items(blockRange: [0, 100]) {
```

As for the other case I think we could drop `lessThan/lessThanOrEqualTo/greaterThan/greaterThanOrEqualTo` as we could cover that with the previous example. But there could be an addition of making the end block nullable.

Example of no end range, this would get the whole block range.
```gql
query {
    items(blockRange: [0, null]) {
```

As for other requirements:
* I think for now this can be limited to singular entities with an ID. 
* There should be a limit to the number of results, same as how paging works, we don't want it to return a huge number of results, the default limit should be the same as the paging limit. 
* Following on from the limit there also needs to be ASC/DESC for block height.

I think these are relatively straight forward in terms of how the API should be implemented but im happy to give examples if clarification is needed.

stwiname:
@abhishek818 I've made a minor change to the issue. If there is an entity that gets deleted it should return null at the block height that it is removed. See the example in the issue for more details

abhishek818:
@stwiname hey, some doubts..

1.
```
example2: {
    // Keyed by block height
    5: {
      id: 'foo',
      field1: 1,
      field2: 'bar',
    },
    10: { 
    .....
```

Related to "Keyed by block height", do you mean there is always a 'created_at_block_height' or similar column in tables containing '_block_range' ?

2. 
` 90: null, // indicates the entity was deleted`

Does this refers to a entity that is about to be deleted in the same graphql query containing blockRange param? If yes, can you give a example query..

3. 
> There should be a limit to the number of results, same as how paging works, we don't want it to return a huge number of results, the default limit should be the same as the paging limit.

does this points to use of "limit <num>" keyword of sql OR use of something like argv('query-limit') ?

abhishek818:
damn, auckland timezone is way too ahead..

abhishek818:
Alrighty, have got some idea after reading https://academy.subquery.network/indexer/run_publish/historical.html , will start implementing soon and then will ask any queries if needed.

Issue #2565, opened 2024-09-25
Bypass blocks limitations


stwiname:
## Issue

When specifying bypassBlocks, internally it gets expanded into a set, if the blocks to bypass are greater than the Set size limit it will throw an error. 

## Solution

Optimise this by not expanding block ranges (e.g. `"1000-100000"`) and filling up the Set. Instead we should not expand the block ranges and directly calculate from the provided data.

PR #2566, opened 2024-09-27
Rework bypass blocks to not expand large block ranges

stwiname:
# Description
If a large range of blocks was specified to be bypassed, then it could result in array/map size limits. This change simplifies the logic to calculate bypass blocks, as well as no longer expanding block ranges to individual blocks. 

Fixes https://github.com/subquery/subql/issues/2565

## Type of change

## Checklist

- [x] I have tested locally

### DIFF
--- a/packages/node-core/src/indexer/fetch.service.ts
+++ b/packages/node-core/src/indexer/fetch.service.ts
@@ -359,7 +359,7 @@ export abstract class BaseFetchService<DS extends BaseDataSource, B extends IBlo
       const currentDS = datasources.get(currentHeight);
       // If the value for the current height is an empty array, then it's a gap
       if (currentDS?.length === 0) {
-        bypassBlocks.push(`${currentHeight}-${nextHeight}`);
+        bypassBlocks.push(`${currentHeight}-${nextHeight - 1}`);
       }
     }
     return bypassBlocks;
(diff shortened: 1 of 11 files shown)


packages/node-core/src/indexer/fetch.service.ts
```diff
@@ -344,49 +317,52 @@ export abstract class BaseFetchService<DS extends BaseDataSource, B extends IBlo
...
-      // inspect limits the number of logged blocks to 100
-      logger.info(`Bypassing blocks: ${util.inspect(pollutedBlocks, {maxArrayLength: 100})}`);
+    for (let i = 0; i < heights.length - 1; i++) {
+      const currentHeight = heights[i];
+      const nextHeight = heights[i + 1];
+
+      const currentDS = datasources.get(currentHeight);
+      // If the value for the current height is an empty array, then it's a gap
+      if (currentDS?.length === 0) {
+        bypassBlocks.push(`${currentHeight}-${nextHeight}`);
```
yoozo, line 362:
```suggestion
        bypassBlocks.push(`${currentHeight}-${nextHeight - 1}`);
```

yoozo: nextHeight should not be skipped

stwiname: Good catch

yoozo (approved):

Issue #2563, opened 2024-09-24
Option to disable the cache


stwiname:
The store cache offers a lot of performance increase but it also adds a lot of complexity, to simplify running and be able to diagnose any issues related to the cache there should be an option to disable it.

PR #2568, opened 2024-10-06
Disable cache

stwiname:
# Description
Provides an option to disable the cache for the store.

## Details

* Define/update and interfaces to provide abstractions to the Model, Metadata and POI interfaces as well as the `StoreCacheService`
* Implement non-cache versions of abstracted interfaces
* Update cache classes to match any interface changes
* Create db transactions for when cache is not used

## TODOs

- [x] Fix remaining build errors and update interfaces to abstract between flushing cache/managing db transaction
- [x] Implement flag to change the option and inject the appropriate class (StoreCacheService or PlainStoreModelService)
- [x] Implement tests and update existing tests
- [x] Improve code reuse by deduplicating any functionality between cache and non-cache models

Fixes https://github.com/subquery/subql/issues/2563

## Type of change

## Checklist

- [ ] I have tested locally


packages/node-core/src/configure/ProjectUpgrade.service.ts
```diff
@@ -150,16 +157,10 @@ export class ProjectUpgradeService<P extends ISubqueryProject = ISubqueryProject
       return;
     }
     this.#initialized = true;
-    this.#storeCache = storeService.storeCache;
+    this.#storeCache = storeService.storeModel;
```
stwiname, line 160: I think this needs a better name `storeModel` makes no sense in this context

packages/node-core/src/indexer/storeCache/storeModel.service.ts
```diff
@@ -19,19 +20,29 @@ export interface IStoreModelService {
   poi: IPoi | null;
   metadata: IMetadata;
 
+  init(historical: boolean, useCockroachDb: boolean, meta: MetadataRepo, poi?: PoiRepo): void;
```
stwiname, line 23: This shouldn't be on the interface in the same way the constructor isn't on the interface.

packages/node-core/src/indexer/storeCache/storeModel.service.ts
```diff
@@ -19,19 +20,29 @@ export interface IStoreModelService {
...
 
   // addExporter(entity: string, exporterStore: CsvStoreService): void;
 
   applyPendingChanges(height: number, dataSourcesCompleted: boolean): Promise<void>;
 
   updateModels({modifiedModels, removedModels}: {modifiedModels: ModelStatic<any>[]; removedModels: string[]}): void;
+
+  resetData?(): Promise<void>;
+
+  flushData?(forceFlush?: boolean): Promise<void>;
```
stwiname, line 35: Optional methods shouldn't exist on the interface. If methods are needed then determine the implementation instead

packages/node-core/src/indexer/store.service.ts
```diff
@@ -79,7 +80,7 @@ export class StoreService {
   constructor(
     private sequelize: Sequelize,
     private config: NodeConfig,
-    readonly storeModel: IStoreModelService,
+    readonly storeModel: BaseStoreModelService<ICachedModelControl | IModel<any>>,
```
stwiname: Is there a reason this cant be `IStoreModelService`?

yoozo: Some methods are only available in cacheModel, and not necessary for plainModel, such as `resetData` and `flushData`. However, I don’t think it’s appropriate to define them in `IStoreModelService`.

stwiname (commented):
Does it make sense for the plain models to be dependencies of the cache equivalent to reduce code?

packages/node-core/src/indexer/storeModelProvider/storeCache.service.ts
```diff
@@ -14,23 +14,23 @@ import {exitWithError} from '../../process';
...
+import {CacheMetadataModel} from './metadata';
+import {METADATA_ENTITY_NAME} from './metadata/utils';
+import {CachedModel} from './model';
+import {CachePoiModel, POI_ENTITY_NAME} from './poi';
+import {Exporter, ICachedModelControl, IStoreModelProvider, FlushPolicy} from './types';
 
 const logger = getLogger('StoreCacheService');
 
+export function isCachePolicy(modelProvider: IStoreModelProvider): modelProvider is StoreCacheService {
+  return modelProvider.flushPolicy === FlushPolicy.Cache;
```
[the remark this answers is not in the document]
yoozo: No, It is not a promise.

packages/node-core/src/indexer/store.service.ts
```diff
@@ -163,13 +177,15 @@ export class StoreService {
     }
     logger.info(`Historical state is ${this.historical ? 'enabled' : 'disabled'}`);
 
-    this.storeCache.init(this.historical, this.dbType === SUPPORT_DB.cockRoach, this.metaDataRepo, this.poiRepo);
+    if (isCachePolicy(this.modelProvider)) {
+      this.modelProvider.init(this.historical, this.dbType === SUPPORT_DB.cockRoach, this.metaDataRepo, this.poiRepo);
```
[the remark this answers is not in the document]
yoozo: No, It is not a promise.

packages/node-core/src/indexer/storeModelProvider/model/model.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+
+jest.setTimeout(50_000);
+
+describe('Model provider consistency test', () => {
+  let sequelize: Sequelize;
+  let schema: string;
+  let model: any;
+
+  beforeAll(async () => {
+    sequelize = new Sequelize(
```
stwiname, line 28: This is an integration test and the file should be `.test.ts`

packages/node-core/src/indexer/storeModelProvider/model/model.ts
```diff
@@ -0,0 +1,199 @@
...
+      transaction: tx,
+      updateOnDuplicate: Object.keys(data[0]) as unknown as (keyof T)[],
+    });
+
+    for (const item of waitCreateDatas) {
+      this._setTxCreatedCache(item);
+    }
+  }
+
+  async get(id: string, tx: Transaction): Promise<T | undefined> {
```
stwiname, line 65: I'm not sure why TX is required here but optional on the set methods

packages/node-core/src/indexer/storeModelProvider/model/model.ts
```diff
@@ -0,0 +1,199 @@
...
+  bulkCreate(data: T[], blockHeight: number, tx?: Transaction): Promise<void>;
+  bulkUpdate(data: T[], blockHeight: number, fields?: string[], tx?: Transaction): Promise<void>;
+
+  bulkRemove(ids: string[], blockHeight: number, tx?: Transaction): Promise<void>;
+}
+
+// All operations must be carried out within a transaction.
+export class PlainModel<T extends BaseEntity = BaseEntity> implements IModel<T> {
+  // Record the data that was first created in the current transaction.
+  private currentTxCreatedRecord: Record<string, T> = {};
```
stwiname, line 27: I think this should be avoided and be done on a DB level, it should be much simpler and less prone to bugs. Its probably just a matter of updating the hooks.

This also lacks an explanation for its purpose.

packages/node-core/src/indexer/storeModelProvider/storeModel.service.ts
```diff
@@ -0,0 +1,87 @@
...
+      if (!this.poiRepo) {
+        return null;
+        // throw new Error('Poi entity has not been set on store cache');
+      }
+      this.cachedModels[POI_ENTITY_NAME] = new PlainPoiModel(this.poiRepo) as any;
+    }
+
+    return this.cachedModels[POI_ENTITY_NAME] as unknown as PlainPoiModel;
+  }
+
```
stwiname, line 51: Could this be moved to the base service?

packages/node-core/src/indexer/storeModelProvider/storeModel.service.ts
```diff
@@ -0,0 +1,87 @@
...
+    if (this.config.csvOutDir) {
+      const exporterStore = new CsvStoreService(entityName, this.config.csvOutDir);
+      this.addExporter(plainModel, exporterStore);
+    }
+
+    return plainModel;
+  }
+
+  private addExporter(model: PlainModel, exporterStore: CsvStoreService): void {
+    throw new Error('Not implemented');
```
stwiname, line 66: Still needs to be done

packages/node-core/src/meta/health.service.ts
```diff
@@ -39,7 +42,7 @@ export class HealthService {
     }
 
     if (healthy !== this.indexerHealthy) {
-      await this.storeService.storeCache.metadata.model.upsert({key: 'indexerHealthy', value: healthy});
+      await this.storeService.modelProvider.metadata.set('indexerHealthy', healthy);
```
stwiname, line 45: This used the model directly because we don't want this to be cached and add delay.

packages/node-core/src/indexer/storeModelProvider/storeModel.service.ts
yoozo: BaseCacheService currently extends BaseService, and I think it should not be moved under baseService.

packages/node-core/src/indexer/storeModelProvider/baseStoreModel.service.ts
```diff
@@ -0,0 +1,47 @@
...
+export abstract class BaseStoreModelService<M = IModel<any>> {
+  protected historical = true;
+  protected poiRepo?: PoiRepo;
+  protected metadataRepo?: MetadataRepo;
+  protected cachedModels: Record<string, M> = {};
+  protected useCockroachDb?: boolean;
+
+  protected abstract createModel(entity: string): M;
+
+  init(historical: boolean, useCockroachDb: boolean, meta: MetadataRepo, poi?: PoiRepo): void {
```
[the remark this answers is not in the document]
yoozo: It needs to rely on StoreService.initCoreTables for its initialization.

packages/node-core/src/indexer/storeModelProvider/model/model.ts
```diff
@@ -0,0 +1,185 @@
...
+    const batchSize = 10000;
+    for (let i = 0; i < uniqueDatas.length; i += batchSize) {
+      const batchDatas = uniqueDatas.slice(i, i + batchSize);
+
+      if (!this.historical) {
+        await this._bulkCreate(batchDatas, tx);
+        continue;
+      }
+
+      await this.model.destroy({
```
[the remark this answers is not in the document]
yoozo: I set up a beforeBulkDestroy hook, which will only delete data at the current height.

packages/node-core/src/indexer/storeModelProvider/model/model.ts
```diff
@@ -0,0 +1,185 @@
...
+      (acc, curr) => {
+        acc[curr.id] = curr;
+        return acc;
+      },
+      {} as Record<string, T>
+    );
+    const uniqueDatas = Object.values(uniqueMap);
+
+    // Batch insert every 10000 data
+    const batchSize = 10000;
```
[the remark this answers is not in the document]
yoozo: To prevent the batch processing from causing a performance impact on the database due to the large amount of data.

packages/node-core/src/indexer/storeModelProvider/model/cacheModel.ts
```diff
@@ -199,24 +169,25 @@ export class CachedModel<
...
-    const [res] = await this.getByField(field, value, {
-      ...defaultOptions,
-      limit: 1,
-    });
-
-    return res;
-  }
-
-  set(id: string, data: T, blockHeight: number): void {
+  // async getByField(
```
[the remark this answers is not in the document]
yoozo: Yes, there is no place where it needs to be used.

packages/node-core/src/indexer/storeModelProvider/model/model.ts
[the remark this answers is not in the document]
stwiname: I think we can remove this. Users can implement it themselves with the current API and it is very unlikely to happen. 

We don't have this function for the cache so i don't think we need it for the plain version

packages/node-core/src/indexer/storeModelProvider/model/model.ts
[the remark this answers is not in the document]
stwiname: I don't think a hook is necessary for that, as there is no where else this could happen.

It also lacks any explanation, is the goal of this to remove all previous entries of the same entity in the same block?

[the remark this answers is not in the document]
yoozo: Yes, To ensure that only one record is retained at the current height. Case: Repeat the process of create - delete - create.

packages/node-core/src/indexer/store.service.ts
```diff
@@ -345,8 +361,18 @@ export class StoreService {
       return !disableHistorical;
     }
   }
-  setBlockHeight(blockHeight: number): void {
+  async setBlockHeight(blockHeight: number): Promise<void> {
     this._blockHeight = blockHeight;
+
+    if (this.modelProvider instanceof PlainStoreModelService) {
+      assert(!this.#transaction, new Error(`Transaction is reopening in setBlockHeight ${blockHeight}`));
```
stwiname, line 368:
```suggestion
      assert(!this.#transaction, new Error(`Transaction already exists for height: ${blockHeight}`));
```

packages/node-core/src/indexer/storeModelProvider/model/model.ts
```diff
@@ -0,0 +1,176 @@
...
+    if (!data.length) return;
+
+    const uniqueMap = data.reduce(
+      (acc, curr) => {
+        acc[curr.id] = curr;
+        return acc;
+      },
+      {} as Record<string, T>
+    );
+    const waitCreateDatas = Object.values(uniqueMap);
```
stwiname, line 46: Can be removed

packages/node-core/src/indexer/storeModelProvider/model/model.ts
```diff
@@ -0,0 +1,176 @@
...
+export class PlainModel<T extends BaseEntity = BaseEntity> implements IModel<T> {
+  private exporters: Exporter[] = [];
+
+  constructor(
+    readonly model: ModelStatic<Model<T, T>>,
+    private readonly historical = true
+  ) {}
+
+  // Only perform bulk create data, without paying attention to any exception situations.
+  private async _bulkCreate(data: T[], tx?: Transaction): Promise<void> {
```
stwiname, line 36: This function can be merged with `bulkCreate`

packages/node-core/src/indexer/storeModelProvider/model/model.ts
```diff
@@ -0,0 +1,176 @@
...
+    if (fields) {
+      throw new Error(`Currently not supported: update by fields`);
+    }
+
+    if (!this.historical) {
+      await this._bulkCreate(data, tx);
+      return;
+    }
+
+    await this.model.destroy({
```
stwiname, line 115: A comment is needed explaining why this happens.

packages/node-core/src/indexer/storeModelProvider/model/model.ts
```diff
@@ -0,0 +1,176 @@
...
+    await this._bulkCreate(data, tx);
+  }
+
+  async bulkRemove(ids: string[], blockHeight: number, tx?: Transaction): Promise<void> {
+    if (!ids.length) return;
+    if (!this.historical) {
+      await this.model.destroy({where: {id: ids} as any, transaction: tx});
+      return;
+    }
+    await this.markAsDeleted(ids, blockHeight, tx);
```
stwiname, line 140: Should this be in an `else`?

Issue #2572, opened 2024-10-17
[cosmos indexer] Question about delay after indexer sync to the latest block.


99Kies:
### Prerequisites

* [x] Are you running the latest version(s)? `subquerynetwork/subql-node-cosmos:latest`
* [x] Have you searched the [documentation for your issue](https://academy.subquery.network/)?
* [x] Did you search for an existing issue or pull request?

### Description

After I sync to the highest block, the indexer service has latency issues if there are new transactions on the chain. (I'm using rpc provided by node I started myself, so there is no ratelimit issue)

### Details

This is my indexer startup configuration parameters:

```
  - -f=/app
  - --db-schema=app
  - --workers=4
  - --batch-size=30
  - --unfinalized-blocks=true
```

This is my log info:
```
2024-10-11T06:55:42.907Z <WorkerBlockDispatcherService> INFO Enqueueing blocks 6471885...6471885, total 1 blocks
2024-10-11T06:55:50.966Z <benchmark> INFO INDEXING: 0.20 blocks/s. Target height: 6,471,885. Current height: 6,471,810. Estimated time remaining: 0 days 00 hours 06 mins
2024-10-11T06:55:51.411Z <WorkerBlockDispatcherService> INFO {"threadId":1,"fetchedBlocks":13,"toFetchBlocks":19,"isIndexing":false}
2024-10-11T06:55:51.412Z <WorkerBlockDispatcherService> INFO {"threadId":2,"fetchedBlocks":14,"toFetchBlocks":19,"isIndexing":false}
2024-10-11T06:55:51.412Z <WorkerBlockDispatcherService> INFO {"threadId":3,"fetchedBlocks":14,"toFetchBlocks":19,"isIndexing":false}
2024-10-11T06:55:51.412Z <WorkerBlockDispatcherService> INFO {"threadId":4,"fetchedBlocks":6,"toFetchBlocks":18,"isIndexing":false}
2024-10-11T06:55:52.909Z <WorkerBlockDispatcherService> INFO Enqueueing blocks 6471886...6471886, total 1 blocks
```

And I got this warning info

```
WARN Endpoint http://xxxxxxxxx:26657 experienced an error (default). Suspending for 960s.
```

```
Suspended endpoints:

- xxxxxxx
  Failures: 5
  Backoff (ms): 960000

- xxxxxxx
  Failures: 5
  Backoff (ms): 960000
```

and error info:
```
[...]

PR #2582, opened 2024-10-30
Fix/multi endpint delay

yoozo:
# Description
Multi-endpoint indexing logic delay and other related issues

Fixes https://github.com/subquery/subql/issues/2572

- Fix the issue where the backoff delay increases excessively after multiple failures.
- Correctly handle timeout and rate limit errors.
- Resolve the issue where default errors are treated as rate limit errors on the endpoint.
- Add control logic for rateLimitDelay, default to 20 seconds.

## Type of change

## Checklist

- [x] I have tested locally

### DIFF
--- a/packages/node-core/src/indexer/connectionPool.service.ts
+++ b/packages/node-core/src/indexer/connectionPool.service.ts
@@ -102,13 +102,6 @@ export class ConnectionPoolService<T extends IApiConnectionSpecific<any, any, an
         if (prop === 'fetchBlocks') {
           return async (heights: number[], ...args: any): Promise<any> => {
             try {
-              // Check if the endpoint is rate-limited
-              if (await this.poolStateManager.getFieldValue(endpoint, 'rateLimited')) {
-                const rateLimitDelay = await this.poolStateManager.getFieldValue(endpoint, 'rateLimitDelay');
-                logger.info(`throtling on ratelimited endpoint ${rateLimitDelay / 1000}s`);
-                await delay(rateLimitDelay / 1000);
-              }
-
               const start = Date.now();
               const result = await target.fetchBlocks(heights, ...args);
               const end = Date.now();
--- a/packages/node-core/src/indexer/connectionPool.service.ts
+++ b/packages/node-core/src/indexer/connectionPool.service.ts
@@ -102,6 +102,13 @@ export class ConnectionPoolService<T extends IApiConnectionSpecific<any, any, an
         if (prop === 'fetchBlocks') {
           return async (heights: number[], ...args: any): Promise<any> => {
             try {
+              // Check if the endpoint is rate-limited
+              if (await this.poolStateManager.getFieldValue(endpoint, 'rateLimited')) {
+                const rateLimitDelay = await this.poolStateManager.getFieldValue(endpoint, 'rateLimitDelay');
+                logger.info(`throtling on ratelimited endpoint ${rateLimitDelay / 1000}s`);
+                await delay(rateLimitDelay / 1000);
+              }
+
               const start = Date.now();
               const result = await target.fetchBlocks(heights, ...args);
               const end = Date.now();
--- a/packages/node-core/src/indexer/connectionPool.service.ts
+++ b/packages/node-core/src/indexer/connectionPool.service.ts
@@ -104,9 +104,9 @@ export class ConnectionPoolService<T extends IApiConnectionSpecific<any, any, an
             try {
               // Check if the endpoint is rate-limited
               if (await this.poolStateManager.getFieldValue(endpoint, 'rateLimited')) {
-                logger.info('throtling on ratelimited endpoint');
-                const backoffDelay = await this.poolStateManager.getFieldValue(endpoint, 'backoffDelay');
-                await delay(backoffDelay / 1000);
+                const rateLimitDelay = await this.poolStateManager.getFieldValue(endpoint, 'rateLimitDelay');
+                logger.info(`throtling on ratelimited endpoint ${rateLimitDelay / 1000}s`);
+                await delay(rateLimitDelay / 1000);
               }
 
               const start = Date.now();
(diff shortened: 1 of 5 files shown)


stwiname (commented):
I don't really see how this solves any issues. From what I can tell all that has changed is that if the error is a rate limit error then it uses 20s rather than exponential backoff. 

The rate limit delay is clearly way too large, (120s, 240s, 480s, 960s, 1920s). I think we would be better changing RETRY_DELAY to 10s (20, 40, 80, 160, 320 ).

packages/node-core/src/indexer/connectionPool.service.ts
```diff
@@ -104,9 +104,9 @@ export class ConnectionPoolService<T extends IApiConnectionSpecific<any, any, an
             try {
               // Check if the endpoint is rate-limited
               if (await this.poolStateManager.getFieldValue(endpoint, 'rateLimited')) {
-                logger.info('throtling on ratelimited endpoint');
-                const backoffDelay = await this.poolStateManager.getFieldValue(endpoint, 'backoffDelay');
-                await delay(backoffDelay / 1000);
+                const rateLimitDelay = await this.poolStateManager.getFieldValue(endpoint, 'rateLimitDelay');
+                logger.info(`throtling on ratelimited endpoint ${rateLimitDelay / 1000}s`);
+                await delay(rateLimitDelay / 1000);
```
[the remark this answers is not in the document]
yoozo: The backoffDelay parameter is mainly used when setting setRecoverTimeout, and there will be an interval function that displays the current endpoint status (function logEndpointStatus).

packages/node-core/src/indexer/connectionPoolState.manager.ts
```diff
@@ -247,35 +252,42 @@ export class ConnectionPoolStateManager<T extends IApiConnectionSpecific<any, an
...
+
+    const nextDelay = this.calculateNextDelay(this.pool[endpoint]);
+    this.pool[endpoint].backoffDelay = nextDelay;
+    await this.setRecoverTimeout(endpoint, nextDelay);
+
+    logger.warn(
+      `Endpoint ${this.pool[endpoint].endpoint} experienced an error (${errorType}). Suspending for ${
+        nextDelay / 1000
+      }s.`
+    );
```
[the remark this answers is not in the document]
yoozo: The original code logic cannot properly handle timeout and rate limit errors. From the context, it seems the goal is to make the endpoint available again after the backoffDelay. So this part of the logic should be moved outside.

yoozo:
I agree with your point about the rate limit delay being too large.  I believe the purposes of backoffDelay and rateLimitDelay are different. backoffDelay is used to mark an endpoint as unhealthy for a period of time, during which we will not select this node for RPC requests. rateLimitDelay is only used in extreme cases when all endpoints are unavailable (all have been set with backoffDelay). At this point, we should set a fixed time (20 seconds) for the request, as this duration might be long enough.

stwiname:
> I agree with your point about the rate limit delay being too large. I believe the purposes of backoffDelay and rateLimitDelay are different. backoffDelay is used to mark an endpoint as unhealthy for a period of time, during which we will not select this node for RPC requests. rateLimitDelay is only used in extreme cases when all endpoints are unavailable (all have been set with backoffDelay). At this point, we should set a fixed time (20 seconds) for the request, as this duration might be long enough.

I don't think we need to have different delays for that, we can see why endpoints are being throttled and can use selection based on that. i.e. If we see endpoints are being rate limited and we still need an endpoint to use we can ignore the delay.

packages/node-core/src/indexer/connectionPoolState.manager.ts
```diff
@@ -247,35 +247,49 @@ export class ConnectionPoolStateManager<T extends IApiConnectionSpecific<any, an
...
+    const delayRule = [10, 20, 40, 80, 160, 320];
+    let delayTime = 0;
+    if (poolItem.failureCount < 1) {
+      delayTime = 0;
+    } else if (poolItem.failureCount >= 1 && poolItem.failureCount <= delayRule.length) {
+      delayTime = delayRule[poolItem.failureCount - 1];
+    } else {
+      delayTime = delayRule[delayRule.length - 1];
+    }
+    return delayTime * 1000;
```
stwiname, line 292: This can use the old logic, Just change RETRY_DELAY to 10*1000

packages/node-core/src/indexer/connectionPool.service.ts
```diff
@@ -102,13 +102,6 @@ export class ConnectionPoolService<T extends IApiConnectionSpecific<any, any, an
         if (prop === 'fetchBlocks') {
           return async (heights: number[], ...args: any): Promise<any> => {
             try {
-              // Check if the endpoint is rate-limited
```
stwiname, line 105: This is still necessary or there will be no throttling

packages/node-core/src/indexer/connectionPoolState.manager.ts
```diff
@@ -191,13 +192,13 @@ export class ConnectionPoolStateManager<T extends IApiConnectionSpecific<any, an
...
   //eslint-disable-next-line @typescript-eslint/require-await
-  async setTimeout(endpoint: string, delay: number): Promise<void> {
+  async setRecoverTimeout(endpoint: string, delay: number): Promise<void> {
     // Make sure there is no existing timeout
     await this.clearTimeout(endpoint);
 
     this.pool[endpoint].timeoutId = setTimeout(() => {
       this.pool[endpoint].backoffDelay = 0; // Reset backoff delay only if there are no consecutive errors
-      this.pool[endpoint].rateLimited = false;
+      // this.pool[endpoint].rateLimited = false;   // Do not reset rateLimited status
```
stwiname, line 201: Do you think here we need to set a new timeout of the same value so it stayed rate limited?

yoozo: I think adding this here doesn’t avoid the problem; it just delays its appearance.

stwiname (approved):

PR #2598, opened 2024-11-22
[release] 20241122

yoozo:
# Description

Fixes # (issue)

## Type of change

## Checklist

- [ ] I have tested locally

### DIFF
--- a/packages/common/package.json
+++ b/packages/common/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@subql/common",
-  "version": "5.1.5",
+  "version": "5.2.0",
   "description": "",
   "scripts": {
     "build": "rm -rf dist && tsc -b",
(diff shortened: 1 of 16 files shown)


packages/common/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@subql/common",
-  "version": "5.1.5-1",
+  "version": "5.1.5",
```
stwiname, line 3: I think this needs to be more than a patch, removing cockroach will not be compatible

packages/node-core/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@subql/node-core",
-  "version": "14.1.8-3",
+  "version": "14.2.0",
```
stwiname, line 3: There are breaking changes so this should be major

stwiname (approved):

PR #2611, opened 2024-11-28
Implement dynamic block fetching based on block size

stwiname:
# Description
Ditches the smart batch size implementation that wasnt working because it didn't correctly calculate block sizes. This new method requires each chain to implement a block size function which is then used to calculate a median size based on the last 1000 indexed blocks. This data is then used to adjust the concurrency of fetching blocks. It will also start indexing with a low concurrency then increase as it has more data to work with. 

This is designed to much better deal with large blocks as it will reduce memory usage and requesting large amounts of data at once over the network. It is not a perfect solution but should be a good step forward. In cases where large blocks are not dealt with a restart and low  startup concurrency should avoid getting stuck indexing.

## TODO
- [ ] Update for worker threads 

Fixes # (issue)

## Type of change

## Checklist

- [x] I have tested locally


packages/node-core/src/utils/graphql.ts
```diff
@@ -13,10 +13,14 @@ import {
...
   GraphQLEntityField,
   GraphQLJsonFieldType,
+  blake2AsHex,
 } from '@subql/utils';
 import {ModelAttributes, ModelAttributeColumnOptions} from '@subql/x-sequelize';
 import {isArray, isObject} from 'lodash';
-import {enumNameToHash} from '../db';
+
+export function enumNameToHash(enumName: string): string {
+  return blake2AsHex(enumName).substr(2, 10);
```
yoozo, line 22: String.substr has been officially marked as not deprecated.

packages/node-core/src/indexer/blockDispatcher/block-dispatcher.ts
```diff
@@ -12,11 +11,12 @@ import {getBlockHeight, IBlock, PoiSyncService} from '../../indexer';
 import {getLogger} from '../../logger';
 import {exitWithError, monitorWrite} from '../../process';
 import {profilerWrap} from '../../profiler';
-import {Queue, AutoQueue, delay, memoryLock, waitForBatchSize, isTaskFlushedError} from '../../utils';
+import {Queue, AutoQueue, RampQueue, delay, isTaskFlushedError} from '../../utils';
 import {StoreService} from '../store.service';
 import {IStoreModelProvider} from '../storeModelProvider';
 import {IProjectService, ISubqueryProject} from '../types';
 import {BaseBlockDispatcher, ProcessBlockResponse} from './base-block-dispatcher';
+// import { RampQueue } from '@subql/node-core/utils/rampQueue';
```
yoozo, line 19: Can remove it

packages/node-core/src/indexer/blockDispatcher/worker-block-dispatcher.ts
```diff
@@ -115,7 +113,7 @@ export abstract class WorkerBlockDispatcher<DS, W extends Worker, B>
       let startIndex = 0;
       while (startIndex < heights.length) {
```
yoozo, line 114: You can remove this loop.

packages/node-core/src/utils/queues/rampQueue.ts
```diff
@@ -0,0 +1,92 @@
...
+      this.addSize(size);
+
+      // Not enough data to construct a median
+      if (this.#sizes.length < MIN_SIZES) {
+        return;
+      }
+
+      if (size > m * 2) {
+        // Inverse of the size compared to the median. E.g if a block is 5x as big as the median then the batch size should be 1/5 of the max
+        const multiplier = 1 / (size / m);
```
yoozo, line 75: I think writing it this way makes it easier to understand. It means scaling down to a smaller concurrency proportionally.
const concurrency = this.#maxConcurrency * (m / size)

packages/node-core/src/utils/queues/rampQueue.ts
```diff
@@ -0,0 +1,92 @@
...
+        this.setConcurrency(this.#maxConcurrency * multiplier);
+      } else if (this.#sizes.length % MIN_SIZES === 0) {
```
yoozo, line 77: When this.#sizes is filled, this condition will always be true.

packages/node-core/src/utils/queues/rampQueue.ts
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  async put(item: Task<T>): Promise<T> {
+    return this.putMany([item])[0];
+  }
+
+  putMany(tasks: Task<T>[]): Promise<T>[] {
+    return super.putMany(tasks).map((r) =>
+      r.then((d) => {
+        this.adjustConcurrency(d);
```
yoozo, line 45: According to this approach, we will only affect the last processed block.

packages/node-core/src/utils/graphql.ts
stwiname: I've just moved this function to fix an issue with circular imports, I don't want to be modifying those things as it's out of scope

packages/node-core/src/indexer/worker/worker.ts
```diff
@@ -56,10 +55,6 @@ export function getWorkerService<S extends typeof workerService>(): S {
   assert(workerService, 'Worker Not initialised');
   return workerService as S;
 }
-// eslint-disable-next-line @typescript-eslint/require-await
-async function getBlocksLoaded(): Promise<number> {
-  return workerService.numFetchedBlocks + workerService.numFetchingBlocks;
-}
 
 // eslint-disable-next-line @typescript-eslint/require-await
 async function getMemoryLeft(): Promise<number> {
```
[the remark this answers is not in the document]
stwiname: They all need to be async because the worker interface requires all methods to be async. This code is untouched anyway

packages/node-core/src/indexer/blockDispatcher/worker-block-dispatcher.ts
stwiname: Why? This code doesn't relate to the PR or changes

yoozo: Because after the modification, it now only executes the loop once.

yoozo (approved):

PR #2613, opened 2024-11-29
Fixed the issue where flags that require special handling were being overwritten.

### LINKED ISSUE #2612: [Bug] Crush when setting --primary-network-endpoint
### Prerequisites

* [x] Are you running the latest version(s)?
* [x] Have you searched the [documentation for your issue](https://academy.subquery.network/)?
* [x] Did you search for an existing issue or pull request?

### Description

Run with `--primary-network-endpoint=xxxx` will cause `Invalid endpoint: x`

yoozo:
# Description
When the `primary-network-endpoint` flag exists, it will report an "Invalid endpoint" error.

After yargs parses the flags, there will be two types of fields: snake_case and camelCase. Therefore, we currently need to optimize the logic for handling camelCase fields.

Fixes #2612

## Type of change

## Checklist

- [x] I have tested locally

### DIFF
--- a/packages/node-core/CHANGELOG.md
+++ b/packages/node-core/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Fixed
-- When the `primary-network-endpoint` flag exists, it will report an "Invalid endpoint" error.
+- When the `primary-network-endpoint` flag exists, it will report an "Invalid endpoint" error.(#2612)
 
 ## [15.0.3] - 2024-11-26
 ### Fixed
--- a/packages/node-core/CHANGELOG.md
+++ b/packages/node-core/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Fixed
-- When the `primary-network-endpoint` flag exists, it will report an "Invalid endpoint" error.(#2612)
+- Fixed the issue where flags that require special handling were being overwritten.(#2612)
 
 ## [15.0.3] - 2024-11-26
 ### Fixed
--- a/packages/node-core/CHANGELOG.md
+++ b/packages/node-core/CHANGELOG.md
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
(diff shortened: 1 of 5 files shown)


packages/node-core/src/configure/configure.module.spec.ts
```diff
@@ -93,7 +93,9 @@ describe('Configure', () => {
       const option = {headers: {'api-key': '<your-api-key>'}};
       const {primaryNetworkEndpoint} = yargsToIConfig({
         'primary-network-endpoint': 'https://example.com',
+        primaryNetworkEndpoint: 'https://example.com',
```
stwiname, line 96: I don't think this is a sufficient test. The issue was the snake case being overwritten not support for snakeCase

packages/node-core/CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- When the `primary-network-endpoint` flag exists, it will report an "Invalid endpoint" error.(#2612)
```
stwiname, line 10: This doesn't clearly explain what was fixed, the issue wasn't related to this flag it was related to flags being overwritten and it would have been the same issue for any flags with special cases

packages/node-core/src/configure/configure.module.ts
```diff
@@ -52,7 +52,7 @@ export function yargsToIConfig(yargs: Args, nameMapping: Record<string, string>
   return Object.entries(yargs).reduce((acc, [key, value]) => {
     if (['_', '$0'].includes(key)) return acc;
 
-    if (key === 'network-registry') {
+    if (isMatchFlag(key, ['network-registry'])) {
```
stwiname, line 55: This fix is prone to the same error occuring again. If another flag is added and there is now implicit behaviour as to what this does the same issue could be introduced again with another flag. 

I think it would be better to not compare using key and instead use the modified key (`nameMapping[key] ?? camelCase(key)`)

stwiname:
Can you please update the PR description and name to describe the fix/cause not the symptom of the bug. I had to debug this myself because there is no explanation and the changes didn't make sense to how it could fix the bug

packages/node-core/src/configure/configure.module.spec.ts
yoozo: The fields we get from yargs will include both snake_case and camelCase.

stwiname (approved):

PR #2614, opened 2024-12-01
Fix interval with 0 time causing high CPU

stwiname:
# Description
Fixes the keep alive interval not being passed correctly causing an interval with 0ms time, and therefore high CPU.

Fixes # (issue)

## Type of change

## Checklist

- [x] I have tested locally

### DIFF
--- a/packages/query/src/graphql/graphql.module.ts
+++ b/packages/query/src/graphql/graphql.module.ts
@@ -118,7 +118,7 @@ export class GraphqlModule implements OnModuleInit, OnModuleDestroy {
   }
 
   private setupKeepAlive(pgClient: PoolClient) {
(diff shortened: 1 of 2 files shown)


packages/query/src/graphql/graphql.module.ts
```diff
@@ -118,6 +118,8 @@ export class GraphqlModule implements OnModuleInit, OnModuleDestroy {
   }
 
   private setupKeepAlive(pgClient: PoolClient) {
+    const interval = argv['sl-keep-alive-interval'] ?? 180000;
```
yoozo, line 121: Here we should use `||`

yoozo (approved):

Issue #2562, opened 2024-09-24
Alternative ID types


stwiname:
### Background

Currently all entities need to have `id: ID` to follow the graphql spec. Internally with postgres this creates a TEXT primary key column. This can have some limitations when using numbers as the id and ordering because it will order lexicographically.

### Solution

We can add an additional directive to the graphql schema to allow specifying a different underlying type.

```graphql
type Entity @entity {
  id: ID @dbType(type: "serial")
}
```

### Requirements

* This should still be compatible with historical indexing
* Limit the type to compatible types. 
* Compatible with schema migrations
* Update the store interface for Id to be optional when creating new entities and return the generated id. This could be pushed to a future release.

jorgecuesta:
This is something we are looking forward for our project at Pocket Network https://github.com/pokt-network/pocketdex

PR #2622, opened 2024-12-10
Add support for alternative ID types

stwiname:
# Description
Allows specifying a different ID type other than text. It can be 'BigInt', 'Int', 'Float', 'ID' or 'String'. This is useful when you want to order by ID but not in a lexicographical order. 

In order to avoid growing this PR and the complexity of types, codegen has been updated to generate some `Compat` types to work with internal types that expect the `id` to be a string.

Fixes https://github.com/subquery/subql/issues/2562
## Type of change

- [x] This change requires a documentation update https://github.com/subquery/documentation/pull/577

## Checklist

- [x] I have tested locally
- [x] Updated any relevant documentation https://github.com/subquery/documentation/pull/577

### DIFF
--- a/packages/cli/src/template/model.ts.ejs
+++ b/packages/cli/src/template/model.ts.ejs
@@ -13,7 +13,13 @@ import {<% props.importEnums.forEach(function(e){ %>
 
 export type <%= props.className %>Props = Omit<<%=props.className %>, NonNullable<FunctionPropertyNames<<%=props.className %>>> | '_name'>;
 
-export class <%= props.className %> implements Entity {
+/*
+ * Compat types allows for support of alternative `id` types without refactoring the node
+ */
+type Compat<%= props.className %>Props = Omit<<%= props.className %>Props, 'id'> & { id: string; };
+type CompatEntity = Omit<Entity, 'id'> & { id: <%=props.idType %>; };
+
+export class <%= props.className %> implements CompatEntity {
 
     constructor(
         <% props.fields.forEach(function(field) { if (field.required) { %>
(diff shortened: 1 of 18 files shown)


packages/cli/src/template/model.ts.ejs
```diff
@@ -56,14 +62,14 @@ export class <%= props.className %> implements Entity {
     static async getBy<%=helper.upperFirst(field.name) %>(<%=field.name %>: <%=field.type %>): Promise<<%=props.className %> | undefined> {
         const record = await store.getOneByField('<%=props.entityName %>', '<%=field.name %>', <%=field.name %>);
         if (record) {
-            return this.create(record as <%= props.className %>Props);
+            return this.create(record as unknown <%= props.className %>Props);
```
yoozo, line 65: Missing one as

packages/cli/src/template/model.ts.ejs
```diff
@@ -31,21 +37,21 @@ export class <%= props.className %> implements Entity {
     }
 
     async save(): Promise<void> {
-        let id = this.id;
+        const id = this.id;
         assert(id !== null, "Cannot save <%=props.className %> entity without an ID");
-        await store.set('<%=props.entityName %>', id.toString(), this);
+        await store.set('<%=props.entityName %>', id.toString(), this as Compat<%=props.className %>Props);
```
yoozo, line 42: The type issue of the id field still needs to be addressed here.

` {...this, id: id.toString() } as CompatTransferProps`

yoozo (commented):
There will be a type error when using the `store.bulkCreate` method because it requires the `entity.id` type to be a string.

packages/utils/src/graphql/graphql.spec.ts
```diff
@@ -1,6 +1,7 @@
 // Copyright 2020-2024 SubQuery Pte Ltd authors & contributors
 // SPDX-License-Identifier: GPL-3.0
 
+import {buildSchema} from 'graphql';
```
yoozo, line 4: can remove it

packages/cli/src/template/model.ts.ejs
stwiname: Functionally this is fine, its just the types we're getting around. Its better to not copy the data if we can avoid it

stwiname:
> There will be a type error when using the `store.bulkCreate` method because it requires the `entity.id` type to be a string.

I think that is fine. People don't generally have much use for `bulkCreate` and if the cache is enabled there is no performance benefit. I want to avoid going down the rabbit hole of trying to update all the types.

yoozo (approved):

PR #2626, opened 2024-12-11
fix: subscriptions payloads _entity w/ relations fields

IcanDivideBy0:
# Description
With recent changes to subscriptions, the `_entity` field is now graphql-typed, but the said field is resolved with raw data from database, hence all camelCase named fields in the gql schema are not properly resolved. This PR solves this issue.

Nonetheless all requested relations of the `_entity` are still not properly resolved... I'll try working on that now... if you have any advice regarding this, that'd be great :)

## Type of change

## Checklist

- [x] I have tested locally

### DIFF
--- a/packages/node-core/CHANGELOG.md
+++ b/packages/node-core/CHANGELOG.md
@@ -5,6 +5,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+- Updated send_notification PG function to include `_block height` and entity `_id` (#2626)
 
 ## [16.1.0] - 2024-12-11
 ### Changed
--- a/packages/node-core/src/db/sync-helper.test.ts
+++ b/packages/node-core/src/db/sync-helper.test.ts
@@ -172,15 +172,15 @@ describe('sync helper test', () => {
       // For that reason the behaviour is kept the same as before delete was fixed.
       expect(listener).toHaveBeenNthCalledWith(
         1,
-        `{"id": "1", "_entity": {"id": "1", "block_number": 1}, "mutation_type": "UPDATE"}`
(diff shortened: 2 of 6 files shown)


stwiname (commented):
Im happy with this change. Can you please update the changelog too.

For getting relations working, i think the goal would be to take the `id` from the payload and use that to run a graphql query and not use anything from the `_entity`. I don't have much direction to give here, postgraphile is not trivial to work with. But the [realtime](https://www.graphile.org/postgraphile/realtime/) feature might be a good reference as to how this could be implemented.

IcanDivideBy0:
I've updated the changelog.

I tried to get relations working too, from what I see in postgraphile documentation, I should be able to get it working with something like this in the resolver
```ts
        const [row] = await resolveInfo.graphile.selectGraphQLResultFromTable(
          sql.fragment`${table.namespace.name}.${table.name}`,
          (tableAlias, queryBuilder) => {
            queryBuilder.where(
              sql.fragment`${tableAlias}.id = ${sql.value(parent.id)}`
            )
            queryBuilder.limit(1);
          }
        );

        return row;
```
unfortunately i always end up with an error `Cannot destructure property 'pgClient' of 'context' as it is undefined.` ... for some reson the context is always undefined in subscriptions resolvers ... i'trie to update apollo server to use the latest `@apollo/server` but the error still there :/

IcanDivideBy0:
I'm thinking of removeing these two lines: https://github.com/subquery/subql/blob/main/packages/node-core/src/db/sync-helper.ts#L197-L198 and use the `_id` in the where clause and the `_block_height`for the query builder context. Do you agree on this ?

stwiname:
> I'm thinking of removeing these two lines: https://github.com/subquery/subql/blob/main/packages/node-core/src/db/sync-helper.ts#L197-L198 and use the `_id` in the where clause and the `_block_range.start`for the query builder context. Do you agree on this ?

I see no issue with that

IcanDivideBy0:
ready for a final review :)

stwiname (commented):
This looks great, can you please update the changelog for node-core then this can be merged.

IcanDivideBy0:
changelog uptated :)

stwiname:
There is a test failing in `packages/node-core/src/db/sync-helper.test.ts` relating to the data structure changes on the notify. Can you please fix that

IcanDivideBy0:
Just updated the tests

PR #2644, opened 2025-01-16
Fix monitor oom

yoozo:
# Description
When encountering large blocks and using multiple workers, the monitor will experience memory overflow and thread blocking due to the need for thread communication and large object parsing. The main cause is that JSON parsing takes a long time, during which the main thread is still sending messages, leading to an OOM error.

The following optimizations were made:
 - Added the `--monitor-object-max-depth` flag to mitigate OOM issues when encountering large chunks.

Fixes # (issue)
 - Fixed the inconsistency between the `monitor-file-size` flag and the expected behavior.

## Type of change

## Checklist

- [x] I have tested locally

### DIFF
--- a/packages/node-core/src/configure/NodeConfig.ts
+++ b/packages/node-core/src/configure/NodeConfig.ts
@@ -57,6 +57,7 @@ export interface IConfig {
   readonly csvOutDir?: string;
   readonly monitorOutDir: string;
   readonly monitorFileSize?: number;
+  readonly monitorObjectMaxDepth: number;
   readonly enableCache?: boolean;
 }
 
@@ -86,6 +87,7 @@ const DEFAULT_CONFIG = {
   storeFlushInterval: 5,
   allowSchemaMigration: false,
   monitorOutDir: './.monitor',
+  monitorObjectMaxDepth: 0,
 };
 
 export class NodeConfig<C extends IConfig = IConfig> implements IConfig {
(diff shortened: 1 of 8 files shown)


packages/node-core/src/yargs.ts
```diff
@@ -266,6 +266,11 @@ export function yargsBuilder<
                 describe: 'monitor file size limit in MB ',
                 type: 'number',
               },
+              'monitor-object-max-depth': {
+                describe: 'The maximum depth of an object recorded by a monitor. Default is 0, no limit',
+                type: 'number',
+                default: 0,
```
stwiname, line 272:
```suggestion
```

The default is provided in the node config.

packages/node-core/src/configure/NodeConfig.ts
```diff
@@ -86,6 +87,7 @@ const DEFAULT_CONFIG = {
   storeFlushInterval: 5,
   allowSchemaMigration: false,
   monitorOutDir: './.monitor',
+  monitorObjectMaxDepth: 0,
```
stwiname, line 90: Is this the default value we want? It means that it will explicitly need to be set to avoid the OOM issues

packages/node-core/src/utils/string.spec.ts
```diff
@@ -58,3 +58,25 @@ describe('handledStringify', () => {
     expect(result).toBe('Error without stack');
   });
 });
+
+describe('handledStringify depth', () => {
+  it('Different truncation situations at various depths', () => {
```
stwiname, line 63: Can we please expand this to include:
* Null values
* Undefined values
* A circular reference.

packages/node-core/src/configure/NodeConfig.ts
yoozo: Maybe we need to set a reasonable default value? For example, 5.

packages/node-core/src/yargs.ts
```diff
@@ -266,6 +266,10 @@ export function yargsBuilder<
                 describe: 'monitor file size limit in MB ',
                 type: 'number',
               },
+              'monitor-object-max-depth': {
+                describe: 'The maximum depth of an object recorded by a monitor. Default is 0, no limit',
```
stwiname, line 270:
```suggestion
                describe: 'The maximum depth of an object recorded by a monitor. Default is 5, 0 is no limit',
```

PR #2658, opened 2025-01-30
[DRAFT] Fix typescript error avoid ID not matching when use @dbType

jorgecuesta:
# Description
Ensure that IDs match correctly even when data types like BigInt, Int, or Float are used in GraphQL schemas. This prevents mismatches and ensures error handling for invalid ID comparisons.

On Models template there are 2 types:
1. https://github.com/subquery/subql/blob/main/packages/cli/src/template/model.ts.ejs#L19
2. https://github.com/subquery/subql/blob/main/packages/cli/src/template/model.ts.ejs#L20

They both Omit the `id` property, but one leave as string, which if we change the ID with `@dbType()` maybe get an error?
For sure the `getByFields` has a typescript error due to that because it expect the T to extend Entity which expect id be an string.

## Type of change

## Checklist

- [x] I have tested locally


jorgecuesta:
This was fixed with #2662 by @stwiname

Issue #2620, opened 2024-12-05
[Feature] Multi chain rewind


yoozo:
### Description
In a multi-chain project, after enabling historical mode, a rewind requires coordination with other processes for rollback. Below are the tasks we need to accomplish.

### Scene
- Caused by a fork in the chain
- Project migrate
- Nodes only need to roll back if they have indexed > the timestamp of the rollback.

### TODO
- [x] cold start requires rollback execution.
- [x] A pause mechanism is needed where all fetch processes stop in case of a rollback.
- [x] A logger notification is needed when the API fetch is paused.
- [ ] Before performing the write, check if `_global.rewind_lock` exist. If they do, perform a rollback.
- [x] When starting the process, check if `rewind_timestamp` exists. If it does, perform a rewind.
- [x] Check if `_global` contain the `rewind_timestamp`. If none exist, release `_global.rewind_lock`.
- [ ] ***Retrieve the corresponding block height to roll back to using a `timestamp`, It may require using the binary search method.***

### Pending matters
- ***The RPC does not support querying block height by timestamp, but our current POI and dynamicDs rely on block height for rollback.***
- Only one node needs to rewind the data, but each node needs to rewind its own metadata (to find the block for that timestamp)

### Solution
The pause mechanism can be implemented using `_global.rewind_lock`, with priority locking during a rewind. Simultaneously write a `rewind_timestamp` record to `_global`, indicating the height that needs to be rolled back.
- `_global` table  ensure that each chain will execute rewind.
- `rewind_timestamp` ensure that the API fetch is paused until all chains have completed the rewind.

``` sql
CREATE TABLE "_global" (
  "key" varchar(255) NOT NULL,
  "value" jsonb,
  "createdAt" timestamptz(6) NOT NULL,
  "updatedAt" timestamptz(6) NOT NULL,
  CONSTRAINT "_global_pkey" PRIMARY KEY ("key")
);
```

When a rollback is required, the data in the global table looks like this.
| Key                       | Value                                                                 |
|---------------------------|-----------------------------------------------------------------------|
| rewind_lock               | $timestamp                                                            |
| rewind_timestamp_ethereum | $timestamp  |
| rewind_timestamp_base     | $timestamp  |

### issues

- How does the rewind lock get lifted?  
> After each process completes its reindex, it needs to check if there are other rewind tasks in `_global`. If none exist, it releases `_global.rewind_lock`.
- Is there polling to check the metadata tables? 
[...]

PR #2673, opened 2025-02-12
multi chain rewind service

yoozo:
# Description
In a multi-chain project, after enabling historical mode, a rewind requires coordination with other processes for rollback. Below are the tasks we need to accomplish.

Fixes # (issue)
#2620

### TODO
- [x] cold start requires rollback execution.
- [x] A pause mechanism is needed where all fetch processes stop in case of a rollback.
- [x] A logger notification is needed when the API fetch is paused.
- [x] Before performing the write, check if `_global.rewind_lock` exist. If they do, perform a rollback.
- [x] When starting the process, check if `rewind_timestamp` exists. If it does, perform a rewind.
- [x] Check if `_global` contain the `rewind_timestamp`. If none exist, release `_global.rewind_lock`.
- [x] ***Retrieve the corresponding block height to roll back to using a `timestamp`, It may require using the binary search method.***
- [x] Use a DB notify/trigger to notify a cooperating process.
- [ ] Extract global lock logic into a separate file.
- [ ] Remove the delay in fetch.service.
- [ ] Modify the lock handling logic to ensure correct release under concurrent conditions. (Add a rewindNum field to rewindLock to ensure the remaining processing count is correct when the lock is released.)
- [ ] When flushing the fetch queue, prevent the currently processing block from being persisted.
- [ ] Test cases for related scenarios.

## Type of change

## Checklist

- [ ] I have tested locally

### DIFF
--- a/packages/node-core/src/indexer/entities/GlobalData.entity.ts
+++ b/packages/node-core/src/indexer/entities/GlobalData.entity.ts
@@ -1,0 +1,53 @@
+// Copyright 2020-2024 SubQuery Pte Ltd authors & contributors
+// SPDX-License-Identifier: GPL-3.0
+
+import {blake2AsHex} from '@subql/utils';
+import {BuildOptions, DataTypes, Model, Sequelize} from '@subql/x-sequelize';
+
+export const RewindTimestampKeyPrefix = 'rewindTimestamp';
+export const RewindLockKey = 'rewindLock';
+export type RewindTimestampKey = `${typeof RewindTimestampKeyPrefix}_${string}`;
+
+export type RewindLockInfo = {
+  /** Timestamp to rewind to. */
+  timestamp: number;
+  /** Number of additional chains to rewind. */
+  rewindNum: number;
+};
+export interface GlobalDataKeys {
+  rewindLock: RewindLockInfo;
+  [key: RewindTimestampKey]: number;
+}
+
+export interface GlobalData<k extends keyof GlobalDataKeys = keyof GlobalDataKeys> {
+  key: k;
+  value: GlobalDataKeys[k];
+}
+
+interface GlobalDataEntity extends Model<GlobalData>, GlobalData {}
+
+export type GlobalDataRepo = typeof Model & {
+  new (values?: unknown, options?: BuildOptions): GlobalDataEntity;
+};
+
+export function GlobalDataFactory(sequelize: Sequelize, schema: string): GlobalDataRepo {
+  const tableName = '_global';
+
+  return <GlobalDataRepo>sequelize.define(
+    tableName,
+    {
+      key: {
+        type: DataTypes.STRING,
+        primaryKey: true,
+      },
+      value: {
+        type: DataTypes.JSONB,
+      },
+    },
+    {freezeTableName: true, schema: schema}
+  );
+}
+
+export function generateRewindTimestampKey(chainId: string): RewindTimestampKey {
+  return `${RewindTimestampKeyPrefix}_${blake2AsHex(chainId)})`.substring(0, 63) as RewindTimestampKey;
+}
--- a/packages/node-core/src/indexer/fetch.service.ts
+++ b/packages/node-core/src/indexer/fetch.service.ts
@@ -3,19 +3,20 @@
 
 import assert from 'assert';
 import {Inject, Injectable, OnApplicationShutdown} from '@nestjs/common';
-import {EventEmitter2} from '@nestjs/event-emitter';
+import {EventEmitter2, OnEvent} from '@nestjs/event-emitter';
 import {SchedulerRegistry} from '@nestjs/schedule';
 import {BaseDataSource} from '@subql/types-core';
 import {range} from 'lodash';
 import {IBlockchainService} from '../blockchain.service';
 import {NodeConfig} from '../configure';
-import {IndexerEvent} from '../events';
+import {EventPayload, IndexerEvent, MultiChainRewindEvent, MultiChainRewindPayload} from '../events';
 import {getLogger} from '../logger';
 import {delay, filterBypassBlocks, getModulos} from '../utils';
 import {IBlockDispatcher} from './blockDispatcher';
 import {mergeNumAndBlocksToNums} from './dictionary';
 import {DictionaryService} from './dictionary/dictionary.service';
 import {mergeNumAndBlocks} from './dictionary/utils';
+import {IMultiChainHandler, MultiChainRewindService, RewindStatus} from './multiChainRewind.service';
 import {IStoreModelProvider} from './storeModelProvider';
 import {BypassBlocks, IBlock, IProjectService} from './types';
 import {IUnfinalizedBlocksServiceUtil} from './unfinalizedBlocks.service';
@@ -24,7 +25,7 @@ const logger = getLogger('FetchService');
 
 @Injectable()
 export class FetchService<DS extends BaseDataSource, B extends IBlockDispatcher<FB>, FB>
-  implements OnApplicationShutdown
+  implements OnApplicationShutdown, IMultiChainHandler
 {
   private _latestBestHeight?: number;
   private _latestFinalizedHeight?: number;
@@ -39,7 +40,8 @@ export class FetchService<DS extends BaseDataSource, B extends IBlockDispatcher<
     private schedulerRegistry: SchedulerRegistry,
     @Inject('IUnfinalizedBlocksService') private unfinalizedBlocksService: IUnfinalizedBlocksServiceUtil,
     @Inject('IStoreModelProvider') private storeModelProvider: IStoreModelProvider,
-    @Inject('IBlockchainService') private blockchainSevice: IBlockchainService<DS>
+    @Inject('IBlockchainService') private blockchainSevice: IBlockchainService<DS>,
+    private multiChainRewindService: MultiChainRewindService
   ) {}
 
   private get latestBestHeight(): number {
@@ -195,6 +197,14 @@ export class FetchService<DS extends BaseDataSource, B extends IBlockDispatcher<
 
       // Update the target height, this happens here to stay in sync with the rest of indexing
       void this.storeModelProvider.metadata.set('targetHeight', latestHeight);
+
+      // If we're rewinding, we should wait until it's done
+      const multiChainStatus = this.multiChainRewindService.getStatus();
+      if (RewindStatus.Normal !== multiChainStatus) {
+        logger.info(`Wait for all chains to complete rewind, current chainId: ${this.multiChainRewindService.chainId}`);
+        await delay(3);
+        continue;
+      }
 
       // This could be latestBestHeight, dictionary should never include finalized blocks
       // TODO add buffer so dictionary not used when project synced
@@ -382,5 +392,12 @@ export class FetchService<DS extends BaseDataSource, B extends IBlockDispatcher<
   resetForIncorrectBestBlock(blockHeight: number): void {
     this.updateDictionary();
     this.blockDispatcher.flushQueue(blockHeight);
+  }
+
+  @OnEvent(MultiChainRewindEvent.Rewind)
+  @OnEvent(MultiChainRewindEvent.RewindTimestampDecreased)
+  handleMultiChainRewindEvent(payload: MultiChainRewindPayload) {
+    logger.info(`Received rewind event, height: ${payload.height}`);
+    this.resetForNewDs(payload.height);
   }
 }
--- a/packages/node-core/src/indexer/multiChainRewind.service.ts
+++ b/packages/node-core/src/indexer/multiChainRewind.service.ts
@@ -1,0 +1,339 @@
+// Copyright 2020-2025 SubQuery Pte Ltd authors & contributors
+// SPDX-License-Identifier: GPL-3.0
+
+import assert from 'assert';
+import {Inject, Injectable, OnApplicationShutdown} from '@nestjs/common';
+import {EventEmitter2} from '@nestjs/event-emitter';
+import {hashName} from '@subql/utils';
+import {Transaction, Op, QueryTypes, Sequelize} from '@subql/x-sequelize';
+import dayjs from 'dayjs';
+import {Pool, PoolClient} from 'pg';
+import {IBlockchainService} from '../blockchain.service';
+import {NodeConfig} from '../configure';
+import {getPgPoolConfig} from '../db';
+import {MultiChainRewindEvent, MultiChainRewindPayload} from '../events';
+import {getLogger} from '../logger';
+import {
+  generateRewindTimestampKey,
+  GlobalData,
+  GlobalDataKeys,
+  RewindLockInfo,
+  RewindLockKey,
+  RewindTimestampKey,
+  RewindTimestampKeyPrefix,
+} from './entities';
+import {StoreService} from './store.service';
+import {Header} from './types';
+
+const logger = getLogger('RewindService');
+
+export enum RewindStatus {
+  /** The current chain is in normal state. */
+  Normal = 'normal',
+  /** The current chain is waiting for other chains to rewind. */
+  WaitOtherChain = 'waitOtherChain',
+  /** The current chain is executing rewind. */
+  Rewinding = 'rewinding',
+}
+interface IMultiChainRewindService {
+  chainId: string;
+  dbSchema: string;
+  getStatus(): RewindStatus;
+  getGlobalRewindStatus(): Promise<{
+    rewindTimestamp: GlobalDataKeys[RewindTimestampKey];
+    rewindLock?: GlobalDataKeys[typeof RewindLockKey];
+  }>;
+  setGlobalRewindLock(rewindTimestamp: number): Promise<void>;
+  /**
+   * Check if the height is consistent before unlocking.
+   * @param tx
+   * @param rewindTimestamp The timestamp to roll back to, in milliseconds.
+   * @returns the number of remaining rewind chains
+   */
+  releaseChainRewindLock(tx: Transaction, rewindTimestamp: number): Promise<number>;
+}
+
+export interface IMultiChainHandler {
+  handleMultiChainRewindEvent(rewindBlockPayload: MultiChainRewindPayload): void;
+}
+
+@Injectable()
+export class MultiChainRewindService implements IMultiChainRewindService, OnApplicationShutdown {
+  private _status: RewindStatus = RewindStatus.Normal;
+  private _chainId?: string;
+  private _dbSchema?: string;
+  waitRewindHeader?: Header;
+  private pgListener?: PoolClient;
+  constructor(
+    private nodeConfig: NodeConfig,
+    private eventEmitter: EventEmitter2,
+    private sequelize: Sequelize,
+    private storeService: StoreService,
+    @Inject('IBlockchainService') private readonly blockchainService: IBlockchainService
+  ) {}
+
+  private set chainId(chainId: string) {
+    this._chainId = chainId;
+  }
+
+  get chainId(): string {
+    assert(this._chainId, 'chainId is not set');
+    return this._chainId;
+  }
+
+  get dbSchema(): string {
+    assert(this._dbSchema, 'dbSchema is not set');
+    return this._dbSchema;
+  }
+
+  private set dbSchema(dbSchema: string) {
+    this._dbSchema = dbSchema;
+  }
+
+  private set status(status: RewindStatus) {
+    this._status = status;
+  }
+
+  getStatus(): RewindStatus {
+    return this._status;
+  }
+
+  onApplicationShutdown() {
+    this.pgListener?.release();
+  }
+
+  async init(chainId: string, dbSchema: string, reindex: (targetHeader: Header) => Promise<void>) {
+    this.chainId = chainId;
+    this.dbSchema = dbSchema;
+
+    if (this.storeService.historical === 'timestamp') {
+      // Register a listener and create a schema notification sending function.
+      await this.registerPgListener();
+
+      if (this.waitRewindHeader) {
+        const rewindHeader = {...this.waitRewindHeader};
+        await reindex(rewindHeader);
+        return rewindHeader;
+      }
+    }
+  }
+
+  private async registerPgListener() {
+    if (this.pgListener) return;
+
+    // Creating a new pgClient is to avoid using the same database connection as the block scheduler,
+    // which may prevent real-time listening to rollback events.
+    const pgPool = new Pool(getPgPoolConfig(this.nodeConfig));
+    this.pgListener = await pgPool.connect();
+
+    this.pgListener.on('notification', (msg) => {
+      Promise.resolve().then(async () => {
+        const eventType = msg.payload;
+        logger.info(`Received rewind event: ${eventType}, chainId: ${this.chainId}`);
+        switch (eventType) {
+          case MultiChainRewindEvent.Rewind:
+          case MultiChainRewindEvent.RewindTimestampDecreased: {
+            this.status = RewindStatus.Rewinding;
+            const {rewindTimestamp} = await this.getGlobalRewindStatus();
+            this.waitRewindHeader = await this.getHeaderByBinarySearch(dayjs(rewindTimestamp).toDate());
+
+            // Trigger the rewind event, and let the fetchService listen to the message and handle the queueFlush.
+            this.eventEmitter.emit(eventType, {
+              height: this.waitRewindHeader.blockHeight,
+            } satisfies MultiChainRewindPayload);
+            break;
+          }
+          case MultiChainRewindEvent.RewindComplete:
+            // recover indexing status
+            this.status = RewindStatus.Normal;
+            this.waitRewindHeader = undefined;
+            break;
+          default:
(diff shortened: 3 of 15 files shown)


packages/node-core/src/indexer/multiChainRewind.service.ts
```diff
@@ -0,0 +1,367 @@
...
+    const globalTable = this.storeService.globalDataRepo.tableName;
+    const chainTotal = await this.storeService.globalDataRepo.count({
+      where: {
+        key: {[Op.like]: `${RewindTimestampKeyPrefix}_%`},
+      },
+    });
+
+    const tx = await this.sequelize.transaction();
+    try {
+      const [_, updateRows] = await this.sequelize.query(
```
stwiname, line 234: I think these global table db operations can be moved into a PlainGlobalModel. It would be good to have all db operations in a single location

stwiname: Could this query string be generalised between metadata?

packages/node-core/src/utils/reindex.ts
```diff
@@ -49,10 +51,12 @@ export async function reindex(
   dynamicDsService: DynamicDsService<any>,
   sequelize: Sequelize,
   projectUpgradeService: IProjectUpgradeService<ISubqueryProject>,
+  multichainRewindService: MultiChainRewindService,
   poiService?: PoiService,
   forceCleanService?: ForceCleanService
 ): Promise<void> {
-  const lastUnit = storeService.historical === 'timestamp' ? lastProcessed.timestamp : lastProcessed.height;
+  const isMultiChain = storeService.historical === 'timestamp';
```
stwiname, line 58: Same as the other comment, historical mode doesn't indicate that its multichain

packages/node-core/src/indexer/multiChainRewind.service.ts
```diff
@@ -0,0 +1,367 @@
...
+            } satisfies MultiChainRewindPayload);
+            break;
+          }
+          case MultiChainRewindEvent.RewindComplete:
+            // recover indexing status
+            this.status = RewindStatus.Normal;
+            this.waitRewindHeader = undefined;
+            break;
+          default:
+            throw new Error(`Unknown rewind event: ${eventType}`);
```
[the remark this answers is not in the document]
yoozo: For the current listener, I only listen to the messages sent by rewindTrigger.

packages/node-core/src/blockchain.service.ts
```diff
@@ -58,6 +58,7 @@ export interface IBlockchainService<
   // Unfinalized blocks
   getHeaderForHash(hash: string): Promise<Header>;
   getHeaderForHeight(height: number): Promise<Header>;
+  getRequiredHeaderForHeight(height: number): Promise<Header & {timestamp: Date}>;
```
stwiname, line 61: I think this can be removed. Lets make timestamp required on Header. There's only one chain that we cant get the timestamp and its not worth making compromises just for that

packages/node-core/src/indexer/fetch.service.ts
```diff
@@ -196,6 +198,23 @@ export class FetchService<DS extends BaseDataSource, B extends IBlockDispatcher<
       // Update the target height, this happens here to stay in sync with the rest of indexing
       void this.storeModelProvider.metadata.set('targetHeight', latestHeight);
 
+      // If we're rewinding, we should wait until it's done
+      const multiChainStatus = this.multiChainRewindService.status;
+
+      if (RewindStatus.Rewinding === multiChainStatus) {
+        assert(this.multiChainRewindService.waitRewindHeader, 'Multi chain Rewind header is not set');
```
stwiname, line 205: I think it would be worth combining `status` and `waitRewindHeader` into a single response. It would clarify the usage and makes the multichain rewind service ensure that the header is available. 

There is no clear definition here as to how the fetch service would recover from `waitRewindHeader` being undefined.

packages/node-core/src/indexer/multiChainRewind.service.ts
```diff
@@ -0,0 +1,272 @@
...
+  }
+
+  private get dbSchema(): string {
+    assert(this._dbSchema, 'dbSchema is not set');
+    return this._dbSchema;
+  }
+
+  private set dbSchema(dbSchema: string) {
+    this._dbSchema = dbSchema;
+  }
```
stwiname, line 88: Im not sure why private getters and setters are needed. Can't there just be a private variable?

stwiname: Same for `rewindTriggerName`

packages/node-core/src/indexer/multiChainRewind.service.ts
```diff
@@ -0,0 +1,272 @@
...
+      this.status = RewindStatus.WaitOtherChain;
+    }
+    if (rewindTimestamp) {
+      this.status = RewindStatus.Rewinding;
+      this.waitRewindHeader = await this.searchWaitRewindHeader(rewindTimestamp);
+    }
+  }
+
+  private async searchWaitRewindHeader(rewindTimestamp: number): Promise<Header> {
+    const rewindDate = dayjs(rewindTimestamp).toDate();
```
stwiname, line 196: Why not change the argument to a date?

packages/node-core/src/indexer/storeModelProvider/global/global.ts
```diff
@@ -0,0 +1,187 @@
...
+
+    const records = await this.model.findAll({
+      where: {key: {[Op.in]: [rewindTimestampKey, RewindLockKey]}},
+    });
+    const rewindLockInfo: GlobalData<typeof RewindLockKey> | undefined = records
+      .find((r) => r.key === RewindLockKey)
+      ?.toJSON();
+    const rewindTimestampInfo: GlobalData<RewindTimestampKey> | undefined = records
+      .find((r) => r.key === rewindTimestampKey)
+      ?.toJSON();
```
stwiname, line 74: I think you could do `toJSON` on `records` then use a destructor.
Also `toJSON` is generic so you can pass the type in there.

packages/node-core/src/indexer/entities/GlobalData.entity.ts
```diff
@@ -0,0 +1,52 @@
...
+  return <GlobalDataRepo>sequelize.define(
+    tableName,
+    {
+      key: {
+        type: DataTypes.STRING,
+        primaryKey: true,
+      },
+      value: {
+        type: DataTypes.JSONB,
+      },
```
stwiname, line 44: Whats the purpose of this being JSONB? Looking at the rest of the code it would simplify the queries to have well defined columns

packages/node-core/src/indexer/multiChainRewind.service.ts
```diff
@@ -0,0 +1,272 @@
...
+    // Check whether the current state is in rollback.
+    // If a global lock situation occurs, prioritize setting it to the WaitOtherChain state. If a rollback is still required, then set it to the rewinding state.
+    const {rewindLock, rewindTimestamp} = await this.globalModel.getGlobalRewindStatus();
+    if (rewindLock) {
+      this.status = RewindStatus.WaitOtherChain;
+    }
+    if (rewindTimestamp) {
+      this.status = RewindStatus.Rewinding;
+      this.waitRewindHeader = await this.searchWaitRewindHeader(rewindTimestamp);
+    }
```
stwiname, line 192: This is implicit behaviour of this function. It's probably best to move this to its own function

packages/node-core/src/indexer/multiChainRewind.service.ts
```diff
@@ -0,0 +1,272 @@
...
+    let left = 0;
+    let {height: right} = await this.storeService.getLastProcessedBlock();
+    let searchNum = 0;
+    while (left < right) {
+      searchNum++;
+      const mid = Math.floor((left + right) / 2);
+      const header = await this.blockchainService.getRequiredHeaderForHeight(mid);
+
+      if (header.timestamp === timestamp) {
+        return header;
```
stwiname, line 252: This is almost never going to be the case. Blockchains don't produce blocks with any sort of syncronisation. 

I don't think this function is doing what it should be, which is to find the nearest block header before the input timestamp

packages/node-core/src/indexer/multiChainRewind.service.ts
```diff
@@ -0,0 +1,272 @@
...
+    }
+
+    const targetHeader = left
+      ? await this.blockchainService.getRequiredHeaderForHeight(left)
+      : {
+          blockHash: '',
+          blockHeight: 0,
+          parentHash: '',
+          timestamp,
+        };
```
stwiname, line 267: This doesn't seem right, we wouldn't get any valuable information from this

packages/node-core/src/indexer/multiChainRewind.service.ts
stwiname, line 1: Where are the tests for this file?

packages/node-core/src/utils/reindex.ts
stwiname: Still relevant

packages/node/src/blockchain.service.spec.ts
```diff
@@ -55,4 +55,10 @@ describe('BlockchainService', () => {
     const interval = await blockchainService.getChainInterval();
     expect(interval).toEqual(5000);
   });
+
+  it('can get the chain create time', async () => {
```
stwiname, line 59: Is a test needed for Shiden to make sure that specific functionality works?

packages/node/src/indexer/fetch.module.ts
```diff
@@ -62,6 +63,8 @@ import { IIndexerWorker } from './worker/worker';
       useClass: ProjectService,
       provide: 'IProjectService',
     },
+    MultiChainRewindService,
+    /* END: Move to node core */
```
stwiname, line 67: This comment can be removed

packages/node-core/src/indexer/fetch.service.ts
```diff
@@ -379,8 +398,11 @@ export class FetchService<DS extends BaseDataSource, B extends IBlockDispatcher<
     return this.latestFinalizedHeight;
   }
 
-  resetForIncorrectBestBlock(blockHeight: number): void {
```
stwiname, line 382: It would be worth checking that other SDKs dont use this function. But it probably can be removed.

packages/node-core/src/indexer/fetch.service.ts
```diff
@@ -379,8 +398,11 @@ export class FetchService<DS extends BaseDataSource, B extends IBlockDispatcher<
...
-    this.updateDictionary();
-    this.blockDispatcher.flushQueue(blockHeight);
+  @OnEvent(MultiChainRewindEvent.Rewind)
+  @OnEvent(MultiChainRewindEvent.RewindTimestampDecreased)
+  handleMultiChainRewindEvent(payload: MultiChainRewindPayload) {
+    logger.info(`Received rewind event, height: ${payload.height}`);
+    this.resetForNewDs(payload.height);
```
stwiname, line 405: Maybe this function needs to be renamed if it has other uses now.

packages/node-core/src/indexer/entities/GlobalData.entity.ts
```diff
@@ -0,0 +1,52 @@
+// Copyright 2020-2024 SubQuery Pte Ltd authors & contributors
+// SPDX-License-Identifier: GPL-3.0
+
+import {blake2AsHex} from '@subql/utils';
+import {BuildOptions, DataTypes, Model, Sequelize} from '@subql/x-sequelize';
+
+export const RewindTimestampKeyPrefix = 'rewindTimestamp';
+export const RewindLockKey = 'rewindLock';
+export type RewindTimestampKey = `${typeof RewindTimestampKeyPrefix}_${string}`;
```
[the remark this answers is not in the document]
yoozo: yes

packages/node-core/src/indexer/multiChainRewind.service.ts
```diff
@@ -0,0 +1,251 @@
...
+  }
+
+  onApplicationShutdown() {
+    this.sequelize.connectionManager.releaseConnection(this.pgListener as Connection);
+  }
+
+  async init(chainId: string, reindex: (targetHeader: Header) => Promise<void>) {
+    this.chainId = chainId;
+
+    if (this.nodeConfig.multiChain) {
```
stwiname, line 109: For readability its a good idea to exit early and unnest where possible.

So here you can do
```ts
if (!this.nodeConfig.multiChain) {
    return;
}
```

packages/node-core/src/indexer/multiChainRewind.service.ts
stwiname: Maybe the suggested change to the table DDL will solve that

packages/node-core/src/indexer/multiChainRewind.service.test.ts
```diff
@@ -0,0 +1,127 @@
...
+  it('should handle rewind correctly', async () => {
+    // Act: Set global rewind lock to rewind to block 5
+    const rewindTimestamp = genBlockTimestamp(5);
+    await multiChainRewindService.setGlobalRewindLock(rewindTimestamp);
+
+    // Wait briefly for the PostgreSQL listener to process the notification
+    await delay(1);
+
+    // Assert: Check that the service is in Rewinding state and has the correct waitRewindHeader
+    expect(multiChainRewindService.status).toBe(RewindStatus.Rewinding);
```
stwiname, line 106: Reading the code, this doesn't need a delay, the status would be updated instantly.

packages/node-core/src/indexer/store.service.ts
```diff
@@ -477,6 +498,30 @@ group by
...
+    if (hasValue(record)) {
+      return record.toJSON().value as number;
+    }
+  }
+
+  private async initChainRewindTimestamp() {
+    if (!this.config.multiChain) return;
+    if ((await this.getRewindTimestamp()) !== undefined) return;
+    const rewindTimestampKey = generateRewindTimestampKey(this.subqueryProject.network.chainId);
+    await this.globalDataRepo.create({key: rewindTimestampKey, value: 0});
```
stwiname, line 514: There could be an update between getting and setting meaning that values could be overwritten. It would be good to make this an atomic operation.

packages/node-core/src/indexer/storeModelProvider/global/global.ts
```diff
@@ -0,0 +1,185 @@
...
+   * Serialize the rewind lock
+   * @param rewindTimestamp ms
+   * @param chainTotal The total number of registered chains.
+   * @returns
+   */
+  private serializeRewindLock(rewindTimestamp: number, chainTotal: number): string {
+    return JSON.stringify({timestamp: rewindTimestamp, chainNum: chainTotal});
+  }
+
+  async getGlobalRewindStatus() {
```
stwiname, line 63: Can you please specify the return type.

packages/node-core/src/indexer/worker/worker.service.ts
```diff
@@ -33,7 +33,7 @@ export abstract class BaseWorkerService<
   private queue: AutoQueue<IBlock<B>>;
 
   protected abstract fetchChainBlock(heights: number, extra: E): Promise<IBlock<B>>;
-  protected abstract toBlockResponse(block: B): R;
+  protected abstract toBlockResponse(block: B): Promise<R>;
```
stwiname, line 36: What is the reason for this change?

packages/node/src/utils/substrate.ts
```diff
@@ -415,6 +443,21 @@ export async function fetchBlocksBatches(
...
+  api: ApiPromise,
+  header: OptionalTsHeader,
+): Promise<Header> {
+  if (!isFullHeader(header)) {
+    const timestamp = await getTimestampFromBlockHash(api, header.blockHash);
+    return { ...header, timestamp };
+  }
+  return header;
+}
+
```
stwiname, line 460: Is this redundant now that timestamp is required?

packages/node/src/utils/substrate.ts
```diff
@@ -90,6 +95,22 @@ export function getTimestamp({
...
 
+export async function getTimestampFromBlockHash(
+  api: ApiPromise,
+  blockHash: string,
+): Promise<Date> {
+  try {
+    const blockTimestamp = await (
+      await api.at(blockHash)
+    ).query.timestamp.now();
+    return new Date(blockTimestamp.toNumber());
```
stwiname, line 106: Is this needed? I thought for Shiden this still didn't work?

packages/node/src/utils/substrate.ts
```diff
@@ -90,6 +95,22 @@ export function getTimestamp({
   return undefined;
 }
 
+export async function getTimestampFromBlockHash(
```
stwiname, line 98: This seems to be very unreliable. It doesn't work for AssetHub. I think we should use the other way of getting block timestamps.

stwiname:
Trying to test the `reindex` subcommand but it doesn't work.
```
2025-03-20T03:27:46.301Z <StoreService> INFO Historical state is height
2025-03-20T03:27:46.496Z <dynamic-ds> INFO Loaded 0 dynamic datasources
2025-03-20 03:27:46        RPC-CORE: getHeader(hash?: BlockHash): Header:: Unable to retrieve header and parent from supplied hash
2025-03-20T03:27:46.990Z <CLI-Reindex> ERROR undefined Error: Reindex failed to execute
```

Theres a few problems:
* The historical state says height but it should be timestamp. Its also set correctly to timestamp in the db
* It cant get the header. I think this might be related to the wrong historical type
* There doesn't seem to be any sync logic? If other indexers are running then they need to rewind or if the `reindex` command is run for the other networks it should complete that chains reindex.

packages/node/src/utils/substrate.ts
yoozo: It works fine on newer blocks, but it can’t retrieve data properly on older blocks. We don’t seem to have a good solution for this.

yoozo:
Now, the multichain reindex requires two additional parameters:
```
--network-endpoint
--multichain
```

stwiname:
> Now, the multichain reindex requires two additional parameters:
> 
> ```
> --network-endpoint
> --multichain
> ```

Why is multichain required? Can't it tell from the db state?

yoozo:
Fixed, Modify to obtain through storeService

packages/node-core/src/subcommands/reindex.service.ts
```diff
@@ -79,6 +78,8 @@ export class ReindexService<P extends ISubqueryProject, DS extends BaseDataSourc
       timestamp,
     };
 
+    this.multiChainRewindService.chainId = this.project.network.chainId;
```
stwiname, line 81: Doing this seems to conflict with the `init` method on the multichain rewind service. It doesn't seem very intuitive that this should be used instead of `init`.

Also what happens here if this subcommand is run when there is a rewind already in progress initiated by another chain? A test for this would be good

packages/node-core/src/utils/reindex.ts
```diff
@@ -59,11 +62,17 @@ export async function reindex(
     logger.warn(
       `Skipping reindexing to ${storeService.historical} ${targetUnit}: current indexing height ${lastUnit} is behind requested ${storeService.historical}`
     );
+    if (storeService.isMultichain) {
+      const tx = await sequelize.transaction();
+      await multichainRewindService.releaseChainRewindLock(tx, targetUnit, true);
```
stwiname, line 67: What if there's an existing rewind in progress?

packages/node/src/utils/substrate.ts
```diff
@@ -48,9 +50,16 @@ export function substrateHeaderToHeader(header: SubstrateHeader): Header {
 }
 
 export function substrateBlockToHeader(block: SignedBlock): Header {
+  const timestamp = getTimestamp(block);
+  // TODO How can this be handled here? Is it possible to add a configuration that allows undefined or sets it to the same value as the height?
```
stwiname, line 54: Is this still a todo?

packages/node-core/src/indexer/entities/GlobalData.entity.ts
```diff
@@ -0,0 +1,59 @@
+// Copyright 2020-2024 SubQuery Pte Ltd authors & contributors
+// SPDX-License-Identifier: GPL-3.0
+
+import {BuildOptions, DataTypes, Model, Sequelize} from '@subql/x-sequelize';
+
+export enum MultiChainRewindStatus {
+  /** The current chain is in normal state. */
+  Normal = 'normal',
```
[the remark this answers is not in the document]
yoozo: No, when a rewind occurs on another chain, it is necessary to use the registered chain information here to send a rollback notification. It means that the current chain is in a normal indexing state.

packages/node-core/src/indexer/fetch.service.ts
```diff
@@ -197,6 +200,29 @@ export class FetchService<DS extends BaseDataSource, B extends IBlockDispatcher<
...
+        assert(this.multiChainRewindService.waitRewindHeader, 'Multi chain Rewind header is not set');
+        await this.projectService.reindex(this.multiChainRewindService.waitRewindHeader);
+        continue;
+      }
+
+      if (MultiChainRewindStatus.WaitOtherChain === multiChainStatus) {
+        logger.info(
+          `Waiting for all chains to complete rewind, current chainId: ${this.multiChainRewindService.chainId}`
+        );
+        await delay(multiChainRewindDelay);
```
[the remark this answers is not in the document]
yoozo: The delay here refers to the time needed for the rewind. Also, I don’t quite understand what you mean by needing a DB trigger. The handling process here is that after the pgListener in multiChainRewindService receives the rewind message, it will change the status to MultiChainRewindStatus.Incomplete, and then the fetchService will stop processing.

packages/node-core/src/indexer/storeModelProvider/global/global.ts
```diff
@@ -151,4 +174,32 @@ export class PlainGlobalModel implements IGlobalData {
...
+    const multiMetadataTables: string[] = flatten(tableRes).filter((value: string) => MULTI_METADATA_REGEX.test(value));
+    assert(
+      multiMetadataTables.length > 0,
+      `No multi metadata tables found in the database. Please check your schema or configuration.`
+    );
+
+    const metadataRes = await Promise.all(
+      multiMetadataTables.map((table) =>
+        this.sequelize.query<{value: string}>(
+          `SELECT "value" FROM "${this.dbSchema}"."${table}" WHERE "key" = 'genesisHash'`,
```
stwiname, line 183: I think `chainId` is a better option, not all networks use genesisHash

packages/node-core/src/indexer/storeModelProvider/global/global.ts
```diff
@@ -59,41 +61,65 @@ export class PlainGlobalModel implements IGlobalData {
...
+        return {lockTimestamp};
+      } catch (e: any) {
+        logger.error(
+          `setGlobalRewindLock failed chainId: ${this.chainId}, rewindTimestamp: ${rewindTimestamp}, errorMsg: ${e.message}`
+        );
+        await tx.rollback();
+        throw e;
+      }
+    } else {
+      const chainIds = await this.getChainIdsFromMetadata();
```
stwiname: I think this will want to use the transaction to avoid a new chain being created between this and bulk create

packages/node-core/src/indexer/entities/Metadata.entity.ts
```diff
@@ -8,6 +8,7 @@ import {HistoricalMode} from '../types';
 
 export interface MetadataKeys {
   chain: string;
+  chainId: string;
```
stwiname, line 11: Chain already exists, it was just named poorly but its the same value

packages/node/src/utils/substrate.ts
```diff
@@ -405,7 +433,7 @@ export async function fetchBlocksBatches(
     const wrappedEvents = wrapEvents(wrappedExtrinsics, events, wrappedBlock);
 
     return {
-      getHeader: () => substrateBlockToHeader(wrappedBlock),
+      getHeader: () => blockHeaderMap[idx],
```
stwiname, line 436: Whats the reason for this change?

packages/node-core/src/indexer/multiChainRewind.service.test.ts
```diff
@@ -0,0 +1,452 @@
...
+      expect(header10001).toEqual({
+        blockHeight: 10000,
+        timestamp: rewindDate10000,
+        blockHash: 'hash10000',
+        parentHash: 'hash9999',
+      });
+    });
+  });
+
+  describe('Init', () => {
```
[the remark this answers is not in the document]
yoozo: Case `Lock exists but rewind is not completed yet`.  The scenario you mentioned already has test cases.
This happens when an Incomplete state lock is found during initialization.

packages/node/src/blockchain.service.spec.ts
stwiname: This test is failing

packages/node/src/blockchain.service.spec.ts
```diff
@@ -55,4 +55,9 @@ describe('BlockchainService', () => {
...
```
stwiname, line 59: This test is failing

PR #2710, opened 2025-03-05
add tsconfig-paths-webpack-plugin

### LINKED ISSUE #2709: SubQl Build doesn't support typescript path mappings
Modern typescript codebases use path mappings to keep imports readable. https://www.typescriptlang.org/tsconfig/#paths

But looking at the webpack config here, ts-loader doesn't automatically resolve path mappings. https://github.com/TypeStrong/ts-loader/issues/213

https://github.com/subquery/subql/blob/5b858076430ec6ef68afbfe78a4eaea6ada8eb5b/packages/cli/src/controller/build-controller.ts#L38-L47

Alternatives would be to either configure paths by reading the project ts-config or use something like: https://github.com/duffman/tspath

seunlanlege:
# Description
Adds the [tsconfig-paths-webpack-plugin](https://github.com/dividab/tsconfig-paths-webpack-plugin) to the base webpack config for automatic paths resolutions

Fixes #2709

## Type of change

## Checklist

- [x] I have tested locally

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -27,6 +27,7 @@
     "terser-webpack-plugin": "^5.3.10",
     "ts-loader": "^9.5.1",
     "ts-node": "^10.9.2",
+    "tsconfig-paths-webpack-plugin": "^4.2.0",
     "tslib": "^2.6.3",
     "typescript": "^5.7.3",
     "update-notifier": "^5.1.0",
--- a/packages/cli/src/controller/build-controller.ts
+++ b/packages/cli/src/controller/build-controller.ts
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: GPL-3.0
 
 import assert from 'assert';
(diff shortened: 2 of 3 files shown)


stwiname (changes requested):
I've tested this locally and it breaks building for our starter projects (and existing projects). This change needs to be backwards compatible.

Here is the error i get:
```
ERROR
 ›         TS6059: File '/Users/scotttwiname/Projects/subql-main/packages/testing/src/index.ts' is not under 'rootDir' '/Users/scotttwiname/Projects/subql-starter/Kusama/kusama-starter/src'. 'rootDir' is expected
 ›   to contain all source files.
 ›     The file is in the program because:
 ›       Root file specified for compilation
```

seunlanlege:
I've made the plugin conditional @stwiname, the plugin will only be added if paths are specified in the tsconfig

stwiname (approved):

PR #2720, opened 2025-03-17
Process already fetched blocks on fetch block failure

stwiname:
# Description
The previous behaviour, when a block failed to fetch the indexer would instantly exit, discarding and already fetched blocks and any changes from processed blocks but not flushed to the db.

The new behaviour means that when a block fails to fetch it will process any already fetched blocks and flush the data before exiting.

This change was driven by chains with subsequent large blocks where indexing a block would take a large amount of time and the follow up block was also large and had a high probability of failing to fetch.

## Other changes

* Better abstraction between the BlockDispatcher and WorkerBlockDispatcher reducing duplicate code
* Tests for BlockDispatchers!
* Fix WorkerBlockDispatcher not discarding blocks on rewinds where `latestBufferedHeight` was not set correctly

Fixes # (issue)

## Type of change

## Checklist

- [ ] I have tested locally

### DIFF
--- a/packages/node-core/src/indexer/blockDispatcher/base-block-dispatcher.ts
+++ b/packages/node-core/src/indexer/blockDispatcher/base-block-dispatcher.ts
@@ -296,7 +296,7 @@ export abstract class BaseBlockDispatcher<Q extends IQueue, DS, B> implements IB
     processQueue: AutoQueue<void>;
     /**
      * A function to abort fetching any other blokcs. This is called when there is an error with the fetch task */
-    abortFetching: () => void;
+    abortFetching: () => Promise<void> | void;
     getHeader: (input: T) => Header;
     height: number;
   }): Promise<void> {
@@ -315,7 +315,7 @@ export abstract class BaseBlockDispatcher<Q extends IQueue, DS, B> implements IB
         if (!this.isShutdown) {
           this.isShutdown = true;
           this.fetchFailureHeight = options.height;
-          options.abortFetching();
+          await options.abortFetching();
           // this.fetchQueue.abort();
           // Wait for any pending blocks to be processed
           await options.processQueue.onIdle();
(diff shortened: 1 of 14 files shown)


packages/node-core/src/indexer/blockDispatcher/base-block-dispatcher.ts
```diff
@@ -187,6 +187,13 @@ export abstract class BaseBlockDispatcher<Q extends IQueue, DS, B> implements IB
           void this.poiSyncService.syncPoi();
         }
         this.eventEmitter.emit(IndexerEvent.RewindSuccess, {success: true, height: reindexBlockHeader.blockHeight});
+
+        // End transaction
+        await this.storeModelProvider.applyPendingChanges(
+          height,
+          !this.projectService.hasDataSourcesAfterHeight(height),
+          this.storeService.transaction
+        );
```
stwiname, line 196: Should this happen before the rewind? It seems wrong to apply cache changes, after rewind.

yoozo:
createIndexerWorker should inject the abortFetching method.\

packages/node-core/src/indexer/blockDispatcher/worker-block-dispatcher.ts
```diff
@@ -172,46 +176,30 @@ export class WorkerBlockDispatcher<
...
-
-        await this.preProcessBlock(header);
+    /*
+     Wrap the promise in a fetchQueue for 2 reasons:
+     1. It retains the order when resolving the fetched promises
+     2. It means `this.queue` doesn't fill up with tasks awaiting pendingBlocks which then means we can abort fetching and wait for idle
+    */
+    const pendingBlock = this.fetchQueue.put(() =>
+      this.blockchainService.fetchBlockWorker(worker, height, {workers: this.workers})
+    );
```
yoozo, line 187: If testing reveals that the process will die due to fetchQueue.put exceeding the maximum capacity and being unable to recover.

packages/node-core/src/indexer/blockDispatcher/base-block-dispatcher.ts
```diff
@@ -235,6 +247,109 @@ export abstract class BaseBlockDispatcher<Q extends IQueue, DS, B> implements IB
...
+          if (this.storeModelProvider instanceof StoreCacheService) {
+            await this.storeModelProvider.flushData(true);
+          }
+
+          exitWithError(new Error(`Failed to fetch block ${options.height}.`, {cause: e}), logger);
+        }
+      })
+      .then((data) => {
+        // Don't add more data to index if theres a failure fetching an earlier block
+        if (!data || (this.fetchFailureHeight && options.height > this.fetchFailureHeight)) {
```
yoozo, line 339: Here, it should be possible to directly determine using isShutdown.

yoozo (approved):

PR #2788, opened 2025-05-15
introduce chat api to the project supported by langchain's sql-agent

ianhe8x:
* 

# Description
integrate llm chat api to query data from database
  * support openai, claude-ai and other openai compatibles
  * use environment to set LLM secrets
  * an openai styled chat api is now exposed from /v1/chat/completions
  * this feature is enabled by default, use --chat=false to turn off
other minor changes
* remove project service to drop support for legacy version of subql node
Fixes # (issue)

## Type of change

## Checklist

- [x] I have tested locally


packages/query/package.json
```diff
@@ -50,12 +54,14 @@
...
     "graphql-query-complexity": "^0.11.0",
     "graphql-ws": "^5.16.0",
+    "langchain": "^0.3.24",
     "lodash": "^4.17.21",
     "pg": "^8.12.0",
     "pg-tsquery": "^8.4.2",
     "postgraphile": "^4.13.0",
     "postgraphile-plugin-connection-filter": "^2.2.2",
     "rxjs": "^7.1.0",
+    "typeorm": "^0.3.23",
```
stwiname, line 64: Does langchain require typeorm? Could we use sequelize here for consistency?

packages/query/src/configure/configure.module.ts
```diff
@@ -85,6 +86,17 @@ export class ConfigureModule {
         pgClient._explainResults = [];
       });
     }
+    // todo: support ssl
+    const dataSource = new DataSource({
```
stwiname, line 90: Can we specify this as read only?

packages/query/src/llm/chat.module.ts
```diff
@@ -0,0 +1,186 @@
...
+            res.write(
+              `data: ${JSON.stringify({
+                id: `chatcmpl-${Date.now()}`,
+                object: 'chat.completion.chunk',
+                created: Math.floor(Date.now() / 1000),
+                model: process.env.OPENAI_MODEL,
+                choices: [
+                  {
+                    index: 0,
+                    message: {role: 'assistant', content: fullResponse},
```
stwiname, line 140: Is this the right behaviour to send the full response again at the end of a stream?

packages/query/src/configure/configure.module.ts
ianhe8x: I couldn't find a way. I think user may have to set up the readonly permission on db user

packages/query/src/llm/chat.module.ts
ianhe8x: This part will need some clean up. Changed from initial version, i return the tool response as reasoning to user.
Although it is called fullResponse it is just the last message from llm which is considered the final answer of the question so it was send to user.

PR #2904, opened 2025-09-04
feat: implemented dynamic datasource removing themseleves and adding them at end blocks

### LINKED ISSUE #2099: Dynamic Datasources destructor
Now that we have support for [Datasource EndBlocks](https://github.com/subquery/subql/pull/2064) it should be possible for dynamic datasources to remove themselves and set an end blocks.

From within a handler we can inject a function to destroy the current datasource and it should succeed if it is a dynamic datasource.

Gmin2:
# Description
Implemented dynamic datasource removing themseleves and adding them at end blocks

Fixes #2099 

## Type of change

## Checklist

- [x] I have tested locally

## Summary by CodeRabbit

* **New Features**
  * Destroy dynamic datasources by template+index; query active datasources by template; lookup datasource params by index.
  * Optional endBlock on datasources to mark lifecycle end; host/worker/global bindings expose destroy/get operations.

* **Behavior**
  * Indexing now supports in-place datasource removal during processing and excludes datasources whose endBlock equals or is below the current processing height.

* **Tests**
  * Expanded coverage for destruction flows, endBlock propagation, template-scoped queries, error cases, and in-memory lifecycle semantics.


stwiname (changes requested):
This is on the right track but it is flawed currently. 

With dynamic datasources there can be multiple instances from the same template, the only differences being the start block and arguments. So there needs to be more information passed through to remove the correct datasource. 

I think the best solution for this is to provide the ability to list dynamic datasources for a template. Then the user can select the appropriate one to remove by providing the index.

Gmin2:
@stwiname just implemented listing of dynamic datasources and to remove it by index no

packages/node-core/src/indexer/dynamic-ds.service.ts
```diff
@@ -91,6 +94,74 @@ export class DynamicDsService<DS extends BaseDataSource = BaseDataSource, P exte
...
+    if (!this._datasourceParams) {
+      throw new Error('DynamicDsService has not been initialized');
+    }
+
+    const matchingDatasources = this._datasourceParams
+      .map((params, globalIndex) => ({params, globalIndex}))
+      .filter(({params}) => params.templateName === templateName && params.endBlock === undefined);
+
+    return matchingDatasources.map(({globalIndex, params}, index) => ({
+      index,
```
stwiname: Should this be the global index?

packages/node-core/src/indexer/dynamic-ds.service.ts
```diff
@@ -91,6 +94,74 @@ export class DynamicDsService<DS extends BaseDataSource = BaseDataSource, P exte
...
+      );
+    }
+
+    // Get the datasource info at the specified index
+    const dsInfo = matchingDatasourcesInfo[index];
+
+    // Find the global index in _datasourceParams
+    const globalIndex = this._datasourceParams.findIndex(
+      (p) => p.templateName === dsInfo.templateName && p.startBlock === dsInfo.startBlock && p.endBlock === undefined
+    );
```
stwiname: Why not just have the index on `DynamicDatasourceInfo` then there's no need to filter by templateName, find the index, then find the global index.

packages/node-core/src/indexer/indexer.manager.ts
```diff
@@ -116,6 +116,32 @@ export abstract class BaseIndexerManager<
...
+        vm.freeze(async (templateName: string, index: number) => {
+          await this.dynamicDsService.destroyDynamicDatasource(templateName, blockHeight, index);
+
+          // Remove destroyed datasources from current processing
+          // Filter out datasources that have been destroyed
+          const updatedFilteredDataSources = filteredDataSources.filter((fds) => {
+            const dsParam = (this.dynamicDsService as any)._datasourceParams?.find(
+              (p: any) =>
+                p.templateName === (fds as any).mapping?.file?.split('/').pop()?.replace('.js', '') ||
+                (p.startBlock === (fds as any).startBlock && p.templateName === templateName)
```
stwiname: I don't think this is sufficient, there could be multiple instances of a dynamic datasource with the same startBlock but different args.

packages/node-core/src/indexer/indexer.manager.ts
```diff
@@ -116,6 +116,32 @@ export abstract class BaseIndexerManager<
...
+            );
+            // Keep datasource if it does not have an endBlock (still active)
+            return !dsParam || dsParam.endBlock === undefined;
+          });
+
+          // Update the filteredDataSources array in place
+          filteredDataSources.length = 0;
+          filteredDataSources.push(...updatedFilteredDataSources);
```
stwiname: Why not just `filteredDatasources = updatedFilteredDataSources`?

packages/node-core/src/indexer/dynamic-ds.service.ts
Gmin2: yes thnx for this !

packages/node-core/src/indexer/dynamic-ds.service.ts
```diff
@@ -91,6 +95,90 @@ export class DynamicDsService<DS extends BaseDataSource = BaseDataSource, P exte
...
+    const updatedParams = {...dsParam, endBlock: currentBlockHeight};
+    this._datasourceParams[index] = updatedParams;
+
+    // Update the datasource object if it exists
+    // Note: _datasources and _datasourceParams arrays should always be in sync.
+    // If the index is valid for params, it must also be valid for datasources.
+    if (!this._datasources[index]) {
+      throw new Error(`Datasources array out of sync with params at index ${index}`);
+    }
+    (this._datasources[index] as any).endBlock = currentBlockHeight;
```
stwiname: Can the `any` cast be removed here please

packages/node/src/indexer/indexer.manager.ts
```diff
@@ -191,6 +195,14 @@ export class IndexerManager extends BaseIndexerManager<
...
     return async (content, dataSources, getVM) => {
       for (const ds of dataSources) {
+        // Skip datasources that have been destroyed at or before this block
+        // When a datasource is destroyed, its endBlock is set to the current blockHeight
+        // We want to exclude it from processing in subsequent handlers within the same block
+        const endBlock =
+          'endBlock' in ds ? (ds as { endBlock?: number }).endBlock : undefined;
+        if (endBlock !== undefined && endBlock <= blockHeight) {
+          continue;
+        }
```
stwiname: This is a bad place to do this, the provided dataSources should already have the filtered dataSources to be relevant.

Having this here makes it really hard to find as it's separate from the rest of the filtering

stwiname (commented):
I still don't see a test that addresses this case. https://github.com/subquery/subql/pull/2904#discussion_r2485235677

packages/node/src/indexer/indexer.manager.ts
Gmin2: Now all filtering happens centrally in `filterDataSources` method  in `filterDataSources` method and the VM callback calls it to reassign the filtered datasources.

stwiname:
> > I still don't see a test that addresses this case. [#2904 (comment)](https://github.com/subquery/subql/pull/2904#discussion_r2485235677)
> 
> Added two test case in [5cf5af1](https://github.com/subquery/subql/commit/5cf5af124bae599226dec4c7cbabc6a0f4a7c48f) to check the filtering logic or am I missing something really obivious ones

There needs to be tests on indexer manager to confirm the behaviour

Gmin2:
Hey @stwiname just implmented the test to verify that the midblock destroys in https://github.com/subquery/subql/pull/2904/changes/de059a3d83fb23ad5494db8db0e9878e55cfa021

just final look

PR #2966, opened 2025-11-15
fix: improve validation error details in validateCommonProjectManifest

Lil-Duckling-22:
Replace TODO comment with detailed error formatting for validation failures. Error messages now include property names and constraint details instead of generic toString() output, making it easier to identify validation issues.

## Summary by CodeRabbit

* **Bug Fixes**
  * Validation errors now show structured, multi-line messages with per-property details and precise paths (including nested fields and array indices) for clearer diagnostics.

* **Tests**
  * Added comprehensive unit tests covering invalid and valid manifests, nested-field validations, array-indexed paths, and the enhanced error-message formatting.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/packages/common/src/project/load.spec.ts
+++ b/packages/common/src/project/load.spec.ts
@@ -31,6 +31,7 @@ describe('validateCommonProjectManifest', () => {
 
     try {
       validateCommonProjectManifest(invalidManifest);
+      fail('Expected validation to throw an error');
     } catch (error: any) {
       const errorMessage = error.message;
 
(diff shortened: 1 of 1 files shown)


stwiname:
Hi @Lil-Duckling-22, thanks for your PR. Can you please update tests and provide an example of how the error output has changed

packages/common/src/project/load.spec.ts
stwiname: All tests that have a try/catch should also have a `fail` call. If the function being tested doesn't throw then there will be no expectations.

